PR #143, opened 2026-07-26
docs(core): add adversarial review as Gate 3 of the merge policy

vinnie357:
Codifies the third merge gate. The repo has been running three gates in practice for a while; only two were written down.

`/core:git` documented a **Dual-Gate CI Policy** — local `mise run ci`, then remote `gh pr checks` — and said "Two gates protect main. Neither is optional." Adversarial review of the PR was convention, not doctrine, so nobody inherited it.

Every PR gets a review from a separate agent on the strongest thinking model the harness offers, named by capability rather than by model literal — the repo's own convention forbids hardcoded model names, and the first draft violated it. **No exemption** — not for one-line fixes, documentation, or version bumps.

The reviewer is briefed to **defeat the change, not approve it**. That framing is the whole mechanism: a reviewer told to "check this over" returns nothing useful; one told "assume a defect exists and find it" returns the defect.

**Three focus areas are standing**, carried by every review whatever else the brief adds:

1. **Restraint** — does the change need to exist, is it the minimum that works, does it add anything unasked for. A reviewer that only hunts bugs approves well-built things that should not have been built.
2. **Documentation the change obligates** — user docs, READMEs, ADRs, system-design documents, and any skill or reference describing the thing. A change that silently invalidates a document is a defect with a delayed fuse.
3. **Claims in the PR body** — every measurement, count, and "verified" is checkable, so check them.

Plus the mechanics: a separate agent (the author cannot review its own work), read-only against the shared tree with a scratchpad clone for destructive tests, findings addressed *or answered* rather than waved through, and the reviewer's claims verified independently — a review is evidence, not a verdict.

## Gate 3 is not the pipeline's review tier

`/core:agent-loop` already has reviewers: P5 verifies tests exercise the acceptance criteria, the test-review stage checks for redundancy. Those judge whether an implementation meets **the issue's spec**.

Gate 3 is identified by its **brief, not its timing** — an earlier draft distinguished it as running "after a PR exists", which does not hold: Forge's Final Reviewer reads the full PR diff and is still not Gate 3. Gate 3 is the review carrying the defeat-the-change brief and the three standing focus areas, dispatched after the pipeline reports done.

## Gate 3 is observable, not asserted

Gates 1 and 2 have artifacts: `mise run ci` output and `gh pr checks`. Gate 3 requires the verdict and each finding's disposition to land as a durable record on the PR, checkable with `gh pr view <n> --comments`. Without that, "review done" is an unverifiable assertion of exactly the kind the gate exists to catch — which review of this PR found, and the policy as first written would not have.

## Why three and not two

Gates 1 and 2 prove the suite passes. They do not prove the change is correct. The doctrine states that without freezing session-specific counts into it — numbers a future reader cannot verify become lore, so the skill names the *classes* of defect a passing suite cannot see instead.

## All five citation sites moved together

A rename that leaves stale citations creates exactly the drift this repo has spent days removing, so the policy and everything pointing at it move in one commit:

- `core/git/SKILL.md` — the policy section, and Merge Strategy now shows all three gates in its command block
- `core/agent-loop/SKILL.md` — Key Conventions, plus the Agent Worker execution order, where step 8 no longer calls a PR "merge-ready" at Gate 2 and step 9 states that Gate 3 is the leader's to arrange
- `claude-code/templates/CLAUDE.md` — the step order, and "once both gates are green" → "all three"

`grep` for `Dual-Gate|dual gate|two gates|both gates` returns one unrelated hit — `proxmox/references/golden-image.md` on PVE version-support gates. No merge-policy citation remains stale.

One deliberate non-change: `rust/agents/rust-implementer.md` uses "Gate 1 / Gate 2" for its own TDD diff-boundary and CI steps. That is an agent-local scheme inside a numbered list, contextually unambiguous, and renaming it would be scope creep into another plugin.


vinnie357:
**Gate 3 — adversarial review (fable): MERGE WITH FIXES.** All findings dispositioned:

| Finding | Disposition |
|---|---|
| F1 — Gate 3 had no verifiable artifact; "review done" was unassertable | **Fixed** — requirement added that the verdict and each finding's disposition land as a durable record on the PR, checkable via `gh pr view <n> --comments`. This comment is the first instance. |
| F2 — PR body claimed the grep returns nothing; it returns one hit | **Fixed** — body corrected to name the unrelated proxmox hit |
| F3 — the P5 distinction rested on timing that does not hold (Forge's Final Reviewer reads the full PR diff) | **Fixed** — Gate 3 is now identified by its brief, not its timing, in both git/SKILL.md and agent-loop |
| F4 — hardcoded model names contradict the repo's own no-literals convention | **Fixed** — capability phrasing, deferring to agent-loop's model-overrides |
| F5 — three restraint cuts: session-specific numbers frozen into doctrine, unverified psychology, redundant metaphor half | **Fixed** — all three cut |
| rust-implementer Gate 1/Gate 2 numbering collision | **Filed** as claude-skills-147 (reviewer judged the non-change defensible; rename is a cheap follow-up) |

Recursion note from the review, worth recording: F2 and F5 fall directly out of the policy's own standing focus areas, so the policy would have produced them. F1 is the gap it would *not* have — nothing forced the review itself to be recorded. That is why F1 was the top finding and why this comment exists.

PR #144, opened 2026-07-26
test(quality): add cross-file duplicate-block check with three exemptions

vinnie357:
- Add corpus-wide duplicate-block check to test/validate-skills-quality.nu: 8-normalised-line sliding window over git-tracked .md/.sh under plugins/ plus root CLAUDE.md, grouped per file set so one duplicated section is one finding
- Normalisation: trim, strip bullet/ordered markers and backticks, collapse whitespace; drop blank/punctuation-only/fence-delimiter lines; single-word lines kept (filtering them hides the core-list flagship case)
- Three exemptions, each with stated reason: all-templates file sets, same-skill file sets, core-list satellite subsets (deferred to test/validate-core-list.nu; satellite list derived by parsing that script, not a second copy)
- New baseline key shape dupe/<md5-8 of sorted member set>:duplicate_block; duplicate_block added to DETAIL_CHECKS so waived groups ratchet on window count; member paths printed in the report since the hash alone is not actionable
- 14 new self-test cases (detection, normalisation, exemptions, key stability, baseline round-trip, hard-failure/stale/count-regression ratchet paths)
- Baseline 7 observed groups (plan predicted 9 at the same corpus sha; the plan's per-pair counts came from its ad-hoc measurement pipelines — the implementation follows the plan's stated Configuration exactly and 7 is what that config measures)
- File burn-down issues claude-skills-148 (beads/bees worker dedupe with corrected skills:-preloading constraint), 149 (five cross-skill dedupe groups), 150 (zig pair assessment)


vinnie357:

Verdict: **MERGE WITH FIXES**. Blocking finding applied in 6bca088; nits filed.

| # | Finding | Disposition |
|---|---|---|
| 1 | BLOCKING — `test/README.md` described the validator as two passes with `plugin/skill:check` keys only; zero mentions of the new pass. A maintainer would have read a `dupe/<hash>` key as invalid. | **Fixed.** Documented the third pass, its corpus scope, the 8-line window and normalisation (including why single-word lines are kept), the key shape, `detail_count`, and all three exemptions with reasons. |
| 2 | NIT — removing the `failing_counts` append in Pass 3 leaves self-tests green; nothing pins that main feeds counts into the ratchet. Architectural, shared with every pre-existing detail check, not introduced here. | **Filed as claude-skills-151.** Behaviour verified correct today: growth exits 1 (`current 8 > baselined 7`), shrink exits 1 on stale count. |
| 3 | NIT — exemption 3's satellite parse matches `path: "` anywhere in `validate-core-list.nu`; self-test asserts only a lower bound. | **Filed into claude-skills-151.** Safe in the failure direction (verified: emptying it surfaces the group as a loud failure), unguarded in the over-match direction. |
| 4 | NIT — claude-skills-150 implied corpus drift removed four zig pairs and an awman pair. | **Fixed.** Re-measurement at the plan's pinned sha returned the identical 7 groups. The extra rows were artifacts of the plan's ad-hoc pipelines; the awman pair exists but is exempted as same-skill. |

Attacks that failed, with the residual risk each leaves:

- **7-vs-9 groups** — re-ran the shipped check against the plan's pinned sha `8e6153a`: identical 7 groups, keys, and counts. No real duplication is silently missed.
- **Key stability** — rename or membership change fails loudly in both directions (stale key + unbaselined set), member paths printed. md5-8 collision at this group count is negligible; accepted residual.
- **Self-tests bite** — mutations caught: exemption-3 branch removed, unsorted key, `duplicate_block` dropped from `DETAIL_CHECKS`. Only uncaught mutation is finding 2.
- **Absorption trap** — `duplicate_block` is in `DETAIL_CHECKS`; null `detail_count` rejected; a waived group that shrinks forces the baseline down rather than silently staying high.
- **Restraint** — each exemption absorbs measured intentional duplication (5 templates, 4 same-skill, 1 core-list). None is a window-size workaround.

Gates: local `mise run ci` exit 0 · `gh pr checks` SUCCESS on 6bca088 · this review.

PR #145, opened 2026-07-26
fix(core): permit core-skill subsetting for tiers that write no code

vinnie357:
Closes claude-skills-125.

`agent-loop/SKILL.md` claimed "Every agent at every tier loads these before any work" for the ten mandatory core skills, while two tier references loaded 3 and 4 of them. Neither file is in `SATELLITES`, so the contradiction was invisible to CI and predates this work.

Operator ruling: per-tier subsets are legitimate, and the discriminator is whether the tier writes code — the CI runner does not need `/core:tdd`, the agent addressing its findings does.

- The canonical claim now reads "Every agent that writes code loads these before any work", with the exemption, its discriminator, and its limit stated: omit a skill only when the tier cannot exercise it.
- `validator.md` drops `/core:tdd` and states why. It keeps `/core:anti-fabrication`, which is the one skill that tier can least afford to omit — its report is the only evidence the lead has.
- `fix-agent.md` keeps `/core:tdd` and states why.

**No new check was needed, because the real risk is already covered.** A subsetting tier could name a skill that does not exist, or one that goes stale after a core-list change. The `invocations` check in `validate-skills-quality.nu` already scans `references/` and fails the run on any unresolvable `/plugin:skill` token — verified empirically in a scratchpad clone by inserting `/core:nonexistent-skill` into `validator.md`, which produced a hard failure. So the two files did not need adding to `SATELLITES`, and no subset-aware satellite shape was needed.

The reworded anchor tripped the core-list drift check, which named the exact fix (`SATELLITES` anchor must move with the wording). Anchor updated in the same commit — that check is #139's machinery catching a real edit rather than a synthetic one.

One judgment left open rather than decided silently: `fix-agent.md` writes code, and `/core:restraint` is a standing principle for code work by this repo's own doctrine, so arguably it belongs in that list too. The operator ruling covered `/core:tdd`, so the list is otherwise unchanged. Two additions I had drafted — `/core:security` to the validator and `/core:restraint` to the fix agent — were reverted before commit; the security one had an invented justification (validator.md never ran gitleaks).


vinnie357:

Verdict: **MERGE WITH FIXES**. Both blocking findings applied in a4fb885; one nit adopted, one filed.

| # | Finding | Disposition |
|---|---|---|
| 1 | **BLOCKING** — the policy contradicted itself. Headline said agents that *write code* load all ten; the exemption paragraph granted subsetting to tiers that write *no* code, then named `fix-agent.md` as one of the two — a file that explicitly writes code. So fix-agent subsetted under a permission whose own discriminator excluded it. | **Fixed.** Rewrote the exemption as **per-skill, not per-tier**: a tier may omit a skill its role cannot exercise. `/core:tdd` is the worked case. The headline reverts to the original wording plus a pointer to the exemption, so the `SATELLITES` anchor also reverts — no second anchor change. |
| 2 | **BLOCKING** — stale comment in `validate-core-list.nu`, a file this PR edits: it said whether subsets are legitimate "is claude-skills-125" (this PR decides it) and cited validator's list as 3 names (now 2). | **Fixed.** Comment states subsets are policy, corrects both counts, and records that membership is unchecked. |
| 3 | NIT — `/core:security` revert correct (validator never ran gitleaks; the drafted justification was invented). `/core:restraint` revert defensible, but fix-agent already restates restraint's ladder as prose. | **Adopted.** Verified the restatement at fix-agent.md:21, 22, 24, 58, so under the new per-skill test that tier demonstrably exercises restraint. Added, with the restatement named as the reason. This resolves the judgment the PR body left open. |
| 4 | NIT — the exemption's floor has no teeth: deleting `/core:anti-fabrication` from validator.md leaves both validators at exit 0. | **Filed as claude-skills-152** with the proposed `{path, expected: [...]}` shape. The SKILL.md text now says plainly that omissions are unchecked rather than implying otherwise. |
| 5 | NIT — no-code tiers (lead, planner, reviewer, hands) are no longer obligated by the headline. | **No change.** The per-skill rewrite restores the general mandate, so this dissolves: the headline binds every tier again and only individual skills are omissible. |

Attacks that failed:

- **The no-new-check justification HOLDS.** A bogus name in either tier file fails (`core/agent-loop:invocations`). The stale case was tested properly by removing `./skills/tdd` from the core plugin manifest — every `/core:tdd` mention then fails, across agent-loop, work.md, and restraint. Registry is built from plugin.json skill arrays and `references/` content is scanned, so a subsetting tier cannot carry a name that has gone away.
- **Anchor attacks** — a fenced copy of the lead-in cannot steal the anchor; deleting the real anchor fails loudly; a duplicated unfenced anchor reports both line numbers. New anchor occurs exactly once.
- **Docs sweep** — no stale site outside finding 2. `templates/CLAUDE.md`'s "always load these" stays valid since subsetting is *may*, not *must*; restraint's "binds at every tier" refers to the ladder's phases, not agent tiers.

Gates: local `mise run ci` exit 0 · `gh pr checks` SUCCESS,SUCCESS on a4fb885 · this review.

PR #146, opened 2026-07-27
chore(deps): remove this repo's dead beads installation

vinnie357:
Closes claude-skills-136.

This repo's CLAUDE.md has said bees is the tracker for a long time, but beads was still installed and still firing. Its `post-merge` hook printed `Failed to import bd changes after merge` on every pull and on both merges earlier tonight.

**Step 1 first: the tracked issue history was checked before anything was deleted.** Of the 9 issues in `.beads/issues.jsonl`, 8 were closed and their outcomes are all present in the marketplace (rust multi-skill, zig, claude-teams, slidev). One was **open** — a fully-specified P2 TypeScript plugin spec, with no such plugin in `marketplace.json` and nothing in bees covering it. Deleting `.beads/` would have destroyed live work, so it was migrated to **claude-skills-153** with the original 72-line spec preserved verbatim, and the migration verified before the delete.

Removed:

- `.beads/` — 6 tracked files, plus the untracked `embeddeddolt/` directory that has been noise in every `git status` for two days
- `.git/hooks/post-merge` — the local beads import hook
- `.gitattributes` — deleted entirely rather than left empty; its only content was the `merge=beads` driver, and bees uses none
- `mise.toml` — the beads tool install and all 20 `beads:*` tasks, **221 lines**. Operator-approved scope extension: the issue's evidence list stopped at the 6 tracked files, but `mise install` was still pulling the beads binary, which is the same "still wired in" complaint.

`.beads/` added to `.gitignore` so a stray local install cannot re-add it.

**`/core:beads` is untouched and stays in the marketplace.** It remains a legitimate skill for users whose repos use beads, and it does not depend on this repo's mise tasks — it tells users to add their own (`beads/SKILL.md:53`). This PR removes an installation, not a skill.

Documentation the change obligated: this repo's own PR workflow instructed `bd close`, `git add .beads/`, and `bd ready` at steps 4 and 7 — dead instructions now, updated to their bees equivalents. Two commit/PR format examples used "add beads tooling to mise.toml" as the model change, which would have read strangely in the same file that deletes it.

Verified: `mise.toml` parses (46 tasks, 4 tools); `mise tasks` shows 0 beads and 17 bees; `mise run ci` exit 0.


vinnie357:

Verdict: **MERGE WITH FIXES**. Both blocking findings applied in 98b0e91.

| # | Finding | Disposition |
|---|---|---|
| 1 | **BLOCKING — partial data loss.** The migrated issue carried **2 comments and 4 labels** that were dropped. Both comments are titled CORRECTION and *supersede the description*: the plugin goes under `plugins/languages/typescript/` (post-reorg), not root `typescript/`, and the structure is the progressive-disclosure pattern with 6 skills, not 4. An agent picking up claude-skills-153 as migrated would have built the wrong thing at the wrong path. | **Fixed.** Both comments recovered verbatim from `b3bef93~1:.beads/issues.jsonl` and re-added with provenance headers marking them as superseding; all 4 labels restored. Verified: 8 mentions of the corrected path, labels present. |
| 2 | **BLOCKING** — `AGENTS.md` still read "This project uses **bd** (beads) for issue tracking. Run `bd onboard`", plus a mandatory `bd sync` session workflow, pointing any non-Claude harness at a deleted `.beads/`. | **Fixed.** Converted to a symlink to `CLAUDE.md`, which is the workspace convention ("one source of truth, no drift") — the file was previously neither a symlink nor cross-named, which is how it drifted. Mode 120000 confirmed. |
| 3 | NIT — "migration verified" in the PR body was overstated; verification covered the description body only. | **Acknowledged here.** It was verbatim for the description and missed the comments entirely — the claim was true of what I checked and false of what mattered. |

**Why finding 1 is the important one:** the whole point of step 1 was preventing data loss, and I reported it done. The description *was* verbatim; I never checked whether the record had other fields. The data was recoverable only because it still existed in git history one commit back.

Attacks that failed:

- **Issue census independently recounted** from git history: 9 issues, exactly 8 closed + 1 open. No other live work destroyed. 4 of 8 closed issues spot-checked as genuinely shipped.
- **`.beads/interactions.jsonl` really was 0 bytes**; `metadata.json` held only db/export filenames. Nothing else destroyed.
- **mise.toml precision** — `mise tasks` set difference between main and branch is exactly the 20 `beads:*` tasks, nothing else; 66 → 46 tasks, 17 `bees:*` intact, tool drop is exactly the beads entry.
- **`.gitattributes`** — only 2 commits ever touched it, sole content was the `merge=beads` driver, no linguist/LFS/eol reliance. Full deletion correct.
- **No recreation vectors** — no `mise.lock`, no beads in workflows, no hooks remain.
- **Docs (a) vs (b)** — README, `/core:setup`, and `templates/CLAUDE.md`'s `bd ready` option all describe the beads *tool* for other repos and legitimately survive. `/core:setup` defaulting new projects to beads while our own convention is bees is filed as **claude-skills-154**.

Gates: local `mise run ci` exit 0 · `gh pr checks` SUCCESS,SUCCESS on 98b0e91 · this review.

PR #147, opened 2026-07-27
test(quality): add syntax-vs-usage vocabulary check for three formats

vinnie357:
- Add a fourth validator pass comparing each format's DOCUMENTED token vocabulary (documenting skill's SKILL.md + references/*.md, prose and tables included) against the REAL vocabulary in repo instances; fires on disjoint sets or a foreign-family token in the doc
- Three formats: commands (real: `plugins/**/commands/*.md`, foreign family `{{...}}`), agents (frontmatter keys of `plugins/**/agents/*.md`), hooks (`hooks` keys of `plugins/**/hooks/hooks.json`)
- Empty documented vocabulary is a hard error (extractor canary), never silence; empty real vocabulary stays silent
- Real command side never runs a general \$[a-z_]+ regex (bash $USER/$SESSION_ID/$BUILD_DIR pollution); named-arg tokens count only when declared in `arguments:` frontmatter, normalised to a $name family marker on both sides
- Key shape `syntax/<format>:vocab_disjoint`; `vocab_disjoint` added to DETAIL_CHECKS with detail_count = size of the disjoint documented set; baseline _comment updated in template and shipped file
- 8 new self-test cases including bash-pollution, mixed foreign-family, doc-empty canary, and an inlined historical replay of the pre-#134 claude-commands defect; 0 new waivers (count stays 101)
- Document the pass in test/README.md with the tripwire framing (real vocabs are 1 command token / 1 hook event / 5 agent keys)


vinnie357:
## Gate 3 — adversarial review (two rounds)

Verdict: **MERGE**. Round 1 returned six findings on 95cd9be; all applied. The fix round introduced new machinery, so the delta was re-reviewed rather than merged on a stale verdict. Round 2 verdict: MERGE — the reviewer tried to break the floor mechanism and the frontmatter stripping and could not.

### Round 1 — six findings, all applied

| # | Finding | Disposition |
|---|---|---|
| 1 | **BLOCKING — the agents leg could never fire.** The doc extractor read the documenting skill's own YAML frontmatter, whose `name:`/`description:` are permanently in the real agent key set, so that intersection was never empty. A fabricated rewrite documenting only `role:`/`capabilities:`/`persona:` returned silent. | **Fixed** — and wider than the finding. Checking all three legs found **all three contaminated**: `claude-commands`' frontmatter description mentions `$ARGUMENTS`/`$N`, `claude-hooks`' mentions `SessionStart`/`PreToolUse`. `strip-doc-frontmatter` now applies to every doc extractor. |
| 2 | **BLOCKING — nothing guarded the real-side globs.** Mutating the commands glob `**`→`*` left every self-test and the full run green; so did pointing a glob at a nonexistent directory. | **Fixed** via measured floors. See the deviation note below — a zero-files guard provably cannot catch this. |
| 3 | NIT — a lowercase `\$[a-z_]+` real-side regex escaped all 8 vocab cases, because the pollution fixture used only uppercase bash vars. It would pull `$claim` into the real set and silence the disjointness leg permanently. | **Fixed** — fixture now carries `$file` and `$claim`. |
| 4 | NIT — README implied foreign-family closed the one-token margin; it closes it for the **brace family only**. | **Fixed** — the same-family residual (`$ARGV`/`$INPUT` alongside one real token stays silent) is now stated plainly, with the reason it is not closed mechanically. |
| 5 | NIT — root README described 2 of 4 passes. | **Fixed** — names all four. |
| 6 | NIT — informational, doc-side extraction yields 6 agent keys not 16. | No behaviour change. |

### The deviation, which was better than my instruction

I asked for a zero-files-matched guard. That **cannot** catch the `**`→`*` under-match: single-star matches 9 files, and both `$ARGUMENTS` users live inside those 9, so the real vocabulary is byte-identical and file count is the only observable. Implemented instead as `VOCAB_REAL_FLOORS = {commands: 25, agents: 29, hooks: 2}` — a hard, non-baselineable error whose message names both causes and the exact knob.

Round 2 attacked that design specifically and endorsed it: derived floors would be **worse** (the pathspec would be authored by the same person from the same pattern — correlated failure, and git-pathspec vs nushell-glob `**` semantics differ), and an exact-match floor test recreates the per-file-add noise the lower-bound design avoids. One residual accepted: an *over*-matching glob passes the floor, but a superset corpus only adds real tokens and the extractors stay narrow.

### Two NITs from round 2, applied here rather than deferred

- **Off-by-one in the strip escaped all 10 cases** — the fixture put no token on the first body line and asserted membership. `role:` is now the first body line with set-**equality**; verified by mutating `skip (+2)`→`(+3)`, which now reports `[capabilities, persona]` with `role` eaten.
- **Stale-low floors escaped** — `agents: 29→1` left all 61 self-tests green. `floor_under` now has a row per format; verified the same mutation now fails.

One residual stated rather than closed: unwiring the floor check from `main` leaves self-tests green, which is true of every pass including the doc-empty canary — the self-tests are unit-level by architecture. Closing it needs an integration harness against a fixture repo, disproportionate for a tripwire.

### Verified across both rounds

- Historical replay fires via **pure disjointness** — defect-era `SKILL.md` with `references/` untracked, as it genuinely was: documented `{{}}` vs real `$ARGUMENTS`, exit 1.
- All three doc-empty canaries fire even with token-bearing frontmatter retained — stronger than the original test, which had to blank the whole file.
- Agents and hooks legs confirmed **live in both directions** after stripping.
- Waivers exactly 101, `--update-baseline` byte-stable, ratchet round-trip intact, globs find 25/29/2.

Gates: local `mise run ci` exit 0 · `gh pr checks` SUCCESS,SUCCESS on 87bf111 · this review.

PR #148, opened 2026-07-27
fix(test): skip tracked-but-deleted files in the corpus scan instead of crashing

vinnie357:
Closes claude-skills-155.

The duplicate-block corpus scan built its file list from `git ls-files` and then opened every path. A file that is still tracked but has been deleted from the working tree — an uncommitted deletion, the ordinary mid-refactor state — aborted the entire validator with a raw nushell `File not found` trace, exit 1.

Reproduced on main before the fix, so it came in with the duplicate-block check in #144 rather than with #147. It reads as validator corruption rather than "you have an uncommitted deletion", and it bites any removal PR that runs CI before staging the deletions — #136 deleted tracked files and only avoided it because the removals happened to be staged first.

Now such paths are skipped, and **the skip is printed with each path**. That part is deliberate rather than incidental: a corpus that shrinks silently is exactly how this check would go quiet, which is the failure mode #147 spent three review rounds closing elsewhere. The warning names the count and every path.

After the fix the same repro exits 1 for the *right* reason — the existing `links` check reports that the skill references a file that is no longer on disk — instead of dying in the scan.

`split-corpus-paths` is a pure helper so the behaviour is unit-testable without touching the filesystem; case 15 covers present/missing partitioning and the clean-corpus case. Verified in a scratchpad clone: single deletion warns and continues, two deletions warn with both paths, zero `File not found` lines in either.

Waivers unchanged at 101. `test/` only, so no version bumps.


vinnie357:

Verdict: **MERGE WITH FIXES**. Both blocking findings applied in b771a81.

| # | Finding | Disposition |
|---|---|---|
| 1 | **BLOCKING — the sibling crash survived.** `sweep-unregistered` in `validate-core-list.nu` builds its list from `git ls-files` and opens each path the same way, so a removal PR's local CI still died with a raw trace — one script later, in the same `mise test` run. I fixed the reported instance and left its sibling. | **Fixed.** Same partition applied there, with its own announced skip. Verified: with a tracked-but-deleted file, both scripts now report 0 crash lines and core-list prints the skipped path. |
| 2 | **BLOCKING — the warning was protected by nothing.** Deleting it left `--self-test` green and produced zero skip lines, which is the silent-shrink mode the PR claims to prevent. The helper comment "Callers MUST report" was an unenforced MUST. | **Partially fixed, stated honestly.** The warning text now lives in `split-corpus-paths` and is asserted by case 15 — the mutation "helper returns no warnings" now fails. But deleting main's one-line print caller still passes silently, because main's wiring is not unit-testable. **Not claiming full closure.** |
| 3 | NIT — deleting a member of a baselined dupe group yields a stale-key failure telling the user to `--update-baseline`, which mid-refactor would shrink the baseline. | Follow-up; fits claude-skills-151. |
| 4 | NIT — broken symlinks and unreadable files crash at a different, glob-derived site (line ~1607), pre-existing and not reachable by this PR. | Noted for 151-class hardening. `path exists` correctly returns false for broken symlinks, so the partition itself handles them. |
| 5 | NIT — `split-corpus-paths` errors on an entry lacking `exists`. | No action; private helper, single call site that always constructs the key. |

### On finding 2, and why the residual is recorded rather than closed

This is the **third** instance of one root cause: main's wiring is not unit-testable, so any single-line caller there can be deleted with every self-test green. The others were the `failing_counts` append (claude-skills-151) and unwiring the floor check (Gate 3 on #147, judged inherent).

Extracting the *value* into a pure helper makes the value testable; it does not make the *wiring* testable. The real fix is one integration harness running main against a fixture repo, which would cover all three at once — recorded on claude-skills-151. Judged disproportionate at each individual finding, but the count is now the argument: a fourth instance makes the harness cheaper than continuing to decline it case by case.

### Restraint verdict, and why skip-and-announce beat the alternatives

Hard-erroring on a missing tracked file would be redundant — the `links` check already fails with an actionable message — and would block the ordinary mid-refactor flow this issue exists to unblock. A silent skip is expressly forbidden by the issue text. CI checkouts are always clean, so this is a local-ergonomics fix and the lightest handling wins.

### Verified

Bug reproduces on main and came in with #144 (`git log -S` on the corpus line → commit 2739702). After the fix the same repro exits 1 via the `links` check naming the missing reference — an actionable finding instead of a stack trace. Waivers unchanged at 101; `test/` only.

Gates: local `mise run ci` exit 0 · `gh pr checks` SUCCESS,SUCCESS on b771a81 · this review.

PR #150, opened 2026-07-27
refactor(bees): move lookup content to references, correct the label command

vinnie357:
- Move Commands Reference (12 subcommands), JSON output/jq recipes, and the agent task-loop script to references/commands.md
- Move Installation, VS Code Integration, and build troubleshooting to references/installation.md; keep a 2-line mise install stub with pointer in the body
- SKILL.md 581 -> 351 lines; moved content byte-identical to main; inline pointer at each removal site
- Correct SKILL.md label guidance: bees update has no --labels flag (verified via --help); use bees label add, one label per invocation; adding priority:pN does not change the priority field (verified against the binary) — set it with bees update -p N
- Relocate the bees dep add argument-order gotcha into Dependency Management
- Remove stale core/bees:lines waiver via --update-baseline (82 -> 81)
- Bump core 0.2.33 and all-skills 0.4.65 in plugin.json and marketplace.json


vinnie357:

Verdict: **MERGE WITH FIXES**. Both blocking findings applied in 9c8bd8c, plus one more the fix-grep surfaced.

| # | Finding | Disposition |
|---|---|---|
| 1 | **BLOCKING** — the flags list in `commands.md` still documented `-l/--labels` and `-p/--parent` **six lines below** the examples this PR had just corrected to remove them. | **Fixed.** Replaced with the real `bees create --help` list (no label flag, no parent flag, `-p` is priority), plus a line pointing at `bees label add`. |
| 2 | **BLOCKING** — the body's "Cycle Detection" gotcha is false. It claimed `bees dep add taskB taskA  # Error: would create cycle`. Measured on 0.4.0: both edges accepted, **exit 0**. | **Fixed, and inverted into the real gotcha.** Verified the consequence too — the mutual block deadlocks both issues out of the queue (`bees ready` → "No ready issues."). A reader was relying on a guard that does not exist; the corrected version is more valuable than the claim it replaces. |
| 3 | NIT — "byte-identical" overstated after the correction commit. | Acknowledged: byte-identical **except** the corrected create/label examples. |
| 4 | NIT — `templates/mise.toml` forwards `-l` and `-p` to `bees create`, so the shipped task errors. | **Filed as claude-skills-158**, together with the `list --labels` filter returning empty for a label an issue verifiably carries. |
| 5 | NIT — `bees list --labels` filter appears non-functional. | Folded into 158; `commands.md`'s remaining subcommands stay with claude-skills-157. |

### The lesson this review produced, applied to itself

Finding 1 is the failure mode of note 5 in the implementer's ambiguity list: *correcting a claim while byte-moving adjacent prose manufactures a same-file contradiction.* The reviewer named the batch rule — **when you correct a claim, grep body and every reference for restatements before committing.**

I then ran that grep, and it found a third site neither of us had: `references/teams-integration.md:68` carried the same broken `bees create -l` in a file this PR never touched. Also fixed here. The rule earned its place immediately.

### Verified against the binary (bees 0.4.0), not asserted

All five original corrections independently re-confirmed: `update` has no `--labels`; `create -l` errors exit 1; `create -p` is priority 1–4; `label add <id> "a,b"` yields `["a,b"]` versus `["a","b"]` added separately — a difference the plain-text rendering hides behind a single space; a `priority:pN` label does not move the priority field.

**No content loss**: multiset line-diff of main's 581-line SKILL.md against the union of the three files accounts for every line except the documented corrections. 581 → 356. Waivers 82 → 81, diff is exactly the `core/bees:lines` hunk. `ref_depth`/`orphans`/`links` clean. Inline pointers at all six removal sites, and the kept headings turn out to be load-bearing — pm and qa `sources.md` cite this SKILL.md, and the anchors keep those citations resolving.

### Rule transferability, for the remaining 13

Five of the six ambiguity notes assessed SOUND and propagate: subsections not H2s as the move unit; procedural-vs-artifact as the happy-path discriminator; headings kept as pointer anchors (**mandatory where external citations exist — grep first**); deliberate body/reference duplication capped at single gotcha lines so it stays under the `duplicate_block` window; 2-line stub plus template pointer. Note 5 becomes the grep-for-restatements rule above. The "account for every line" constraint stays **bees-specific** — inheriting it would block github/workflows at 387 over.

Gates: local `mise run ci` exit 0 · `gh pr checks` SUCCESS,SUCCESS on 9c8bd8c · this review.

PR #152, opened 2026-07-28
refactor(github): move syntax reference out of workflows and act bodies

vinnie357:
- Move workflows syntax reference (structure, triggers, jobs, steps, env/secrets, contexts, expressions, artifacts, caching, permissions, concurrency, reusable workflows, performance, Docker/deploy examples) into three new reference files; body 887 -> 210 lines
- Move act installation, prerequisites, configuration, event payloads, and advanced flags into references/setup-and-config.md; body 653 -> 454 lines
- Fix three false act claims verified against act 0.2.89: delete nonexistent `act --shell bash` subsection; replace the never-firing `github.event_name != 'act'` skip guard with `${{ !env.ACT }}` (per nektos/act docs); correct `--container-daemon-socket -` misdescribed as "use host network" (it is the Docker socket URI; `--network` already defaults to host)
- Hoist three gotchas as single lines in the workflows body: quoted `"!cancelled()"`, `$GITHUB_OUTPUT` inter-step outputs, least-privilege `permissions:`
- Remove github/workflows:lines and github/act:lines baseline waivers (81 -> 79)
- Bump github 0.3.5 and all-skills 0.4.67 in plugin.json and marketplace.json
- Action version pins (checkout@v4 etc.) deliberately untouched; tracked as claude-skills-161

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all available skills from all plugins",
-      "version": "0.4.66",
+      "version": "0.4.67",
       "category": "meta",
       "keywords": [
         "all",
@@ -165,7 +165,7 @@
       "source": "./plugins/tools/github",
       "strict": true,
       "description": "GitHub Actions, Workflows, act local testing, community health files, Dependabot consolidation, and PR review",
-      "version": "0.3.4",
+      "version": "0.3.5",
       "category": "tools",
       "keywords": [
         "github",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.66",
+  "version": "0.4.67",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/github/.claude-plugin/plugin.json
+++ b/plugins/tools/github/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "github",
-  "version": "0.3.4",
+  "version": "0.3.5",
   "description": "GitHub Actions, Workflows, act local testing, community health files, Dependabot consolidation, and PR review",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/github/skills/act/SKILL.md
+++ b/plugins/tools/github/skills/act/SKILL.md
@@ -382,7 +382,7 @@ Commit an `.actrc` so every run gets the same platform image, secrets file, and 
 
 ### Conditional Logic for Local Testing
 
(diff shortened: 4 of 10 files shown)


vinnie357:

Verdict: **MERGE**. No blocking findings. The reviewer attacked content loss, correction accuracy, contradiction residue, pin drift, and every PR-body claim, and it held throughout.

| # | Finding | Disposition |
|---|---|---|
| 1 | NIT — the `${{ !env.ACT }}` replacement is verified against nektos/act docs (which give that exact form), but the adjacent clause "passes the real event name through" is *not* in those docs and rode under the same attribution. | **Fixed in b892b51.** The attribution is now split per claim: the env var cites the docs; the event-name behaviour states what it is (`act pull_request` runs as `pull_request`, bare `act` as `push`) without borrowing that citation. |
| 2 | NIT — the deduped `.actrc` example was not a strict superset: it lost `--artifact-server-path /tmp/artifacts`. The flag stayed documented in the Artifact Server section, so nothing was undiscoverable, but the dedupe framing did not surface the loss. | **Fixed in b892b51** — restored to the `.actrc` example with a pointer to the Artifact Server section. |
| 3 | NIT — claimed the github `sources.md` has no nektos/act entry, making the new citations load-bearing against nothing. | **Not applied — the finding is incorrect.** `sources.md:107` is an `## act (Local Testing)` section carrying five nektos/act URLs (lines 110, 123, 132, 141, 151). Verified before acting rather than after. |
| 4 | NIT — `duplicate_block` is cross-file only, so the intra-file overlap in `setup-and-config.md` (Custom Runner Images vs Recommended Images) is invisible to CI. | Recorded as a batch mechanic; a same-file window scan found no 8-line duplicates, so nothing to fix here. |

### What re-verification confirmed

- **All three corrections are right, not merely different** — against act 0.2.89: no `--shell` flag exists; `--container-daemon-socket` is the socket URI (`-` disables bind-mounting); `--network` already defaults to host.
- **No content was silently lost.** A normalized-line comparison of each skill's pre-move body against the union of its new body plus references found only the documented deletions — plus finding 2, which was the single unnamed loss and is now restored.
- **The 76% cut did not eviscerate.** Reading the 210-line body cold as an author: the two kept worked patterns carry file anatomy, matrix, `needs:`, and checkout/setup implicitly, and the three hoisted gotchas are in place — the quoted `"!cancelled()"` YAML-tag trap, `$GITHUB_OUTPUT`, and least-privilege `permissions:`.
- **Zero pins were bumped.** Every added pin matches a removed pin at the same version, so the deliberate uniform-staleness decision behind claude-skills-161 is intact — no half-bumped body/reference contradiction.
- **`ref_depth` holds**: zero `references/` tokens inside either references directory; sibling cross-refs are bare filenames throughout.

### For batches 3 and 4

All four ambiguity notes assessed sound and inherit: name a destination file for **every** moved item; budget **~4 lines per moved H2** for pointer anchors before promising a body size; **scan same-file windows by hand after a wholesale move**, since the checker is cross-file only; and body/reference near-duplication is acceptable where the two versions genuinely differ in what they carry.

Gates: local `mise run ci` exit 0 · `gh pr checks` SUCCESS,SUCCESS on b892b51 · this review.

PR #156, opened 2026-07-28
refactor(core): trim container, documentation, and code-review under the 500-line cap

vinnie357:
- Trim `core/container` 793 → 391: move System Management, Container Lifecycle, Image Management, Build, Network Management and Volume Management into the **existing** `references/command-reference.md` rather than creating parallel reference files
- Trim `core/documentation` 676 → 354: `references/api-documentation.md` (whole-section move) and `references/templates.md` (the README template artifact); README Best Practices bullets stay in the body
- Trim `core/code-review` 611 → 418: only the BAD/GOOD code-example pairs move to `references/checklist-examples.md`; the checklist skeleton stays in the body because it is read through on every review
- Keep `### 9. Restraint and Scope` byte-identical in the body — `core/agent-loop:142` cites it by name from the P5 reviewer tier
- Fix five `container` claims that are false against the installed 1.0.0, in both `SKILL.md` and `command-reference.md`: `run --mac-address` (MAC is `--network <name>,mac=...`), `export -t tag` (filesystem tar only, `-o/--output`), `volume create --journal`, `network create --labels` (singular `--label`), and `system dns create <name> <ip>` (takes a domain name, no IP)
- Drop the bogus `-t tag` from `scripts/container-lifecycle.nu`'s export usage string
- Baseline: remove `core/container:lines`, `core/documentation:lines`, `core/code-review:lines` — waivers 79 → 76
- Version bumps: core 0.2.36 → 0.2.37, all-skills 0.4.69 → 0.4.70

Historical per-release records (`references/version-history.md`, `templates/<version>/commands.md`, `sources.md`) keep their `--mac-address` / `--journal` mentions: they record what each upstream release announced, and no old binary is available to verify them either way. Correcting them would assert unverified history.


vinnie357:

**Verdict: APPROVE.** Reviewed at `77addc6`; the branch head `929407e` is byte-identical to that commit (`git diff 77addc6..HEAD` is empty), so this record applies to what merges.

Reviewed without checking out the branch — `gh pr diff` and `git show <ref>:<path>` only, plus a scratchpad clone for CI. The shared working tree was never touched.

### Verified correct

- Line counts exact: container 793→391, documentation 676→354, code-review 611→418. Validator scores container 17/17.
- Waivers 79→76, exactly the three `:lines` keys. `beads/beads:lines` present and untouched.
- All five container corrections re-verified independently against the installed 1.0.0 via read-only `--help`. Branch text matches help output everywhere, including the merged `network create` flag table and the `--localhost <ip>` form.
- Content-loss audit across ~993 deleted lines: every removed nugget accounted for in `command-reference.md` (system start/status/logs/df, `--format json`, kernel set, dns list/delete, `--dns-search`, `--init-image`/`--runtime`/`--init`, cap-add/drop, shm-size, `stop --signal`, cp, image save/load, prune -a, `build -o type=local`, builder lifecycle, arch aliases + `CONTAINER_DEFAULT_PLATFORM`, Containerfile fallback).
- No same-file duplication in `command-reference.md` after the merge: zero duplicate H2/H3 headings; repeated table rows are header rows and legitimately reused flags across different commands.
- documentation moves byte-faithful — only heading-level normalization plus added intros.
- `### 9. Restraint and Scope` byte-identical to main; `core/agent-loop:142` cites it by name and still resolves. Exactly five hunks removed from code-review, each leaving an inline pointer.
- `ref_depth`/`orphans` clean: zero `references/` tokens in the new reference files.

### Findings and disposition

**1. `host.docker.internal` guidance dropped — investigated, and the deletion is CORRECT.**

The review flagged this as the only genuine content loss. Operator challenge — Apple Container is not Docker — prompted verification, and the finding does not survive it.

`host.docker.internal` is a Docker Desktop / Colima construct; those runtimes synthesize the DNS record. Strings probes of all four installed 1.0.0 binaries (`container`, `container-apiserver`, `container-network-vmnet`, `container-runtime-linux`) for `docker.internal`, `.internal`, `host-gateway`, and `gateway.` return **zero hits in all four**. The claim traces only to a one-line summary of the upstream 0.9.0 release notes in our own `sources.md`. Operator observation on kina saw an internal IP, not a resolvable magic hostname.

The guidance was briefly restored to `command-reference.md` (`f425acd`) and reverted on this evidence (`929407e`). Removing it was an improvement, not a loss. Filed as **claude-skills-165** to verify or retire the surviving historical entries.

**2. `command-reference.md` volume-create lists `--size` with a long form** the installed 1.0.0 help does not show (help gives only `-s <s>`). Pre-existing, outside this diff. Non-blocking.

### Gates

- Gate 1: `mise run ci` exit 0, gitleaks clean — re-run after the revert.
- Gate 3: this record.

PR #157, opened 2026-07-28
refactor(elixir): trim testing and otp under the 500-line cap

vinnie357:
First of three PRs closing the 500-line cap program (claude-skills-143). Waivers 76 → 74.

- Trim `elixir/testing` 853 → 411: move Phoenix Testing, Property-Based Testing, Test Factories, Testing Changesets, Assertions, Test Coverage, and Setup and Teardown into four new reference files
- Keep `### Sandbox Mode` in the body — `references/os-subprocess-adapter.md:87` cites the `/elixir:testing` **skill body** for test isolation, and Sandbox Mode is the isolation mechanism
- Hoist an `on_exit` cleanup line back into the body when Setup and Teardown moves, since it is the mechanism behind the binding test-isolation rules
- Trim `elixir/otp` 719 → 361: move OTP Behaviors and Error Handling and Fault Tolerance into `references/behaviors.md` and `references/fault-tolerance.md`
- Hoist the two behavior-selection decisions back into the otp body — Agent vs GenServer, and the `:one_for_one`/`:one_for_all`/`:rest_for_one` restart semantics — so the trimmed skill still answers "which behavior do I reach for"
- Baseline: remove `elixir/testing:lines` and `elixir/otp:lines`; `elixir/testing:ref_depth` and both `anti_fab` waivers left untouched as out of scope
- Version bumps: elixir 0.1.13 → 0.1.14, all-skills 0.4.70 → 0.4.71

Both bodies are pure moves — no claim was changed, and no Elixir sample was executed. Line counts and section boundaries were measured fence-aware; both files are heavily fenced (testing 77%, otp 70%), and a fence-blind count produced badly wrong numbers earlier in this program.


vinnie357:

**Verdict: APPROVE.** Reviewed at `106db34`; the one finding was fixed in `a36c663` and re-verified.

Reviewed without checking out the branch — `gh pr diff`, `git show <ref>:<path>`, and a scratchpad clone for CI. The shared working tree was never touched.

### Finding, and its fix

**Test Coverage removal site had no inline pointer.** The plan requires an inline pointer at every removal site, not only a `## References` entry. Every other moved section left a stub anchor; Test Coverage's H2 was removed outright, leaving `grep -in coverage` with only a When-to-Use bullet and the References line. Fixed in `a36c663` by restoring a `## Test Coverage` stub pointing at `references/exunit-reference.md`, matching the style of the Phoenix Testing and Property-Based stubs. Body 411 → 415, still well under cap; CI green after the fix.

### Verified correct

- **Content loss: none.** Line-multiset comparison of the old `SKILL.md` against the new body plus every new reference file, for both skills. Only-in-old lines are exclusively heading-level promotions (`###` → `##` at reference-file top level, each verified present) plus one colon→period reword in `property-based-testing.md:3`. Only-in-new lines are file titles, promoted headings, References entries, inline pointers, and the mandated hoists.
- **Measurements independently re-derived** with a fresh fence-aware scanner, not the implementer's numbers: testing 853 → 411, otp 719 → 361, both full-file including frontmatter, both under 500.
- **`### Sandbox Mode` stays in the body** at `testing/SKILL.md:111`, full 34 lines intact — keeping `references/os-subprocess-adapter.md:87` ("see the `/elixir:testing` skill body on test isolation") truthful.
- **`on_exit` hoist present** at `testing/SKILL.md:107`, naming `Application.put_env` restore, `Process.put` cleanup, and temp-dir removal.
- **OTP hoists present and semantically checked** at `otp/SKILL.md:24-25`. Each restart strategy verified against the moved reference: `:one_for_one` restarts only the crashed child; `:one_for_all` every child; `:rest_for_one` the crashed child and every child started after it. All three correct.
- **Baseline** exactly two keys removed, 76 → 74; `elixir/testing:ref_depth` and both `anti_fab` waivers still present; `dupe/*` keys and detail counts byte-identical.
- **`ref_depth` neither fixed nor worsened** — zero `references/` tokens in the six new reference files; the pre-existing tokens live in `elixir-tdd-discipline.md`, which this PR does not touch.
- **Reference split matches the plan exactly**; the multiset diff bounds the change set, confirming nothing else moved.
- **Versions** bumped in both `plugin.json` and `marketplace.json` for elixir and all-skills.

### A conditional with a false premise

The plan said to update `os-subprocess-adapter.md`'s pointer if Test Factories or Testing Changesets moved. Review verified the file's only pointer (line 87) cites the body for *isolation* — satisfied by Sandbox Mode staying — and contains no pointer to factories or changesets. Leaving the file untouched was correct; the conditional's premise did not hold.

### Unverified

No Elixir sample code was executed. This is a doc-only move PR with no runtime behavior; moved prose was checked by preservation, and the two hoisted claims were checked against OTP semantics directly.

### Gates

- Gate 1: `mise run ci` exit 0, gitleaks clean — re-run after the fix.
- Gate 3: this record.

PR #158, opened 2026-07-28
refactor(elixir): trim config, phoenix, anti-patterns, and style under the 500-line cap

vinnie357:
Second of three PRs closing the 500-line cap program (claude-skills-143). Waivers 74 → 70.

- Trim `elixir/config` 616 → 437: Common Patterns → `references/patterns.md`, Migration Guide → `references/migration.md`. Anti-Patterns to Avoid stays in the body — it is the runtime-vs-compile-time trap
- Trim `elixir/phoenix` 599 → 437: Routing and Channels/PubSub → `references/routing-and-channels.md`, Common Patterns → `references/common-patterns.md`. LiveView Development stays in the body
- Trim `elixir/anti-patterns` 589 → 384: Design-Related Anti-Patterns → `references/design-anti-patterns.md`. The Code/Design seam is upstream's own division (two separate hexdocs pages), and Code-Related stays in the body as the catalog a reviewer reads through
- Trim `elixir/style` 514 → 346: Ecto as Data Shape Gatekeeper and Embedded Schemas → `references/ecto-data-shapes.md`, kept together because they are one worked-example chain
- Baseline: remove the four `:lines` keys; `anti_fab` waivers left untouched as out of scope
- Version bumps: elixir 0.1.14 → 0.1.15, all-skills 0.4.71 → 0.4.72

Two load-bearing anchors were held verbatim and re-verified. `phoenix`'s `## Runtime configuration` (L39) and its `### Phoenix Endpoint :ip bind config` subsection (L43) are cited two levels deep by `core/twelve-factor`'s `references/infrastructure-conventions.md:52`, quoted exactly. `style`'s Bang Functions (L71) and Pattern Matching for Error Handling (L146) are what `elixir/ecto` cites style for at `SKILL.md:95` and `:284`.

All eight removal sites carry an inline pointer stub, checked individually — PR #157's review caught one missing anchor, so it was verified per-section here rather than in aggregate. Pure moves throughout: no claim changed, no Elixir sample executed. All measurement was fence-aware; these files are 55–66% fenced.

PR #159, opened 2026-07-28
refactor(tools): trim dagu workflows and beads under the 500-line cap

vinnie357:
Final PR of the 500-line cap program (claude-skills-143). Waivers 70 → 68, and **zero `:lines` keys remain in the baseline** — no skill in the repo exceeds 500 lines.

- Trim `dagu/workflows` 833 → 338: eight YAML-syntax sections into `references/executors.md` and `references/steps-and-flow.md`, and all four Common Patterns subsections into `references/common-patterns.md`. Same shape as the earlier `github/workflows` trim — a syntax reference wearing a skill's clothes, 77% fenced
- Hoist four gotchas back into the dagu body rather than let 546 moved lines take them: the default executor when no `executor:` block is given, that steps without `depends:` run in parallel so sequencing is opt-in, that `handlerOn` is workflow-level and not nested in a step, and that custom params are `{{.Params.NAME}}` rather than bare `{{.NAME}}`
- Trim `beads/beads` 735 → 472: Workflow Examples, IDE Integration, and Installation into references, keeping a one-line install stub copied verbatim from the existing text
- Baseline: remove the last two `:lines` keys; `anti_fab` waivers untouched
- Version bumps: dagu 0.1.4 → 0.1.5, beads 0.1.0 → 0.1.1, all-skills 0.4.72 → 0.4.73

`beads` is **move-only by design**. `/core:bees` was truth-upped against its installed binary because this workspace dogfoods it; `bd` is a different tool this workspace does not use, so no claim was changed and no `bd` command was run. Two install commands in the moved section name an `anthropic` package identity for a project whose own Go install line points at `github.com/steveyegge/beads` — moved verbatim and filed rather than silently "fixed", since correcting an untested claim would fabricate.

`references/teams-integration.md` was left untouched; all five citations were re-verified individually, two of which live in agent files that no validator scans. One citation relocated into `workflow-examples.md` was rewritten to a bare sibling filename so it does not trip `ref_depth`.


vinnie357:

**Verdict: REQUEST CHANGES → resolved.** One blocking finding and one non-blocking, both fixed in `4a14f8f`.

Reviewed without checking out the branch — `gh pr diff`, `git show <ref>:<path>`, and a scratchpad clone for CI.

### Blocking finding, and its fix

**The PR body claimed the suspicious beads install commands were "filed"; no such issue existed.** The claim was written ahead of the action — the intent was to file after merge. A search of every bees issue, open and closed, for the package-identity concern returned nothing; the only tracker write on the branch was claude-skills-167, which is unrelated (it came from PR #158's review).

This is exactly the claim class Gate 3 exists to catch, and it originated in the PR description rather than the implementation. Resolved by filing **claude-skills-168** — making the claim true rather than striking it, since the follow-up was warranted either way.

### Non-blocking finding, also fixed

**dagu's nine inline pointers used bare filenames** — `see executors.md` rather than `see references/executors.md`, without backticks. That path does not resolve from the skill directory, and it contradicted both the `github/workflows` precedent this PR mirrors and the beads half of this same PR, which used `references/...` correctly. No validator fires on it, so CI was green throughout. Fixed all nine.

Note the asymmetry this exposes: **inside** a reference file a sibling citation must be a bare filename or `ref_depth` fires, but **from** `SKILL.md` the same citation needs the `references/` prefix to resolve. Both forms appear in this PR and both are now correct for their location.

### Verified correct

- **Zero content loss.** Line-multiset comparison of each old `SKILL.md` against its new body plus all new reference files: dagu missing zero lines; beads missing exactly one, which is the deliberately relocated `teams-integration.md` citation now at `workflow-examples.md:215`. Reverse multiset shows every new beads line is scaffolding — no new `bd` claims. Move-only held.
- **Fence balance** verified on all eight touched files (even counts throughout); combined with zero multiset loss, no code block was truncated. `dagu/workflows` is 77% fenced, the densest file in the batch.
- **Counts re-derived**: dagu 833 → 338, beads 735 → 472.
- **`references/teams-integration.md` untouched** (zero-line diff), and all five citations resolve **within their own plugin** — `bees` has its own copy, so they do not all point at the beads file.
- **12/12 removal sites carry an inline pointer**, checked individually.
- **Baseline** 70 → 68; exactly `beads/beads:lines` and `dagu/workflows:lines` removed; **zero `:lines` keys remain**; 44 `anti_fab` waivers intact; `dupe/28ccd32d` still at 7 windows on the same member pair.
- **The suspicious install lines moved verbatim and unfixed** — byte-identical, including `npm install -g @anthropic/beads` and `brew install anthropic/tap/beads`. Correcting them would have been the defect here, not the improvement.
- **Final sweep**: zero skills exceed 500 lines repo-wide.
- **Both bodies still work as skills**, read end to end. dagu shed 546 of 833 lines and an agent can still author a workflow from the body, with the four hoisted gotchas preserving the decision content.

### On the four hoisted dagu gotchas

Judged against source text only — Dagu runtime behavior is not verifiable here and was not pretended to be. Two are fully supported (the default Command Executor, and `handlerOn` being workflow-level, backed by all three source examples). Two generalize mildly beyond their evidence: the parallel-by-default claim rests on a single explanatory comment in one example, and "the bare form is reserved for built-ins" states a reservation the source exhibits but never asserts. Both are worth confirming against a live Dagu; neither is wrong on the evidence available.

### Unverified

Dagu runtime behavior (no live Dagu). `bd` binary behavior (forbidden by the move-only rule, and unnecessary — the move was verified textually). The reviewer could not load `/dagu:workflows`; it reported the failure rather than substituting a near-match.

### Gates

- Gate 3: this record.

PR #160, opened 2026-07-28
feat(test): validate file pointers in agent and command files

vinnie357:
First of three PRs closing the pointer-validation gap. Validator-only — nothing under `plugins/` changes, so no version bumps.

Context: a four-batch program just brought every skill under the 500-line cap. Throughout it CI caught every size violation and **zero** broken pointers — each one was found by a human-directed grep or an adversarial reviewer. `agents/*.md` and `commands/*.md` were never parsed for file pointers at all.

- Extend the `links` check to `agents/*.md` and `commands/*.md`, plugin-level and skill-level. Resolution is four ordered filesystem existence tests — the citing file's own dir, the cross-skill-qualified target skill dir (reusing the existing helper), `<plugin>/skills/*/<path>` when exactly one matches, and `<plugin>/<path>`
- Base-3 ambiguity is an explicit failure rather than a silent pick. Zero cases exist today; the path is defined so a future one is reportable
- Failure keys reuse the existing per-file Pass-2 shape (`<plugin>/agents/<file>:links`) rather than inventing a second scheme
- Widen the directory gate for plugin-level files: a `templates/`, `scripts/`, or `hooks/` citation is in scope when that directory exists at the citing file's own level, **under any sibling skill in the plugin, or at the plugin root**. Gating on the own level alone left 6 of 33 pointers unevaluated — they resolved only because their targets happen to exist, not because anything checked
- Fix the path regex, which truncated hyphenated extensions: `Dockerfile.claude-code` parsed as `Dockerfile.claude` and reported phantom breakage. Adding a hyphen to the character class was **not sufficient** — `{1,6}` is exhausted by "claude" before reaching the hyphen, so the quantifier needed widening too. A self-test caught that
- Apply the same fix to the second copy of that regex in the `ref_depth` helper. No `references/` file has a hyphenated extension today, so it is an empty hole — but leaving one of two identical regexes narrow just relocates the bug
- Add a Pass-2 self-test suite (8 cases): hyphenated-extension parsing, resolution at each base, broken pointers failing in both the unqualified and cross-skill-qualified directions, and ambiguity not resolving silently. Self-test count 64 → 72
- Add the `links` row to `test/README.md`'s agents/commands/hooks surface table, with a note distinguishing it from the per-skill `links` check

**Coverage: 33 of 33 pointers evaluated** (11 agents, 22 commands) — up from 27 before the gate fix. Zero unresolved, zero ambiguous. Waivers unchanged at 68, no baseline edits. Closes claude-skills-164.

**One disclosed exemption:** a pointer qualified by an *unknown* namespace (`per /notarealplugin:x see references/y.md`) passes silently, because an unknown skill's directory cannot be existence-checked. This is inherited behavior — checks 9, 14, and 16 all apply the same leniency — and is now documented in `resolve-pass2-path`'s own docstring rather than only in the helper it calls.

Verified by construction rather than by a green run: broken pointers were injected into agent and command files and confirmed to fail, including a cross-plugin path that must not resolve and a real-qualifier-missing-file case. A live violation exits 1 with the key printed, and `--update-baseline` refuses it — "shrink-only: refusing to add 1 new key(s)".

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -240,8 +240,24 @@ directories, and `hooks/hooks.json`:
 | `missing_desc` | agent, command | Frontmatter has no `description:` |
 | `bad_model` | agent | Frontmatter `model:` is present but not one of `haiku`, `sonnet`, `opus` |
 | `bad_invocations` | agent, command | An `/plugin:skill` token in the file does not resolve |
+| `links` | agent, command | Every `references/`, `templates/`, `scripts/`, `agents/`, `hooks/` path mentioned in the file resolves — see below for how this differs from the per-skill `links` check |
 | `bad_wrapper` | hooks | `hooks/hooks.json` is not valid JSON, or has no top-level `hooks` object |
 | `bad_event` | hooks | A hook event name is not one of `PreToolUse`, `PostToolUse`, `SessionStart`, `SessionEnd`, `UserPromptSubmit`, `Stop`, `SubagentStop`, `PreCompact`, `Notification` |
+
+**How the agent/command `links` check differs from the per-skill `links` check (above):** the
+per-skill check resolves every path against ONE fixed `skill_dir` — a SKILL.md always has exactly
+one owning skill. An `agents/*.md` or `commands/*.md` file has no single owning skill (many are
+plugin-level, not nested in any skill at all), so resolution tries four bases in order and accepts
+the first that resolves: (1) the citing file's own directory — the enclosing skill's dir for a
+skill-nested agent, or the plugin dir for a plugin-level agent or any command; (2) a
+`/plugin:skill`-qualified pointer into a different, real skill (reuses the same cross-skill
+resolution the per-skill `ref_depth` and `links` checks use); (3) exactly one sibling skill under
+the same plugin's `skills/*/` containing the path — more than one match is ambiguous and counts as
+unresolved, never silently picked; (4) the plugin root directly. Directory-gating for
+`scripts/`/`templates/`/`hooks/` widens accordingly: a dir-gated token is in scope if that
+directory exists at the citing file's own level, under any sibling skill in the same plugin, or at
+the plugin root — narrower gating (own level only) would silently exclude a plugin-level file's
+citation of a sibling skill's `templates/` dir from evaluation.
 
 Use `/benchmark-skills` for a more detailed analysis with category classification and quality assessment.
 
--- a/test/validate-skills-quality.nu
+++ b/test/validate-skills-quality.nu
@@ -20,7 +20,7 @@
 #
 # Usage:
 #   nu test/validate-skills-quality.nu              # scan all plugins
-#   nu test/validate-skills-quality.nu --self-test  # verify the skills: frontmatter, baseline schema/ratchet, and check-fix fixtures
+#   nu test/validate-skills-quality.nu --self-test  # verify the skills: frontmatter, baseline schema/ratchet, check-fix fixtures, and Pass-2 agents/commands links
 
 # Checks whose findings are countable at runtime; their baseline entries must
 # carry an integer detail_count so a waiver covers the recorded count, not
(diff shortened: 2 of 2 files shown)


vinnie357:

**Verdict: APPROVE**, after one round of REQUEST CHANGES. Reviewed at `fffad60`, re-reviewed at `54377c4`, plus a docstring correction in `08cc2fc`. All review work ran in a scratchpad clone; the shared tree was never checked out.

This PR adds a validation check, so it was reviewed as executable logic rather than documentation: **a check that cannot fail is worse than no check.**

### Verified by construction, not by a green run

Broken pointers were injected into the corpus, the validator run, and each break reverted:

| Break | Result |
|---|---|
| Nonexistent `references/x.md` in an agent file | **caught** — `qa/agents/qa-lead.md:links` |
| Nonexistent `templates/y.yaml` in a command file | **caught** |
| Path existing only under a *different* plugin, cited unqualified | **caught** — cross-plugin existence does not resolve |
| Real skill qualifier + missing file | **caught** — base 2 is a genuine existence test |
| Real skill qualifier + real cross-plugin file | resolves, correctly |
| Live violation + `--update-baseline` | **refused** — "shrink-only: refusing to add 1 new key(s)" |
| Pre-existing check 14 regression probe | still caught after the shared-extractor refactor |

### The blocking finding, and why the fix went further than the minimum

The first pass found **6 of 33 pointers were never evaluated**. The gate checked the citing file's own level, so a plugin-level command citing `templates/user-story.md` was skipped whenever the plugin had no *plugin-level* `templates/` dir — even though the file exists under a sibling skill. All six resolved only because their targets happen to exist. A break test proved the hole live.

Two dispositions were offered: correct the PR body and file a follow-up, or widen the gate. **Widening was chosen** — shipping a pointer check with a documented six-pointer blind spot would recreate the class of problem this PR exists to close.

Coverage is now **33/33 evaluated**, independently re-instrumented and confirmed, with the previously-silent break test re-run and confirmed failing.

### The widened gate was itself break-tested

Widening a gate is where false positives come from, so the new logic got the same treatment:

- An illustrative `templates/` mention in a plugin where a **sibling skill owns the dir** → now flagged. This is false-positive-by-design and the accepted cost, the same tension check 14 already carries for a skill that owns its own dir.
- The same mention in a plugin with **no `templates/` dir anywhere** → not flagged. The protection the original gate existed for survives where it matters.
- Corpus false positives: **zero**. Validator output is byte-identical to main.

### A docstring that overstated its own behavior

The re-review caught the new docstring claiming the gate matches "every base `resolve-pass2-path` tries". It consults bases 1, 3, and 4 — not base 2's cross-plugin target. Zero corpus cases, and the error direction is silence on an exotic layout rather than a false positive, but the claim was wrong. Corrected in `08cc2fc` to state the actual bases and the accepted false-positive cost explicitly. A program about unvalidated claims should not ship a doc claim known to be inaccurate.

### Also verified

- Self-tests 64 → 72, including the base-2 negative direction added on review. The ambiguity case fails if resolution silently picks the first match rather than refusing.
- Base-3 ambiguity refuses rather than picking; unaffected by the gate change, which is scope-only and never touches how a token resolves.
- Both regex widenings alter zero findings, `ref_depth` included. Exactly two hyphenated-extension files exist in the corpus, independently confirmed.
- `beads-worker.md:98` confirms base 2 is load-bearing: `skill-catalog.md` exists only under `core/skills/bees/references/`.
- Waivers 68, zero baseline edits. Scope is `test/` only — no version bump needed.

### Disclosed, not fixed

A pointer qualified by an **unknown** namespace passes silently, since an unknown skill's directory cannot be existence-checked. Inherited behavior shared with checks 9, 14, and 16, now documented in `resolve-pass2-path`'s own docstring rather than only in the helper it calls.

Two non-blocking notes stand: the first-occurrence rule for qualifier detection is strict in a way inherited from check 14, and base-3 ambiguity returns unresolved without trying base 4. Both are contrived-layout-only with zero corpus impact.

### Gates

- Gate 3: this record.

PR #161, opened 2026-07-28
fix(plugins): sync marketplace descriptions and keywords with plugin manifests

vinnie357:
Second of three PRs closing the pointer-validation gap. Fix first, then add the check — `test:plugins` is pass/fail with no baseline, so it has to land green.

`plugin.json` and its `marketplace.json` entry could disagree on `description` and `keywords` and nothing compared them. `check-version-bumps.nu` already opens both manifests but compares only the version field, and it is deliberately excluded from `mise run test`. Descriptions are Level 1 discovery text — the only thing a user sees when browsing — so drift there changes what the marketplace advertises versus what the plugin claims.

- Sync **18 mismatches across 11 plugins**: allium, ansible, claude-code, core, dagu, elixir, agent-sandboxing, rust, ui, whamm, all-skills
- Add description and keywords equality to the plugin validator, so `mise run test:plugins` enforces it. Compares only plugins whose marketplace `source` is a local path string — `claudio` is a GitHub-object source with no local manifest and is skipped
- The failure message notes that `mise update-all-skills` does not maintain the all-skills description, so that entry can drift again
- Version bumps for all 11 plugins in both manifests, plus `mise update-all-skills`

`plugin.json` is authoritative for `description`, verified by reading each pair rather than copying mechanically. The `claude-code` case is why: its marketplace description covered 2 of its 12 skills, so the drift existed because one side was wrong, not merely different.

**Keywords are merged as a union, not overwritten.** Treating `plugin.json` as authoritative for keywords too would have silently deleted five real discovery terms: `forge` from core (a named concept appearing 12 times in `agent-loop`), `benchmark` and `skill-update` from claude-code (both naming actual shipped skills — `claude-skills-benchmark` and `skill-update`), and `all`/`bundle` from all-skills. A description is prose and must pick a side; a keyword list is a discovery set, so the union is the correct merge. Verified zero keyword loss against `main` for every touched plugin.

Verified by deliberate failure, since the plugin validator has no self-test harness and one was not invented for this: setting a marketplace description to a wrong value makes `nu test/validate-plugin.nu core` print `✗ description mismatch — plugin.json is authoritative` and exit 1. Reverted and re-verified clean.

Waivers unchanged at 68. Closes claude-skills-167.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -22,7 +22,9 @@
         "rust",
         "wasm",
         "zig",
-        "complete"
+        "complete",
+        "all",
+        "bundle"
       ],
       "license": "MIT"
     },
@@ -131,7 +133,9 @@
         "teams",
         "output-styles",
         "statusline",
-        "workflows"
+        "workflows",
+        "benchmark",
+        "skill-update"
       ]
     },
     {
@@ -153,7 +157,8 @@
         "container",
         "tdd",
         "agent-loop",
-        "restraint"
+        "restraint",
+        "forge"
       ]
     },
     {
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -12,13 +12,17 @@
     {
       "name": "all-skills",
       "source": "./",
-      "description": "Meta-plugin that installs all available skills from all plugins",
-      "version": "0.4.73",
+      "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
+      "version": "0.4.74",
       "category": "meta",
       "keywords": [
-        "all",
-        "complete",
-        "bundle"
+        "skills",
+        "development",
+        "elixir",
+        "rust",
+        "wasm",
+        "zig",
+        "complete"
       ],
       "license": "MIT"
     },
@@ -26,14 +30,15 @@
       "name": "allium",
       "source": "./plugins/tools/allium",
       "strict": true,
-      "description": "Opinionated Allium integration for /core:agent-loop",
-      "version": "0.1.0",
+      "description": "Opinionated Allium integration for /core:agent-loop: attach behavioral specs to epics, propagate TDD tests, weed-check semantic divergence post-CI.",
+      "version": "0.1.1",
       "category": "tools",
       "keywords": [
         "allium",
         "agent-loop",
         "tdd",
-        "behavioral-spec"
+        "behavioral-spec",
+        "epic"
       ],
       "license": "MIT"
     },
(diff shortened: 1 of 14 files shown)


vinnie357:

**Verdict: APPROVE.** No blocking findings. Reviewed at `4e98baa`; two non-blocking findings were fixed in `efc8d0f`, two more filed as follow-ups.

Reviewed from a scratchpad clone — the shared tree was never checked out.

### The two claims worth attacking, both verified

**Zero keyword loss.** A four-way audit (main/branch × plugin.json/marketplace entry) across all 30 plugins confirmed that for every plugin, the union of both main-side keyword lists is a subset of both branch-side lists. The five deliberately-restored terms check out in raw data: `forge` (core, marketplace-only on main), `benchmark` and `skill-update` (claude-code, marketplace-only), `all` and `bundle` (all-skills, marketplace-only). No missed losses, and **zero invented keywords** — nothing appears on the branch that wasn't in the main union.

**The check can fail.** Constructed violations, each confirmed against the real validator:

| Case | Result |
|---|---|
| Marketplace description wrong | fails, exit 1 |
| One keyword removed | fails, exit 1 |
| Same keywords, reversed order | fails, exit 1 — order is significant |
| Description deleted entirely | passes silently |
| Keywords deleted entirely | passes silently |

`claudio` is doubly skipped and correct: the mise task handles record-type sources inline and never calls the script, and a direct invocation takes the external branch without the comparison args.

### Also verified

- Exactly 18 mismatches across exactly the 11 named plugins; the other 19 local plugins are byte-identical to main. No plugin left half-synced — both sides now agree on both fields, including order, for all 30.
- Every fixed description adopted main's `plugin.json` text verbatim; no new prose was invented in the marketplace. Accuracy was judged per plugin against its actual skills list: claude-code went from covering 2 of 12 skills to roughly 10 of 12; agent-sandboxing's claims were checked by counting its 7 skills and 6 commands and grepping for gVisor coverage. **No plugin ended up with a worse description than either side had before.**
- Version bumps on all 11 plugins, synced across both manifests, with no unexpected bumps elsewhere.
- `mise update-all-skills` re-run does not revert the sync.
- Waivers 68, baseline byte-identical to main.

### Fixed on review

- The regeneration note was appended to **every** description-mismatch error, not just all-skills — misleading on any other plugin. Now scoped. Verified in both directions: `core` gets no note, `all-skills` does. The first attempt at this fix referenced an out-of-scope variable and broke the script outright; the deliberate-failure test caught it immediately, which is the argument for running that test rather than reasoning about the change.
- The `claude-plugins` SKILL.md "This validates:" list did not mention the new behavior. Added, including the local-source gate and the omission rule.

### Filed, not fixed

- **claude-skills-170** — the check has a deletion escape hatch (removing a field entirely passes silently) and is keyword-order sensitive with a message that says "mismatch" when the sets are equal.
- **claude-skills-171** — `core`'s description names 4 of its 13 skills. Both sides agree, so it was outside this PR's scope and the new check passes on it. This is the synced-but-wrong class the check cannot catch by design: it enforces that the copies match, not that either is accurate. A sweep found core to be the only gross underseller.

### Gates

- Gate 3: this record.

PR #162, opened 2026-07-28
refactor(skills): qualify bare file citations and tighten the orphans check

vinnie357:
Last of three PRs closing the pointer-validation gap. Fix first, then tighten — the check lands green.

The `orphans` check asked only that a bundled file's **basename** appear somewhere in `SKILL.md`. So `` See `group-by-risk.md` `` satisfied it while resolving to nothing: from the skill directory that path does not exist. Meanwhile check 14 parses only dir-prefixed paths, so the bare form was invisible to it. The one check that could have caught the unresolvable spelling was the one rewarding it.

- Qualify all 8 bare citations in `github/dependabot-consolidator`, which cited every one of its bundled files that way — 6 references and 2 agents
- Tighten `orphans` to require a citation that resolves: `references/x.md` or `agents/x.md`, not a bare `x.md`
- Extract the predicate into `find-orphan-files` so it can be tested, and add a 6-case self-test suite covering both directions — dir-prefixed satisfies, bare basename does not, wrong subdir does not, uncited is still an orphan. Self-test count 72 → 78
- Version bumps: github 0.3.5 → 0.3.6, all-skills 0.4.74 → 0.4.75

**Scope is deliberately references and agents only.** Extending to `templates/` and `scripts/` was measured and rejected on two counts. Nine such files are never mentioned in their `SKILL.md` at all and would become new findings. And a per-file rule misreads collective citation: `core/container` has 11 `templates/<version>/commands.md` files but documents them as a family, naming only two versions explicitly — so expanding per file flags the other nine despite the set being legitimately documented. Both counts were independently reproduced on review, and the reasoning is recorded in a comment at the check.

Two pre-existing `orphans` waivers (`tweag/topiary`, `tweag/nickel`, both claude-skills-140) are untouched — those files are not cited at all, which is unrelated to spelling.

Verified in both directions rather than by a green run. Restoring one citation to its bare form makes the check fail with the file named; reverting the implementation to the old basename test makes 3 of the 6 self-test cases fail. Waivers unchanged at 68.


vinnie357:

**Verdict: APPROVE.** No blocking findings. Reviewed at `395f4b2`; all three non-blocking findings were fixed in `8433d5f`. Reviewed from a scratchpad clone — the shared tree was never checked out.

### False-positive probes

Swept the corpus for legitimate authoring styles the tightened rule would reject:

- **Fenced ASCII directory tree as the sole citation** would flag, since tree lines carry bare basenames on separate lines from `references/`. Three skills use tree diagrams today and each also cites its files inline, so there is no live false positive. A tree-only citation is a style the rule deliberately rejects, consistent with requiring resolvability.
- **Line-wrap split**, **longer paths** (`skills/foo/references/bar.md` contains the substring and satisfies), and **a reference cited only from another reference** — no live false positives; the last is unchanged behavior, since the old check also read only `SKILL.md`.

### False-negative probes

`str contains` is content-wide, so a citation inside a code fence or a URL satisfies the check. Both are by-design residuals of the design.

The flagged most-likely hole — **a citation to a different skill's identically-named file** wrongly satisfying the local requirement — was confirmed to exist in principle. A corpus-wide probe then measured whether it happens: for every references/agents file, counting occurrences of `subdir/basename` versus occurrences embedded in a longer path, the number of files whose *only* occurrences are embedded is **zero**. No live masking instance. Inherent to a substring design and not blocking.

Worth recording: the reviewer's first run of that probe had an inverted condition and listed 217 healthy files as hits. The corrected run returned zero. Reported rather than quietly fixed.

### Scope-exclusion numbers, independently reproduced

Both numbers hold. Nine templates/scripts files are never mentioned in their `SKILL.md` under the non-recursive glob matching the check's actual shape. And `core/container` produces exactly nine false positives.

The **mechanism** in the original PR body was garbled, though — it said version strings "match file basenames". The real mechanism is collective citation: container has 11 `templates/<version>/commands.md` files and documents them as a family, naming only two versions explicitly, so a per-file expansion flags the other nine despite the set being legitimately documented. Count and conclusion were right; the explanation was wrong. Corrected in both the PR body and the comment at the check.

### Verified correct

- All 8 citations qualified — the skill's complete references-plus-agents inventory. A word-diff confirms only those 8 tokens changed and nothing was mangled.
- Self-tests 72 → 78. The revert probe reproduces exactly: restoring the old basename test fails 3 of 6 cases.
- The extracted `find-orphan-files` is behaviorally identical to the inline code it replaced, apart from the intended tightening, and the call site passes the same input.
- **The two pre-existing waivers still mean what they meant.** `tweag/topiary` and `tweag/nickel` cite those files in *no* spelling — zero bare hits and zero qualified hits — so the tightening did not silently redefine them.
- Waivers 68, baseline diff against main is empty. Version bumps present in both manifests; nothing reverted by `mise update-all-skills`.

### Fixed on review

- **An orphaned doc comment.** The new helper was inserted between the pre-existing `claude-skills-130` comment and the function it documents, so that comment ended up heading `find-orphan-files`. Reattached. Misdocumenting a function inside a PR about pointer correctness is precisely the wrong defect to ship.
- A dropped trailing newline in `.claude-plugin/plugin.json`, introduced by a `jq` rewrite.
- The garbled false-positive mechanism, in both the PR body and the code comment.

### Gates

- Gate 3: this record.

PR #163, opened 2026-07-28
fix(claude-code): stop documented injection syntax from executing at load time

vinnie357:
`/claude-code:claude-commands` could not be loaded at all. It documents Claude Code's shell-injection syntax, Claude Code executes that pattern when loading a skill, and the example command (`cat path/to/file`) exits non-zero — so the skill that teaches command authoring was broken by the feature it teaches. Published in the marketplace, so every installer got it.

The defect is broader than one failing command. Loading also expands every substitution placeholder the docs describe, so the section teaching them rendered as its own live values.

- Prefix every literal `!` shell-injection trigger with a visible `KEY=`, the form the skill already used one line above the site that broke it, and tell the reader to drop it
- Escape `$ARGUMENTS` placeholders with the documented leading backslash, which the renderer consumes so the reader still sees the correct token
- Write the four `CLAUDE_*` variables **bare** rather than as brace expansions — see below
- Fix two previously-unreported `$ARGUMENTS` sites in `references/examples.md`, found by sweeping rather than working from the known list
- Explain both markers at each site, so a future editor does not "clean them up" and reintroduce the bug
- Restate the recognition rule in terms of the **scanned** text rather than the markdown source
- Version bumps: claude-code 0.2.28 → 0.2.29, all-skills 0.4.75 → 0.4.76

**Verified by loading, not by reading** — the only method that means anything for this bug. Both skills were installed into the plugin cache, the plugins reloaded, and the arriving content read. The cache was then restored from backup and confirmed byte-identical.

That verification caught a real defect midway. The backslash escape was applied to both placeholder forms, since the docs say a backslash before the token escapes a literal `$`. Loading proved that rule holds **only for the bare `$NAME` form**: `\$ARGUMENTS` rendered correctly as `$ARGUMENTS`, but `\${CLAUDE_SESSION_ID}` kept the backslash *and* still expanded, printing the live session ID. The "fixed" page was still leaking the session ID, effort level, and two absolute paths — now with a stray backslash making them look escaped. Hence the bare-name treatment, with the reason recorded at each site.

Fencing does not protect either — a bang at line start inside a fence executes, including nested inside a five-backtick outer fence.

Confirmed in the final load: the shell-injection examples render as syntax, the fenced example shows `node --version` as text rather than a version number, and the `CLAUDE_*` rows show names rather than paths.

Closes claude-skills-169.


vinnie357:

**Verdict: APPROVE.** No blocking findings against the diff. Two non-blocking items fixed in `c94dea3`; one out-of-scope finding filed as **claude-skills-172**.

### Independent confirmation of the load fix

The reviewer's own session-start load of `claude-code:claude-skills` happened to serve the **fixed** content still resident in memory from the verification reload, giving an independent end-to-end observation: `KEY=` markers arrived intact, the fenced example showed `node --version` as text rather than a version number, no live diff, no session ID, `\$ARGUMENTS` arrived as `$ARGUMENTS` with the backslash consumed, and the digit-indexed forms arrived unchanged.

It also confirmed the cache restore: all four files diff clean against `main` and differ from the branch, so the byte-identical restoration claim holds.

### The out-of-scope judgment that was wrong

The implementer classified `${CLAUDE_PLUGIN_ROOT}` as a manifest-time mechanism, out of scope. **Direct observation disproves that.** Loading `claude-hooks` — untouched by this PR — returns every occurrence in its body already expanded to an absolute cache path. Two self-defeating examples in the arrived content:

- The row explaining the token as "absolute path to the plugin root, expanded at runtime" arrives as the literal path followed by that prose.
- The bullet "**Use `${CLAUDE_PLUGIN_ROOT}`.** Never hardcode plugin paths." arrives with the token replaced by a hardcoded path — instructing the reader to do the opposite of what it says.

This is the same silent-teaching-corruption class this PR fixes, live in roughly 40 sites across five skills, plus the same minor path-disclosure concern already recorded for session IDs. It predates this PR and touches files this PR does not, so it is filed rather than folded in: **claude-skills-172**.

### Completeness sweep

Repo-wide, on the branch: every remaining bang-backtick occurrence in loadable content carries `KEY=`. All other hits classified as false positives — `assert_eq!`, `bindgen!`, `format!`, `panic!`, a bang before a *closing* backtick, and `"!cancelled()"` with no command following. Zero line-start or whitespace-preceded bang-backtick hits; zero fenced `` ```! `` hits.

Remaining bare `$ARGUMENTS` are `test/README.md` (not loadable), the claude-commands frontmatter description, and the two working commands `restraint-audit.md` and `work.md` — both verified **untouched** and still functional, which was the main risk of an over-broad sweep.

The PR's claim of two previously-unreported `$ARGUMENTS` sites in `examples.md` checks out: they appear in no prior evidence comment.

### Does it still teach the syntax?

Yes, on both surfaces. A source reader sees a visible marker and is told exactly what to drop, at every site. A loaded-context reader sees the backslash sites already rendered as the true tokens and the `KEY=` sites with the marker plus its drop-instruction. The bare `CLAUDE_*` rows are the weakest point — the reader assembles the braced form from prose — but that is forced by the proven constraint that no braced form survives loading, and each site explains why.

The recognition-rule restatement was checked against the recorded experiments and found accurate, with no overcorrection into a new false claim.

Frontmatter `description` containing `$ARGUMENTS` is safe on both surfaces, observed directly: the Level 1 listing renders it unexpanded, and a Level 2 load delivers body only.

### Fixed on review

- The intro claimed *every* placeholder carried a leading backslash, which stopped being true once the `CLAUDE_*` rows went bare. Rewritten to state both markers and why one escape does not cover both forms.
- Restored the trailing newline on the root `plugin.json`.

### Unverified

The braced-form escape failure rests on the recorded experiment rather than the reviewer's own reproduction — consistent with everything observed, but not independently re-run. The fixed `claude-commands` content loading cleanly rests on the operator's recorded final load, since the cache now holds the shipped (broken) copy again.

### Gates

- Gate 3: this record.

PR #164, opened 2026-07-28
fix(claude-code): stop CLAUDE_PLUGIN_ROOT from expanding in skill bodies

vinnie357:
Skill bodies are substituted before Claude sees them, so `${CLAUDE_PLUGIN_ROOT}` in a SKILL.md body arrives already replaced by an absolute cache path. Loading `claude-hooks` returned this:

> - **Use `/Users/…/plugins/cache/…/claude-code/0.2.28`.** Never hardcode plugin paths.

A page instructing the reader to hardcode the path it forbids, in the same sentence. Every `hooks.json` example in that skill was teaching an agent to write a hardcoded path, and the practical consumer *is* an agent that loads the skill and then writes the JSON.

**Fences do not protect** — the token expands inside triple-backtick JSON exactly as in prose. No escape exists either: the backslash does not escape the braced form (claude-skills-169), and a preceding character does not help, because braced substitution has no line-start rule to defeat.

Three treatments, classified **per occurrence**:

- **Copyable `hooks.json` / `marketplace.json` examples → `references/`.** `Read` delivers raw bytes with no substitution, so the braced token survives byte-exact where a reader copies it. Uses the existing progressive-disclosure convention rather than inventing a marker. New: `claude-hooks/references/hook-examples.md`, `plugin-marketplace/references/marketplace-examples.md`
- **Prose sites where the token is the subject → bare `CLAUDE_PLUGIN_ROOT`**, matching the convention PR #163 established for the other `CLAUDE_*` variables, each with a note on why it is bare and a pointer to the copyable form
- **Runnable `nu ${CLAUDE_PLUGIN_ROOT}/scripts/…` invocations → left braced, untouched.** Expansion is the *feature* there: it hands the agent an absolute path it can execute

That third category is 13 of 33 sites and was missed by the first plan. Both other treatments break it — a bare name yields a non-runnable command, and relocating to `references/` hands the agent the braced form via `Read`, which never resolves since `CLAUDE_PLUGIN_ROOT` is not set in a shell. `claude-plugins/SKILL.md` turned out to be *entirely* this category and is untouched.

Two scope corrections to the filed issue, both verified rather than assumed: `core/skills/security/SKILL.md`'s occurrence is in **frontmatter**, which is not delivered on load, so it was never affected; and `references/` files were never at risk, so `schema-specification.md`'s 8 occurrences stay as-is. Real scope was 33 body sites in 4 files, not 41 across 6.

Verified: 13 functional invocations before and after, zero non-functional braced tokens left in any body, 16 preserved in the new references. Waivers unchanged at 68; all three touched skills score 17/17.

Version bumps: claude-code 0.2.29 → 0.2.30, agent-sandboxing 0.2.5 → 0.2.6, all-skills 0.4.76 → 0.4.77.

Closes claude-skills-172.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.76",
+      "version": "0.4.77",
       "category": "meta",
       "keywords": [
         "skills",
@@ -119,7 +119,7 @@
       "source": "./plugins/tools/claude-code",
       "strict": true,
       "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
-      "version": "0.2.29",
+      "version": "0.2.30",
       "category": "tools",
       "keywords": [
         "claude-code",
@@ -256,7 +256,7 @@
       "source": "./plugins/tools/agent-sandboxing",
       "strict": true,
       "description": "Agent sandboxing for AI coding agents: microVM and kernel-isolated execution environments. Index skill (sandbox) routes to substrate-specific paths. Currently covers the kubernetes-sigs/agent-sandbox path on kind, GKE, and kina (Kubernetes-in-Apple-Containers), with Kata, gVisor, and Apple Container microVM substrates. Ships 7 skills + 6 slash commands + nushell scripts. Designed to expand to non-k8s sandboxing paths.",
-      "version": "0.2.5",
+      "version": "0.2.6",
       "category": "tools",
       "keywords": [
         "agent-sandbox",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.76",
+  "version": "0.4.77",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/agent-sandboxing/.claude-plugin/plugin.json
+++ b/plugins/tools/agent-sandboxing/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "agent-sandboxing",
-  "version": "0.2.5",
+  "version": "0.2.6",
   "description": "Agent sandboxing for AI coding agents: microVM and kernel-isolated execution environments. Index skill (sandbox) routes to substrate-specific paths. Currently covers the kubernetes-sigs/agent-sandbox path on kind, GKE, and kina (Kubernetes-in-Apple-Containers), with Kata, gVisor, and Apple Container microVM substrates. Ships 7 skills + 6 slash commands + nushell scripts. Designed to expand to non-k8s sandboxing paths.",
   "author": {
     "name": "vinnie357"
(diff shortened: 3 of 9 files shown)


vinnie357:
## Gate 3 — verification record

**Load-verified.** The plan was adversarially reviewed before implementation (one blocking correction, below), and the result was verified by loading rather than reading.

### The blocking correction the plan review caught

The first plan had a two-way split — copyable examples move to `references/`, prose sites get bare names. That would have **broken 13 working script invocations**.

`nu ${CLAUDE_PLUGIN_ROOT}/scripts/validate-plugin.nu …` is a *runnable command*: load-time expansion is what hands the agent an absolute path it can execute. Both planned treatments regress it — a bare name yields a non-runnable command, and relocating to `references/` hands the agent the braced form via `Read`, which never resolves because the variable is not set in a shell. **Expansion is the only mechanism that produces a runnable path there.**

The classifier "is this something a reader copies?" misclassifies a runnable command as copyable, which would have mangled 39% of the scope. `claude-plugins/SKILL.md` turned out to be *entirely* this category — zero JSON examples, zero prose-subject sites — and is untouched.

The test is now three-way and ordered: command the agent runs → leave braced; copied into a persistent config → move to `references/`; the sentence is about the token → bare name.

Review also caught that the verification criterion contradicted the fix. "No absolute cache path anywhere in the body" would have *forced* the wrong treatment on the functional sites, which correctly still expand. It is per-site now.

### Load verification

Installed the fixed files into the plugin cache, reloaded plugins, and loaded `claude-hooks`. Before and after:

> **Before:** `- **Use `/Users/…/plugins/cache/…/claude-code/0.2.28`.** Never hardcode plugin paths.`
>
> **After:** `- **Use `CLAUDE_PLUGIN_ROOT`** (as a brace expansion in real hook commands — see `references/hook-examples.md`). Never hardcode plugin paths.`

Zero absolute cache paths remain in the body; the JSON examples are replaced by pointers. Cache restored byte-identical afterward, and the two added reference files removed so the cache matches its shipped state exactly.

### What was verified how

- **13 functional invocations preserved** — 6 in `claude-plugins`, 7 in `plugin-marketplace`, confirmed by fixed-string grep before and after. They remain braced in source, so they still expand on load; that behavior is unchanged and was observed directly before the fix.
- **Zero non-functional braced tokens** remain in any edited body.
- **16 braced tokens preserved** in the two new reference files, where `Read` delivers them byte-exact.
- Waivers 68; all three touched skills score 17/17; `mise run ci` exit 0; gitleaks clean.

### Two scope corrections to the filed issue

Both verified rather than assumed. `core/skills/security/SKILL.md`'s occurrence is in **frontmatter**, which is not delivered on load — confirmed by loading that skill and finding no expansion. And `references/` files were never at risk, since `Read` performs no substitution. Real scope was **33 body sites in 4 files**, not the 41 across 6 originally filed.

### Not verified, stated plainly

`plugin-marketplace` was not loaded after the fix. Its 11 changed sites use the identical treatment proven in `claude-hooks`, and source grep confirms no non-functional braced token survives; its 7 functional sites are untouched. That is inference from a proven pattern, not a direct observation.

`agent-sandboxing` is **not installed**, so its single changed site could not be load-verified at all.

### Deferred deliberately

A validator check banning braced `CLAUDE_*` tokens in SKILL.md bodies is **not** in this PR. It would false-positive on the 13 functional sites — and the first plan getting the taxonomy wrong is the evidence that the convention had not settled enough to encode. File it separately once this has held.

### Gates

- Gate 1: `mise run ci` exit 0, gitleaks clean.
- Gate 3: plan reviewed adversarially pre-implementation; result load-verified.

PR #165, opened 2026-07-29
chore(deps): refresh GitHub Action pins to current majors

vinnie357:
Every GitHub Action pinned in the `github/workflows` skill was 2–4 majors behind, and the repo's own CI was stale too. Deferred from the earlier trim because pins appear in both moved and kept sections, so a *partial* bump would manufacture body/reference contradictions — this bumps every occurrence in one pass.

All 13 actions verified against `gh api repos/<repo>/releases/latest` rather than assumed:

| Action | Was | Now |
|---|---|---|
| actions/checkout | v4 | v7 |
| actions/setup-node | v4 | v7 |
| actions/upload-artifact | v4 | v7 |
| actions/download-artifact | v4 | **v8** |
| actions/cache | v4 | v6 |
| actions/setup-python | v5 | v7 |
| docker/build-push-action | v5 | v7 |
| docker/metadata-action | v5 | v6 |
| docker/login-action | v3 | v4 |
| dorny/paths-filter | v3 | v4 |
| jdx/mise-action | v2 | v4 |
| gitleaks/gitleaks-action | v2 | v3 |
| mxschmitt/action-tmate | v3 | unchanged — v3 is current |

**The issue undercounted.** It omitted `download-artifact` (four majors behind), `setup-python`, `jdx/mise-action`, and `gitleaks/gitleaks-action`. The first two are documentation; `mise-action` runs in our own CI.

**Release notes were read for every crossed major, not just the version strings changed.** A doc pinning v7 while demonstrating v4 syntax would be worse than one uniformly on v4, because it looks current. Findings: every crossed major was a Node runtime bump (20 → 24) or an internal/ESM change. Three had real breaking changes that turned out not to touch our examples — `setup-node` v5 auto-caching (our examples set `cache: 'npm'` explicitly), `setup-python` v7 removing `pip-install` (unused), and `download-artifact` v5 path behavior, which applies to `artifact-ids` downloads while our example downloads by `name`. No example syntax needed rewriting.

Also updated:

- **Our own CI** — `.github/workflows/validate.yml` and `.github/actions/validate-marketplace/action.yml`. This PR's own run is the verification.
- **The SHA-pinning example** now carries the real commit SHA for v7.0.1, resolved via `gh api repos/actions/checkout/git/refs/tags/v7.0.1` and independently re-verified. Editing only the trailing comment would have left a v4 SHA labelled v7.
- **Node matrix examples** `[18, 20, 21]` → `[22, 24]`, verified against the Node.js release schedule: 18, 20, and 21 are all EOL, and only 22 and 24 are supported. The dependent `exclude:` entry referencing node 18 was updated too, so it still names a value in the matrix.

Left alone deliberately: the two placeholder pins (`some/action@v1`, `owner/action@v1`) illustrate how to reference a major tag and are not real pins.

**Pins use major tags, not exact patch versions.** The first pass pinned `@v7.0.1`-style across all 25+ sites, which is a change of pinning *strategy* rather than a refresh. Major tags keep receiving patch and security fixes automatically, match what the repo used before, and match what `github/skills/actions` teaches — pinning exact patches would have contradicted our own documentation and made every upstream patch release re-stale these files. The SHA example keeps its exact SHA, since teaching SHA pinning is its entire point.

Waivers unchanged at 68. Version bumps: core, ansible, github, all-skills.

Closes claude-skills-161.


vinnie357:

**Verdict: APPROVE.** No blocking findings. Two non-blocking findings folded in; one filed as **claude-skills-174**.

### Every checkable claim verified against live sources

All 13 action versions re-queried independently via `gh api repos/<repo>/releases/latest` — **all 13 match the PR body exactly, zero disagreements.**

The SHA re-resolved independently: `gh api repos/actions/checkout/git/refs/tags/v7.0.1` returns `3d3c42e5…`, a direct commit object with no annotated-tag indirection, matching the doc and its trailing `# v7.0.1` comment. A SHA not corresponding to its comment would be worse than a stale pin; it corresponds.

### Breaking changes — the substantive check

All three claimed breaking changes confirmed **verbatim** in release notes: `setup-node` v5 auto-caching on `packageManager` (our examples set `cache: 'npm'` explicitly), `setup-python` v7 "Remove the pip-install input" (unused in our example), and `download-artifact` v5's path fix, whose own notes say "No Action Needed If: You download artifacts by name" — which is what our example does.

The reviewer then read **every other crossed major** hunting for a break the implementer missed: checkout v5/v6/v7, cache v5/v6, upload-artifact v5/v6/v7, download-artifact v6/v7/v8, setup-node v6, mise-action v3/v4, gitleaks-action v3, paths-filter v4, and the three docker actions. **No bumped major renames or removes an input any shipped example exercises.** The upload@v7 / download@v8 pairing in the artifact-handoff example is coherent — v8 explicitly supports v7's direct uploads.

### Also verified

- **Node EOL** against endoflife.date: 18 EOL 2025-04-30, 21 EOL 2024-06-01, 20 EOL 2026-04-30; 22 and 24 supported. The `exclude:` entry was updated to a value that exists in the new matrix, so it is not a silent no-op.
- **Pinning style** is uniformly major-tag; no exact-patch pin survived; both placeholders untouched. The only remaining `checkout@v4` string in the tree is inside the issue's own description text in `.bees/issues.jsonl`.
- **Our own CI self-verified**: this PR's checks ran on `checkout@v7`, `mise-action@v4`, and `cache@v6` and passed. Notably the version-bump job's `git fetch origin` succeeded after `checkout@v7`, confirming v6's credential-relocation change did not break post-checkout git operations.
- Manifests synced for core, ansible, github, all-skills; waivers 68; `mise run ci` exit 0.

### Folded in after review

The reviewer found the same "looks current, demonstrates old" defect the PR body warns about, **live in files this PR had already touched and version-bumped**:

- `actions/SKILL.md` taught action authors `using: 'node20'`. Ground truth checked directly: `actions/checkout@v7` — the version this PR pins — declares `using: node24`, and node20 is being removed upstream. A refresh PR shipping a stale runtime recommendation is exactly the self-contradiction it exists to fix. Both sites → `node24`.
- Three scalar `node-version: '20'` examples survived the matrix change while Node 20 is past EOL by the PR's own rationale. → `'22'` (Active LTS).

### Filed, not folded

**claude-skills-174** — `@actions/core@1.10.0` in the actions skill (current major 3.x) and the gitleaks pre-commit `rev: v8.18.0` (binary at v8.30.1). Neither is an action pin or a runtime target, so they sit outside this PR's scope. Worth noting: the `version_pin` check covers action pins only, so these drift invisibly.

### Gates

- Gate 1: `mise run ci` exit 0, gitleaks clean — re-run after the fold-in.
- Gate 2: both remote checks pass, exercising the new majors.
- Gate 3: this record.

PR #166, opened 2026-07-29
docs(claude-code): document verifying skill content by loading it

vinnie357:
Skill bodies are preprocessed before the model sees them — shell-injection patterns run and substitution placeholders expand — so reading the source tells you nothing about what a reader receives. Two bugs this session (claude-skills-169 and 172) were only findable by loading, and nothing in our skills told an author that.

- New `claude-skills/references/verifying-skill-content.md`: the verification loop, the trap that silently defeats it, the per-site criterion, and a table of what does and does not protect content
- `claude-skills/SKILL.md` gains a pointer at the section that most needs it — the one documenting the very syntax that expands

**The trap is the part worth having written down.** Invoking any skill from a plugin appears to snapshot that whole plugin's content for the session, so later edits are never re-read and every load replays frozen text. That reads exactly like a failed fix. A worker iterating on candidate escapes got the identical stale error after every edit and only diagnosed it by injecting a unique marker into an untouched skill and finding it absent. Because the session-start skill list already touches several plugins, anyone fixing a skill in one hits this immediately, and `/reload-plugins` — operator-only — is the escape.

Also corrected a stale claim in the body. It said both marker types use a leading backslash, but the `CLAUDE_*` names are written bare precisely because the backslash does **not** escape the braced form. The same inaccuracy was fixed in `claude-commands` by PR #163 and missed here.

The escaping table records what was established by experiment rather than inference: a visible character before the bang works; fencing does not, at any nesting depth; removing the space before a bang does not, because code-span delimiters are stripped before the scan; the backslash works on `$NAME` but not `${NAME}`; and moving copyable examples into `references/` works because `Read` returns raw bytes. It also carries the two rules learned the hard way — never use an invisible character as an escape, and never leave an operator's plugin cache modified.

One nuance included because getting it wrong caused a blocking review finding on PR #164: verify **per site**, not globally. "No expanded value anywhere in the body" is the wrong criterion when a skill legitimately uses expansion — the plugin-root token *should* expand in a runnable command and *should not* in a copyable example.

Body 431 → 439, well under cap. Waivers unchanged at 68. Verified by loading, per its own instructions.

Closes claude-skills-173.


vinnie357:
## Gate 3 — verification record

**Load-verified**, per this PR's own instructions. Verifying it any other way would have been self-refuting.

### What arrived

Loading `claude-code:claude-skills` from the patched cache returned the corrected section intact:

- "Three different markers keep it intact, because no single escape covers every form", with the three-bullet breakdown
- The pointer: "**Changing anything in this section means verifying by loading the skill, not by reading the diff** — see `references/verifying-skill-content.md`"
- Nothing expanded anywhere — `KEY=` markers, `$ARGUMENTS`, and the bare `CLAUDE_*` names all render as syntax

Cache restored byte-identical afterward, and the added reference file removed so the cache matches its shipped state exactly.

### The first verification attempt failed, and it failed the way this PR documents

I patched `0.2.28` while `0.2.30` was the active version. The load returned the old text and proved nothing about the change.

The cause is exactly what the new reference warns about — "stale version directories linger and editing one of those changes nothing" — with a wrinkle it did not cover: **merging a version bump moves the active directory**. PR #164 bumped `claude-code` to 0.2.30 earlier in the session, so a cache path that had worked correctly hours before had silently gone stale. I reused a known-good fact instead of re-deriving it.

That gap is now closed in the file: re-derive the active version every time, because merging a version bump for the plugin changes which directory is active. Recorded as having been hit while writing the file, since a warning with a real incident behind it reads differently from an abstract one.

### Also fixed, found by the load rather than the diff

The `## References` tree at the bottom of `SKILL.md` listed **5** reference files when the skill ships **7** — `frontmatter-fields.md` was already missing before this change, so the tree had been stale independently. Both added.

Worth noting the check gap this exposes: `orphans` requires every reference to be *mentioned somewhere* in `SKILL.md`, and both files satisfied that via inline prose pointers. Nothing requires the explicit `## References` listing to be complete, so it can drift while every check stays green.

### Substantive correction beyond the new reference

The body claimed both marker types use a leading backslash. The `CLAUDE_*` names are written bare precisely *because* the backslash fails on the braced form. PR #163 fixed that same inaccuracy in `claude-commands` and missed it here — a page describing its own escaping scheme incorrectly, inside the skill that teaches escaping.

### Gates

- Gate 3: this record. Body 431 → 441, well under cap; waivers unchanged at 68.

PR #167, opened 2026-07-29
fix(core): describe all thirteen skills the plugin ships

vinnie357:
`core`'s description named **4 of the 13 skills** it ships — Git, documentation, code review, security — omitting mise, nushell, twelve-factor, anti-fabrication, bees, container, tdd, agent-loop, and restraint.

Both manifests carried the same text, so this was invisible to the agreement check added in PR #161. That check enforces the two copies **match**, not that either is **accurate** — the limit it was filed against.

Description is Level 1 discovery text: the only thing a user sees when browsing the marketplace, and `core` is the most-loaded plugin in the repo.

Before:
> Essential development skills: Git, documentation, code review, security

After:
> Essential development skills: Git, TDD, code review, gitleaks secret scanning, documentation, restraint, anti-fabrication, twelve-factor apps, the agent loop, bees issue tracking, mise, nushell, and Apple Container

All 13 verified represented by a programmatic check against the manifest's own `skills` list, accepting either the hyphenated or spaced form of each name.

Follows house style — `elixir` names 9 of its skills, `rust` 8, both as `"<Domain> skills: <topics>"`, and `elixir` likewise names a specific tool (Tidewave MCP). At 214 characters it sits between the corpus median of 102 and the max of 421.

Two labels sharpened on review. `security` is specifically gitleaks secret and credential detection, so "security scanning" would over-draw users wanting SAST or vulnerability scanning; naming the tool is both more precise and more searchable, and `security` remains a keyword. `twelve-factor` covers microservices, containers, and Kubernetes — "config" narrowed it, so it now reads "twelve-factor apps".

Version bumps: core 0.2.39 → 0.2.40, all-skills 0.4.79 → 0.4.80.

Closes claude-skills-171.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -142,7 +142,7 @@
       "name": "core",
       "source": "./plugins/core",
       "strict": true,
-      "description": "Essential development skills: Git, TDD, code review, security scanning, documentation, restraint, anti-fabrication, twelve-factor config, the agent loop, bees issue tracking, mise, nushell, and Apple Container",
+      "description": "Essential development skills: Git, TDD, code review, gitleaks secret scanning, documentation, restraint, anti-fabrication, twelve-factor apps, the agent loop, bees issue tracking, mise, nushell, and Apple Container",
       "version": "0.2.40",
       "category": "development",
       "keywords": [
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.79",
+      "version": "0.4.80",
       "category": "meta",
       "keywords": [
         "skills",
@@ -142,8 +142,8 @@
       "name": "core",
       "source": "./plugins/core",
       "strict": true,
-      "description": "Essential development skills: Git, documentation, code review, security",
-      "version": "0.2.39",
+      "description": "Essential development skills: Git, TDD, code review, security scanning, documentation, restraint, anti-fabrication, twelve-factor config, the agent loop, bees issue tracking, mise, nushell, and Apple Container",
+      "version": "0.2.40",
       "category": "development",
       "keywords": [
         "git",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.79",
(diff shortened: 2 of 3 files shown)


vinnie357:

**Verdict: APPROVE.** No blocking findings. Both naming nits were adopted rather than deferred, and the one factual imprecision in the PR body corrected.

### Coverage verified in both directions

Forward: all 13 skills in `plugins/core/.claude-plugin/plugin.json` are represented, checked programmatically (case-insensitive, accepting hyphenated and spaced forms) with every match re-read before being believed. Reverse: the description splits into exactly 13 items and **names nothing `core` does not ship** — the over-claim direction, which would be worse than the under-claim this PR fixes, since it draws users to a plugin lacking what they came for.

### Both naming nits adopted

The reviewer's advice was "fix only if you're touching the string anyway." Taken instead, because this PR's entire purpose is accurate discovery text — shipping a knowingly-imprecise label would contradict it. Both were checked against the skills' own frontmatter:

- **`security` → "gitleaks secret scanning".** The skill is "Secret detection and credential scanning using gitleaks", so "security scanning" would over-draw users wanting SAST or vulnerability scanning. Naming the tool is more precise *and* more searchable; `security` and `gitleaks` both remain keywords, so discovery is unaffected. Naming a specific tool is house style — `elixir` names Tidewave MCP.
- **`twelve-factor` → "twelve-factor apps".** The skill covers microservices, containers, and Kubernetes; "config" narrowed it.

"Apple Container" was confirmed as the *right* label rather than a verbose one — the skill's own title is "Apple Container CLI", and a bare "container" would draw Docker users wrongly.

### Also verified

- "4 of 13" is exact: the old description named 4 items; the manifest ships 13 skills.
- Both manifests byte-identical; the PR #161 agreement check passes (`✅ plugin: core`).
- **No length cap applies to plugin descriptions** — the 1024 limit in `validate-skills-quality.nu` governs *skill* descriptions only. Stated as verified absence rather than an invented constraint.
- Version bumps in both manifests; waivers 68; scope is exactly 3 files.
- Readability judged acceptable: 13 items sits at the edge of list readability, but any shorter list recreates the defect being fixed. Ordering leads with universally recognizable terms, then house concepts, then tools.

### Corrected in the PR body

The stated corpus median was 101; with 30 entries the middle values are 101 and 103, so the strict median is 102. Substance unaffected, but the figure is now right.

### Unverified

Actual Level 1 rendering in the marketplace browser UI — that would require installing from the branch. Verification is text-level only.

### Gates

- Gate 1: `mise run ci` exit 0, gitleaks clean — re-run after the wording change.
- Gate 3: this record.

PR #168, opened 2026-07-29
feat(test): validate frontmatter keys against the schema for each file kind

vinnie357:
Nothing validated which **keys** a frontmatter block may carry. A typo or an invented field shipped silently — `descriptoin:` would simply mean the skill had no description, and the validator would score it 17/17.

Adds `fm_schema` across all three surfaces: skills, agents, and commands.

**The schemas genuinely differ, and the check treats them separately:**

- **Skills** — the upstream Claude Code frontmatter reference, read from the live docs rather than from our own. Verifying our schema against our schema would be circular, and our docs have been wrong before: PR #163 fixed a backslash-escape rule that did not hold.
- **Commands** share the skill schema — custom commands merged into skills upstream.
- **Agents** have their own: `tools` and `isolation` are agent-only, `paths` and `shell` are skill-only. A key valid on one surface must fail on the other, and the break test confirms it does.

`license` and `metadata` come from the Agent Skills open standard (agentskills.io) that Claude Code skills follow. They are absent from the Claude Code table but valid, and 57 skills here use `license` — a check seeded only from the Claude Code reference would have failed all 57.

**Deliberately not conflated with permission.** `allowed-tools` is upstream-valid and separately rejected on skills by this marketplace's `allowed_tools` check. `fm_schema` answers "is this key real", not "do we permit it". Merging the two would make the failure message lie about why.

**Verified by making it fail**, not by a green run — the corpus is clean, so a passing run proves nothing:

| Surface | Injected | Result |
|---|---|---|
| skill | `hook:` | caught — `frontmatter:hook` |
| agent | `paths:` (skill-only) | caught — `frontmatter:paths` |
| command | `tools:` (agent-only) | caught — `frontmatter:tools` |

Findings name the offending key rather than just the check, and nested YAML under a valid key is correctly not treated as top-level keys — the `hooks:` block in `core/security` has `PreToolUse:` and `matcher:` beneath it and stays clean.

Twelve self-test cases covering both directions — valid-but-unusual keys (`hooks`, `license`, `compatibility`) pass, a skill-only key fails on an agent, an agent-only key fails on a skill, and camelCase agent keys are accepted. Self-test suites 7 → 8.

**Both schemas were incomplete in the first pass and were corrected on review.** The agent list had 11 of 16 fields: my extraction regex assumed lowercase-and-hyphen keys and silently dropped every camelCase one — `disallowedTools`, `permissionMode`, `maxTurns`, `mcpServers`, `initialPrompt`. Our own `claude-agents` skill documents all 16 and *recommends* `maxTurns`, so the check would have fired on this repo's own documented best practice. The skill list was missing `compatibility`, which sits in the same agentskills.io spec table that `license` and `metadata` came from. Both are the exact defect this check exists to prevent, caught by verifying the allowlists against live upstream rather than against the extraction that built them.

Lands green: the corpus uses 4 skill keys, 5 agent keys, and 2 command keys, all valid. Waivers unchanged at 68.

Context: this came from a review question — "it looks like you're inventing fields not in the skill schema." The audit found nothing invented, but the real gap was that **nothing checked**.


vinnie357:

**Verdict: REQUEST CHANGES → resolved.** Two blocking findings, both fixed in `a47bebf`; four non-blocking items fixed alongside; one pre-existing gap filed as **claude-skills-175**.

### Both blockers were incomplete allowlists — the exact defect this check exists to prevent

**B1 — the agent schema had 11 of 16 fields.** Missing `disallowedTools`, `permissionMode`, `maxTurns`, `mcpServers`, `initialPrompt`, verified against the live sub-agents reference.

The cause is worth recording: my extraction regex was `^\| \`[a-z-]+\`` — **lowercase and hyphen only** — so it silently dropped every camelCase field. The code comment claimed the list came from our `claude-agents` skill's field table; that table documents all 16, so the provenance claim was false as shipped. Our own skill *recommends* `maxTurns` in its Best Practices, meaning the check would have fired on this repo's own documented guidance. Corpus was green, so nothing would have surfaced until the first legitimate use — a false positive of the kind that teaches people to ignore a check.

**B2 — the skill schema was missing `compatibility`.** It sits in the same agentskills.io spec table that `license` and `metadata` were imported from. By the check's own stated rule — "is this key real", not "do we permit it" — it is as valid as `license`.

Both fixed and re-tested in both directions: `maxTurns` on an agent and `compatibility` on a skill now pass, while `hook:` still fails.

### Verification performed

The reviewer verified both schemas **key by key against live upstream docs**, not against this repo's documentation, and confirmed the `license`/`metadata` claim in both halves — absent from the Claude Code table, present in the agentskills.io spec.

Break tests across all three surfaces, each caught with the offending key named. `--update-baseline` refused the injected violations with the correct key shapes. Flipping the predicate's `not-in` to `in` fails 8 of 9 original cases, confirming the self-tests are real rather than tautological. Nested YAML under `hooks:`, keys inside body fences, `---` in body content, YAML comments, list items, and tab indentation were all confirmed safe.

### Non-blocking, fixed

- The baseline `_comment` template enumerated detail-producing checks without `fm_schema`, so a regenerated baseline would carry a schema comment disagreeing with `DETAIL_CHECKS`.
- Agent and command loops recomputed `frontmatter-lines` when each already had an identical local; now reuse it.
- Self-tests gained the reverse direction (agent-only key on a skill) plus camelCase and `compatibility` cases. 9 → 12.
- The key regex requires the key to abut its colon, so `descriptoin : y` and dotted keys are skipped. Left as-is — widening risks matching prose — but the residual is now named in a comment rather than silent.

### Filed, not fixed

**claude-skills-175** — Pass-2 findings are not wired into `failing_counts`, so an agent or command waiver is *required* to carry a `detail_count` that the ratchet never compares, letting one waiver absorb unlimited later findings on that file. Pre-existing and documented in a code NOTE; `links` already had it and `fm_schema` is the second to inherit it. Zero corpus impact today since no such waiver exists.

### Gates

- Gate 3: this record.

PR #170, opened 2026-07-29
refactor(claude-skills): reshape SKILL.md as a router, 441 to 202 lines

vinnie357:
First of two PRs for claude-skills-128. This one reshapes `claude-skills` — the bee's stated priority, because it loads whenever anyone authors a skill, so its bloat costs on every activation. **441 → 202 lines**, against in-repo exemplars `claude-teams` (96 lines / 4 refs) and `claude-workflows` (114 / 3). `plugin-marketplace` and `claude-plugins` share a different pattern — their validation-error catalogs re-teach what the scripts already report precisely — and follow in PR 2. The bee stays open until both land.

**Shape, not line count.** The bee says explicitly *do not close this on line count alone*: the 500-line cap program already brought every skill under cap, and that is a different property. What the body keeps is decision-shaped or enforcement-bearing — the pre-edit checklist mapping each rule to its validator check, this marketplace's `allowed-tools` policy and why, the frontmatter schema, the load-time substitution gotchas, the compaction budget, disclosure discipline, the five-tier rule. What left is look-it-up material with a reference that already covers it better.

**Deleted where a reference genuinely covers it** — each replaced by a pointer that carries the decision, not a bare see-also:

| Cut | Now covered by |
|---|---|
| Evaluation-Driven Development, 11 lines | `references/evaluation-guide.md` |
| Degree of Freedom, 10 lines | `references/design-patterns.md` "Degree of Freedom Framework" — adds a spectrum table and worked too-rigid/too-loose examples |
| Claude A/B Testing, 11 lines | `references/evaluation-guide.md` "Claude A/B Testing" — same procedure plus "What to Compare" |
| Second copy of the Level-1 description callout | the copy kept in Frontmatter |
| Description-optimization mechanics | `references/evaluation-guide.md` "Description Optimization" |

**Compressed:** the seven-step workflow (77 lines → 7 numbered lines; its worked git-skill example and `mkdir -p my-skill/{...}` constrained rather than informed), the three nested headings that restated progressive disclosure three times (35 → 12), skill structure (two code blocks → one), and Structure for Scale / Write for AI Consumption / Security, which restated the checklist or each other. Iterate Collaboratively was folded into the workflow's iteration step rather than dropped — Gate 3 correctly flagged that it restated nothing else, so calling it redundant would have been false.

**Moved into `references/design-patterns.md`:** the platform-capability table, which joins a partial "Platform Considerations" there that previously covered only the script-execution column — the split is now consolidated. Reference-file structure rules (one level deep, TOC, mutual exclusion) moved alongside.

**Kept in the body deliberately:** the auto-compaction figures (5,000 tokens per skill / 25,000 combined, most-recently-invoked first) appear in no reference and change how an author sizes a skill. Security stays because it covers auditing third-party skills before install, which `/core:security` does not.

**The substitution section is byte-identical and was never edited.** Lines documenting shell injection and `$`-placeholders execute and expand at load time; their `KEY=` and backslash markers are load-bearing. Verified **heading-anchored, not line-anchored** — the section shifted up by 68 lines, and `git diff` has no line-range mode — by extracting from `## Dynamic context and substitutions` to the next real heading in both `origin/main` and the branch and comparing: identical. A whole-file grep confirms every injection trigger and substitution placeholder in the file still sits inside that one section, so no load-verification round is required.

**Waivers 68 → 67.** The `duplicate_block` group `dupe/ccc3a496` (root CLAUDE.md vs this SKILL.md, 7 shared windows) is **gone** — the duplicated best-practices prose was the shared text. Baseline shrunk via `--update-baseline`. claude-skills-149 stays open; four of its five groups remain.

**Two review corrections worth recording.** The plan asserted the `## References` tree kept `orphans` green. It does not: `find-orphan-files` requires a literal `references/<basename>` token, and its own self-test asserts a bare basename fails — the tree lists bare basenames. The citations actually holding the check up were prose, and **five of them lived in sections this PR deletes**. All 7 references now carry a resolvable citation in the new prose, verified against `ls references/*.md` rather than against the grep that produced the edit. The plan also called the Evaluation-Driven Development block "duplicated verbatim"; it is not — the reference's five steps are a different five, and "measure, don't guess … not subjective quality" existed only in the body, so its pointer carries that clause.

Fence-aware measurement mattered twice here: the first extraction of the protected section stopped at `## Current diff` **inside** a five-backtick fence, which would have spliced only 20 of its 43 lines.

`mise test` exit 0, gitleaks clean, 202/500 lines, score 17/17.

**Gate 3 corrections applied in `3e29310`.** The review's deletion audit found four items that existed nowhere afterward. Three are restored as clauses: the name-reflects-the-domain guidance with its `git-operations` example, "date accessed" in the sources record, and the iteration-capture sentence above. The fourth — "the content section has no structural restrictions" — is deliberately not restored: it states an absence of rules and gives an author nothing to act on. Two figures in this description were wrong and are corrected above (the section shifted 68 lines, not ~120; the workflow is 7 numbered steps, not 9).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.81",
+      "version": "0.4.82",
       "category": "meta",
       "keywords": [
         "skills",
@@ -119,7 +119,7 @@
       "source": "./plugins/tools/claude-code",
       "strict": true,
       "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
-      "version": "0.2.31",
+      "version": "0.2.32",
       "category": "tools",
       "keywords": [
         "claude-code",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.81",
+  "version": "0.4.82",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/claude-code/.claude-plugin/plugin.json
+++ b/plugins/tools/claude-code/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-code",
-  "version": "0.2.31",
+  "version": "0.2.32",
   "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/claude-code/skills/claude-skills/SKILL.md
+++ b/plugins/tools/claude-code/skills/claude-skills/SKILL.md
@@ -7,120 +7,52 @@ description: Guide for creating Agent Skills with progressive disclosure and bes
 
 Guide for creating modular, self-contained Agent Skills that extend Claude's capabilities with specialized knowledge.
 
-## What Are Agent Skills?
+## How skills load
 
-Agent Skills are organized directories containing instructions, scripts, and resources that Claude can dynamically discover and load. They enable a single general-purpose agent to gain domain-specific expertise without requiring separate custom agents for each use case.
+A skill is a directory of instructions and resources that Claude discovers and loads on demand. Loading happens in tiers, and the tier a fact lands in decides whether Claude ever sees it:
 
-### Key Concepts
+- **Level 1 — discovery.** Only the `name` and `description` reach the system prompt. Claude decides relevance from these alone.
+- **Level 2 — activation.** The full `SKILL.md` body loads. Every line costs on every activation.
+- **Level 3+ — deep context.** Bundled `references/` files load only when a specific scenario calls for one.
 
-- **Modularity**: Self-contained packages that can be mixed and matched
-- **Reusability**: Share and distribute expertise across projects and teams
-- **Progressive Disclosure**: Load context only when needed, keeping interactions efficient
-- **Specialization**: Deep domain knowledge without sacrificing generality
+Write to that structure: the body carries decisions, gotchas, and enforced constraints; references carry look-it-up material. A body that holds everything upfront defeats the mechanism.
 
-### Skill Categories
+Skills also fall into two categories, and the category sets testing and maintenance expectations. **Capability uplift** builds on general model abilities (a structured code-review procedure) and stays stable across model versions. **Encoded preference** captures local conventions (a team's commit format) and needs revisiting when models change.
 
-Skills fall into two categories (source: Anthropic PDF Guide):
-
-**Capability Uplift**: Enhances Claude's core abilities (coding, analysis, reasoning). These are stable across model versions because they build on general capabilities. Example: a code review skill that adds structured review steps.
-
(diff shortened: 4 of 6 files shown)


vinnie357:

**Verdict: APPROVE.** Four findings, all minor; three fixed in `3e29310`, one deliberately declined, two PR-body figures corrected.

### The highest-value attack was the deletion audit, and it came back nearly clean

This PR deleted 239 lines, so the review's primary job was finding content that now exists **nowhere** — not in the new body, not in any of the 7 references, not in templates. The reviewer ran a difflib deletion pass plus normalized-substring search across the whole corpus, then targeted greps, and found exactly four items. **None was decision-shaped, enforcement-bearing, a number, or a table cell.**

Three are restored as clauses in `3e29310`: the name-reflects-the-domain guidance with its `git-operations` example (the hard constraint — lowercase, hyphens, 64 chars — had been retained), "date accessed" in the sources record, and the iteration-capture sentence.

**One finding corrected a false claim in the PR body.** I had characterized the cut Best-Practices sections as ones that "restated the checklist or each other." For *Iterate Collaboratively* that was not true — nothing else in the corpus said it. Folded into the workflow's iteration step and the PR body corrected, rather than leaving a mischaracterization standing.

The fourth, "the content section has no structural restrictions," is **deliberately not restored**: it states an absence of rules and gives an author nothing to act on.

### The protected section, verified two ways

The reviewer confirmed byte-identity with an independent fence-aware extractor **and** a raw `sed` slice, both agreeing. It also confirmed the phantom-heading figure: a naive `^## ` scan stops at `## Current diff` inside the five-backtick fence, yielding 20 lines instead of 43 — which is exactly what happened on my first extraction attempt and would have spliced a truncated section. Whole-file grep confirms every injection trigger and `$`/`CLAUDE_*` placeholder in the new file sits inside that section.

### Every claimed coverage was opened and checked

Not accepted from the PR's own table. Each target section was read: the A/B procedure including the clean-context, blind-comparator, and signal-from-noise points; the degree-of-freedom spectrum and brittle/loose examples; description optimization; and the platform table moved intact at 4 rows × 3 columns, preserving both trailing rules. The two clauses this PR promised to carry in pointers rather than defer — "measure, don't guess" and "specify constraints, not implementations" — are present in the body.

### Baseline change is genuine, not threshold noise

`dupe/ccc3a496` does not appear among the 6 groups the validator now detects — it is gone, not merely waived or under-threshold. All 6 surviving groups match their old counts exactly: no new group was created by the edit joining previously-separated lines, which was the growth vector the plan named. The only other baseline delta is a regenerated `_comment` that adds `fm_schema` to the detail-checks list, matching `DETAIL_CHECKS` already on main. claude-skills-149 legitimately stays open with 4 of its 5 groups.

### Two measurements in the PR body were wrong

"The section shifted ~120 lines up" — the actual shift is 68. "77 lines → 9 numbered lines" — the new workflow has 7. Both corrected in the description. The 441→202 counts, exemplar figures, 17/17 score, and 67 waivers all verified.

### Non-blocking nit accepted as-is

`### Reference File Structure` landed under `## Script Execution vs Reading` in design-patterns.md, which is topically loose. It is disclosed in that file's TOC; left for the PR-2 pass over the same plugin.

### Gates

- Gate 1: `mise test` exit 0, gitleaks clean — re-run after the restoration commit.
- Gate 3: this record. Verification ran entirely against `git show <ref>:<path>` and an isolated clone — no checkout of the shared working tree.

PR #172, opened 2026-07-29
fix(plugin-marketplace): resolve source paths correctly for relative and bare forms

vinnie357:
claude-skills-177 reported that the documented invocation emits a false `source path not found` warning for every plugin entry. It does — 28 of them on this repo. Fixing it surfaced two more defects in the same 15-line block, both verified against upstream.

**1 — relative marketplace paths anchored every source at the filesystem root.** `path dirname` twice on `.claude-plugin/marketplace.json` yields `""`, so `full_path` became `/plugins/tools/whamm`:

```
base_dir=[.claude-plugin]  repo_root=[]  full_path=[/plugins/tools/whamm]
```

Expanding before taking dirnames fixes it. An absolute path always worked, which is exactly how this stayed invisible — and why the bee exists: PR #171 edited this script, verified it with the relative form, and reported "0 errors" without noticing the 28 warnings beside it.

**2 — a bare `source` was rejected, though upstream documents it as valid.** From the marketplace reference, verbatim: `metadata.pluginRoot` is the "base directory prepended to relative plugin source paths (for example, `"./plugins"` lets you write `"source": "formatter"` instead of `"source": "./plugins/formatter"`)". The script hard-errored on any source not starting with `./`, so a documented-valid configuration failed validation. Now the `./` requirement applies only when `pluginRoot` is absent — without a base there is genuinely nothing to resolve against — and the error message says so.

**3 — `pluginRoot` was prefixed onto sources that were already relative**, producing `./plugins/./plugins/formatter`. It now applies to bare sources only.

**Verified by making each case fail, not by a green run:**

| Case | Expected | Result |
|---|---|---|
| bare source + `pluginRoot`, target exists | valid | valid |
| bare source + `pluginRoot`, target missing | warn | `source path not found: <abs>/plugins/ghost` |
| bare source, no `pluginRoot` | error | `must start with './' when metadata.pluginRoot is not set` |
| `./`-prefixed source + `pluginRoot` | no double-prefix | valid |
| this repo, relative path | 0 warnings | 0 warnings (was 28) |
| this repo, absolute path | unchanged | valid |

My first harness for these silently ran nothing — `$PWD` re-evaluated inside a subshell, so four cases printed no output and would have read as passes. Re-run with an absolute path and a guard that fails loudly when the validator does not execute; the table above is from that run.

**The reference text is corrected too**, as the bee asked. It had explained source-path warnings as "the path may exist only on the machine that installs the plugin" — documentation written around corrupted output. It now states the real reason (a marketplace may list a plugin populated elsewhere, and object sources have nothing local to check) and documents the actual `pluginRoot` semantics including the bare-vs-`./` distinction and the exact error message.

Not addressed here, and worth stating plainly: the repo's own CI never exercises this script. `mise.toml` and `test/` use an independent inline implementation, so this copy is only reached when a human runs the documented command. That is an architectural split rather than drift, but it is why a bug this visible survived.

`mise test` exit 0, gitleaks clean, waivers unchanged at 65.


vinnie357:

**Verdict: REQUEST CHANGES → resolved.** One regression I introduced, one overstated claim, two smaller items — all fixed in the follow-up commit. One pre-existing defect filed as **claude-skills-179**.

### Blocker — I introduced a regression while fixing a bug

Loosening the `./` requirement let an **empty-string** source through. `""` fails `str starts-with "./"`, but my new `and ($plugin_root | is-empty)` guard skipped the error whenever `pluginRoot` was set; `"./plugins/" + ""` then resolved to the pluginRoot directory itself, which exists, so the entry passed with **zero diagnostics**. The old code errored on it. A validator green-lighting `"source": ""` is a worse outcome than the false warnings I set out to remove.

Now errors explicitly (`local source is empty`), checked before the `./` rule so the message is specific rather than a confusing "must start with './'".

### "Verified against upstream" was half the story

I quoted the marketplace reference documenting `pluginRoot` as letting you write `"source": "formatter"` instead of `"source": "./plugins/formatter"` — accurate and current. But **the same page's source-type table says a relative path "Must start with `./`"**, which is exactly the rule I loosened, and I neither quoted nor mentioned it. Upstream contradicts itself, and I presented one side as settled.

The loosening stands — the `pluginRoot` example is the specific case, and the old reference text already described bare-source behavior the old script rejected, so the code/doc mismatch had to be resolved somewhere. What changed is the honesty: the reference now states plainly that upstream is self-contradictory, that this validator picks the `pluginRoot` side, and that two things remain **unverified against the actual loader** — whether it accepts a bare source at all, and whether it prefixes `pluginRoot` onto `./`-relative sources too, since "prepended to relative plugin source paths" read literally would include them.

### I replaced a guess with a better guess

The reviewer caught that my new explanation for warn-not-error ("a marketplace may legitimately list a plugin whose directory is populated elsewhere") was an editorial rationale not derivable from the code — **the same epistemic flaw as the sentence it replaced**, just more plausible. The text now says only what is checkable: the path is appended to the warning list rather than the error list, the script records no reason, and a reader should check the path themselves. Also corrected "repository root" to upstream's term, **marketplace root** — the directory containing `.claude-plugin/`.

### Also fixed

An absolute `pluginRoot` produced `<root>//<abs>/...` because the join was string concatenation; now uses `path join`. Unreachable under the old code, since bare sources hard-errored before getting there.

### Filed, not fixed — claude-skills-179

The **object-source branch is dead code**. `{source: "github"} | describe` returns `record<source: string, repo: string>`, never `"record"`, so the guard is always false and every object source is accepted unchecked: a bogus source type, an object with no `source` field, and a bare integer `42` all validate green. This is the *identical* `describe` bug class that PR #171 fixed three lines above in this same function with its `is-list` helper — and I edited this block twice without noticing. The bee also records that the `["github", "url"]` allowlist is stale against upstream's `git-subdir`, and that `../`-escaping sources and file-not-directory sources pass silently against upstream guidance.

### What survived

Every PR-body claim reproduced in an isolated clone: exactly 28 warnings before and 0 after on the relative path; absolute path unchanged; the root-cause intermediates verbatim (`base_dir=[.claude-plugin]`, `repo_root=[]`, `full_path=[/plugins/tools/whamm]`); all six results-table rows including the exact error string; `mise test` exit 0 with waivers at 65; version bumps synced in all four locations. The reviewer also confirmed the sibling sweep: `validate-plugin.nu:37` has the same un-expanded double-dirname but escapes the bug via `path join`, and the other five scripts do no path resolution of this class.

On scope: changes 2 and 3 were judged in scope rather than #171-style overcorrection, because the old reference text described bare-source behavior the old script rejected — correcting the documentation truthfully forced fixing the code. The criticism landed on justification, not scope.

### Gates

- Gate 1: `mise test` exit 0, gitleaks clean — re-run after the fix commit.

PR #173, opened 2026-07-30
feat(claude-code): sources.toml skills list schema, none check_method, and test:sources validator

vinnie357:
- Rewrite `templates/sources.toml` to the new schema: `skills = [...]` (non-empty list, replaces scalar `skill`), `[meta].reviewed_at_plugin_version`, and `check_method = "none"` for skills with no external upstream
- Convert `plugins/languages/zig/skills/sources.toml` and `plugins/tools/awman/skills/sources.toml` to the new schema (notes fields untouched — verified byte-for-byte)
- Remove the stale `- **Version**:` bullet from both plugins' `sources.md` and add a `sources.toml` index pointer line under each H1
- Update all five `sources-*.nu` scripts for the `skills` list field and the new `none` check_method (internal/no-upstream rendering)
- Rewrite `sources-init.nu`'s emitter to the new schema with a `DRAFT`/`NEEDS-JUDGMENT` header; default mode is now dry-run (print only), `--write` required to save — no file it generates is committed in this PR
- Add `test/validate-sources.nu`: schema + plugin.json/sources.md three-way agreement validator, with a 34-case embedded self-test. Missing `sources.toml` is reported, not failed (26/28 plugins today)
- Wire `mise test:sources` into `[tasks.test]`; update `sources:init`/`sources:validate` task descriptions
- Add a `version_pin` dual-acceptance path in `test/validate-skills-quality.nu` (toml `current_version` as an alternate source of truth) — proven subset-only, no new failing key possible
- Update `skill-update/SKILL.md` (198/500 lines) for the new schema, `none` check_method, and draft-only `sources-init` behavior
- Version bumps: `claude-code` 0.2.34→0.2.35, `zig` 0.1.4→0.1.5, `awman` 0.2.4→0.2.5, `all-skills` 0.4.84→0.4.85 (verified `git diff --numstat` shows 1/1 per plugin.json, 4/4 for marketplace.json — no `json.dumps` reflow)

Adopted the adversarial-review amendments from the plan (`/Users/vinnie/.claude/plans/181-sources-foundation.md` AMENDMENTS section) verbatim:
- A-F1 (blocker): dropped the reverse-direction `b4_conditional` check and its self-test case — zig's `github_repo` under `check_method = "manual"` stays legal
- A-F2: coverage union computed per-record (`each {|s| $s.skills? | default [] }`), not via whole-table `get`, to avoid a hard error on heterogeneous entries
- A-F3: self-test case 10 states `dirs: []` explicitly
- A-F4: added guards + self-test cases for `[meta]` absent, `[[sources]]` absent, and TOML-native unquoted dates (`last_checked = 2026-06-12` parses to `datetime`; converted via `into string` before any regex)
- A-F5: dropped the unverified `validate-plugin.nu` attribution in the deviation note
- A-F6: took the smaller `sources-init.nu` branch — schema-correct emitter + DRAFT header only; URL-regex widening and full NEEDS-JUDGMENT skill-resolution deferred to PR2

**Verification (verbatim command output)**

Gate 1 — `mise run ci`: exit 0. `mise run gitleaks` (Apple Container runtime): "No secrets detected".

Waiver count: **65** (unchanged from main; histogram identical — 44 anti_fab / 8 source / 6 ref_depth / 4 duplicate_block / 2 orphans / 1 examples / 0 version_pin). No shrink occurred because zero `version_pin` keys fail today, matching the plan's prediction exactly.

`test/validate-skills-quality.nu` duplicate_block groups: still 4 (unmoved by the two `sources.md` line edits).

Negative case 1 — planted unknown key (`awman`):
```
❌ 1 finding(s):
 0  b1_entry_unknown  awman/awman: unknown key 'surprise_field' — did you mean one of: skills, name, url, ...
exit=1
```

Negative case 2 — planted scalar `skill = "x"` (`awman`):
```
❌ 3 finding(s):
 0  a2_uncovered      awman: skills with no sources entry: [awman] — add them to an existing entry's skills list, or add a check_method="none" entry
 1  b1_entry_unknown  awman/awman: unknown key 'skill' — did you mean one of: skills, name, url, ...
 2  b1_entry_missing  awman/awman: missing required key 'skills'
exit=1
```

Negative case 3 — `check_method = "none"` carrying a `url` (planted on `zig`):
```
❌ 2 finding(s):
 0  b4_none_forbidden  zig/internal-doctrine: check_method=none forbids 'url' — an entry with no upstream has no url or version
 1  c2_md_no_mention   zig: sources.md does not mention entry 'internal-doctrine' — the index must cover every sources.toml entry
exit=1
```
(second finding is correct and expected — the planted entry isn't documented in sources.md; not a defect)

Crash-vector guards (A-F4), all verified to report rather than crash: `[meta]` absent, `[[sources]]` absent, TOML-native unquoted date. All three are embedded self-test cases (34 total) and pass.

zig `notes` length re-measured after conversion: **632** characters (unchanged).

All three files were restored after each perturbation; `git status --short -- . ':!.bees'` was empty before branching and the diff after commit contains exactly the 19 files listed above.

**Found but not fixed (out of scope for this PR, noted for the lead)**

- `sources-stale.nu`'s `priority-weight` maps `critical|high|normal|low` but the schema enum is `high|medium|low` — `medium` falls to the `_ => 4` bucket and sorts below `low`. Pre-existing bug, unrelated to this bee.
- `plugins/languages/zig/skills/sources.md` still says "1 comprehensive Zig skill" while the plugin ships 7 — pre-existing false claim, out of scope for PR1 (flagged for the PR2/wave-A author per the plan).
- The plan's own §9 Case 4 expected `nu test/validate-sources.nu` to run clean after adding a valid `none` entry to zig with no `url`; it instead (correctly) reports `c2_md_no_mention` because the planted entry isn't mentioned in `sources.md`. This is the C2 invariant working as designed (§4d) — a planning-time inconsistency in the verification script, not an implementation defect. `mise run sources:check` did show the expected third row exactly as specified: `zig | language | internal-doctrine | unset | internal | no | medium`.

**Suggested tracker writes for the lead** (read-only on the tracker per instructions — not performed by this agent):
- Close `claude-skills-181` referencing this PR once merged.
- No new bees issues filed for the two "found but not fixed" items above unless the lead wants them tracked separately from the PR body.


vinnie357:

**Verdict: APPROVE.** The reviewer set out to defeat this and could not: every acceptance criterion, all six amendments (A-F1 through A-F6), and every numeric claim in the PR body reproduced independently in an isolated clone. Two minor defects confirmed; one fixed here, one filed.

### What was independently reproduced, not accepted on report

The 34-case self-test passes and the corpus run is clean (2 validated, 26 pending). All three bee-required negatives — planted unknown key, scalar `skill = "x"`, `check_method = "none"` carrying a `url` — exit 1 naming the offending field, and the reviewer's reruns matched the PR body's pasted output **verbatim**. All four A-F4 crash vectors report rather than crash.

It then went well beyond the spec: `skills = []`, a phantom directory, an entirely empty entry, `skills = [1,2]`, an empty file, non-UTF8 bytes, and broken TOML all report cleanly. No bare `main` call at file end; no `2>&1`.

`version_pin`'s subset property was proven by reading rather than accepting: the new predicate is `OLD and (v not-in toml_versions)` — a strict conjunction, so `stale_pins_new ⊆ stale_pins_old` and it can only turn failures into passes. Waivers are exactly **65** with **zero** `version_pin` entries, matching the amendment's prediction, and `duplicate_block` is still 4 groups.

zig's `notes` measured **632 characters three independent ways** and is absent from the diff entirely. `github_repo` under `check_method = "manual"` was kept per A-F1 — the amendment that would otherwise have red-lit this PR's own CI.

On the atomicity argument that justified this being one PR: `grep -rn '\.skill?' scripts/` returns nothing. Zero residual `? | default` mask sites, so the silently-empty-row failure mode is closed rather than merely believed closed.

### Fixed before merge

**`b3_date` misnamed the offending field.** A bad `meta.last_full_check` reported as `last_checked`. One word, but it fails the "messages name the offending field" bar exactly, and PR2/PR3 authors will read these messages across 26 hand-authored files. Fixed in `a520c1e` and verified by breaking it:

```
b3_date │ awman/meta: last_full_check '13-07-2026' is neither YYYY-MM-DD
```

### Filed, not fixed

**Wrong-TYPE top-level values crash instead of reporting.** `meta = "awman"` (scalar where a table belongs) aborts at `$meta | columns`; top-level `sources = ["not-a-table"]` aborts at `$s.skills?`. Both fail **red**, never green, so CI cannot silently pass — but the abort names the validator line rather than which plugin's file is bad, and the rest of the scan is skipped. That diagnosis cost lands squarely on the 26 hand-authored files in PR2/PR3, which is why it should be closed before those waves. Same `describe`/cell-path family as A-F2 and A-F4; the implementer guarded null, missing, and wrong-shape at the two table roots, but not wrong-type.

Also recorded there so it is a decision rather than an oversight: **duplicate `[[sources]]` entries covering the same skill are silently accepted.** No rule exists in the plan or the bee.

### A-F6 restraint call — judged acceptable

`sources-init.nu` landed slightly above the amendment's letter (print-by-default with `--write`, inline `# NEEDS-JUDGMENT` markers) but well below the plan body's full rewrite: no URL-regex widening, no plugin.json skill resolution, no judgment-list block, no no-URL→`none` derivation, all deferred as A-F6 required. Print-by-default was §3e mechanism 2, which A-F6 did not strike. No file generated by that script is committed.

### Gates

- Gate 1: `mise test` exit 0, `validate-sources.nu` exit 0, gitleaks clean — re-run after the fix commit.

PR #174, opened 2026-07-30
fix(test): report wrong-type sources roots and duplicate entry names

vinnie357:
claude-skills-185, filed from the Gate 3 review of #173, before the migration waves land 26 hand-authored `sources.toml` files.

**Wrong-TYPE values at the two table roots aborted the whole scan.** `meta = "awman"` (a scalar where the table belongs) died at `$meta | columns`; a top-level `sources = ["not-a-table"]` died at the `$s.skills?` cell path. Both failed red rather than green, so this was never a correctness hole — the cost was diagnosis. The abort named the validator's own line, not the offending plugin, and **stopped the scan**, so one bad file hid every other plugin's findings.

Proven against identical broken input, main versus this branch:

```
=== MAIN's validator on the broken corpus ===
Error: nu::shell::only_supports_this_input_type
   exit=1
=== THIS BRANCH on the same corpus ===
   finding rows: 2      # awman AND zig both reported
```

**The first fix was incomplete, and that is the finding worth recording.** Guarding the `[[sources]]` root left the C-group loop at line 364 still doing `$entry.name?` on a string, in a different scope — the same fix-one-site-leave-the-sibling pattern that has now bitten this repo repeatedly. Both sites are guarded, and the guards are per-entry rather than via `describe` on the collection, because a heterogeneous table reports only its column intersection.

**Duplicate entries: decided rather than left implicit.** Two entries covering the same *skill* stays legal on purpose — a skill can genuinely have several upstreams, and `claude-code` documents 35 across 12 skills. Two entries sharing a *name* is the copy-paste error, and it also makes `mise sources:check` output ambiguous. That is now `b6_duplicate_name`, with a self-test asserting the same-skill-different-name case still passes clean.

**Verified by making each fail**, then restoring and confirming the corpus is clean:

| Injected | Result |
|---|---|
| `meta = "awman"` | `b1_meta_not_table` — `awman: sources.toml 'meta' is a string, expected a [meta] table` |
| `sources = ["not-a-table"]` | `b1_entry_not_table` — `entries at index 0 are not tables` |
| duplicated `[[sources]]` block | `b6_duplicate_name` — `duplicate [[sources]] name(s): [awman] — entries may share a skill, but not a name` |
| two entries, same skill, different names | clean, exit 0 |

Self-tests 34 → 38. Two of the new cases failed on first run and both were **my** bugs, not the validator's: a fixture put `sources = [...]` after the `[meta]` header so it parsed *into* `[meta]`, and the duplicate check leaned on `group-by --to-table`, which with a closure emits a generated column name (`closure_0`) — a silent breakage waiting for a nushell bump. Replaced with an idiom that depends on no generated names.

`mise test` exit 0, waivers unchanged at 65, gitleaks clean.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.85",
+      "version": "0.4.86",
       "category": "meta",
       "keywords": [
         "skills",
@@ -119,7 +119,7 @@
       "source": "./plugins/tools/claude-code",
       "strict": true,
       "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
-      "version": "0.2.35",
+      "version": "0.2.36",
       "category": "tools",
       "keywords": [
         "claude-code",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.85",
+  "version": "0.4.86",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/claude-code/.claude-plugin/plugin.json
+++ b/plugins/tools/claude-code/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-code",
-  "version": "0.2.35",
+  "version": "0.2.36",
   "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
   "author": {
     "name": "vinnie357"
--- a/test/validate-sources.nu
+++ b/test/validate-sources.nu
@@ -66,6 +66,16 @@ def check-sources [
             severity: "fail"
             message: $"($plugin): sources.toml missing required table '[meta]'"
         })
+    } else if not (($meta | describe) | str starts-with "record") {
+        # claude-skills-185: a WRONG-TYPE root, e.g. `meta = "awman"`. Without
+        # this arm `$meta | columns` aborts the whole scan with a raw nushell
+        # error naming this file's line, not the offending plugin — so one bad
+        # file hid findings in the other 27.
+        $findings = ($findings | append {
+            rule: "b1_meta_not_table"
+            severity: "fail"
+            message: $"($plugin): sources.toml 'meta' is a ($meta | describe), expected a [meta] table"
+        })
     } else {
         let meta_cols = ($meta | columns)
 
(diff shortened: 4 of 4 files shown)


vinnie357:

**Verdict: REQUEST CHANGES → resolved.** F1 and F2 both fixed in `f6934d8`. The review was right, and the fix it forced exposed a fifth instance of the defect this repo keeps hitting.

### F1 — my guard aborted on exactly the inputs it was written to report

`$sources | any {...}` throws on non-list input, so `sources = "nope"`, a record, and the `[sources]` single-bracket typo all died **at the new guard line**. That last case is the likeliest hand-authoring mistake of the family, with 26 files still to write. Corpus-confirmed by the reviewer: one bad file killed the whole 28-plugin scan — the precise failure this PR exists to eliminate, reintroduced at a root it claimed to have guarded.

### Fixing it surfaced the fifth `describe` bug in three days

My first F1 fix used `describe | str starts-with "list"`. **My own self-tests immediately failed it**: `[[sources]]` with entries renders as `table<...>`, not `list<...>`, so the guard false-positived on every valid file. That is the same trap as PR #172's `starts-with "list"` and PR #171's dead `== "record"` branch.

The repo had already solved this — `validate-marketplace.nu` uses `describe --detailed | get type`, which returns `list` for both. **All seven `describe` sites in this file now use that form**, and a grep for the fragile spelling returns zero. This should be the default posture here, not a per-site decision.

### F2 — six field-level abort inputs, one inside the loop this PR had just patched

`name = 123` died on `str downcase` **one line below a comment lecturing about fixing one site and leaving the sibling**. Lists and records reaching `into string` then `=~` died at four more sites. All now route through a `scalar-str` helper that coerces only scalars and returns null otherwise, so wrong types report with the observed type named.

Then a **third** site of the same root cause appeared: the C-group's `if $sources != null` let a string root reach `where`. Also caught by self-test. The lesson is in the code comment — the guard belongs at every consumer, not just the first.

### F3 — duplicate-name edges, decided

Case-variant and whitespace-variant names are not treated as duplicates. Left deliberate: the rule targets copy-paste, and folding case would surprise a legitimate `GitHub` vs `github` pair. Recorded here so it is a decision rather than an omission.

### Verification, run by a haiku agent

Self-tests **34 → 43**. All eight negative cases report rather than abort — including `[sources]` → `b1_sources_not_list`, `name = 123` → no crash, and both list-valued date fields → `b3_date` naming `list<string>`. The scan-continues proof holds: breaking awman and zig simultaneously yields findings for **both** in one run.

`mise test` exit 0, waivers 65, `validate-sources.nu` clean (2 validated, 26 pending), gitleaks clean, both remote checks pass, `mergeStateStatus: CLEAN`.

### A latent bug fixed in passing

The self-test's own failure message for the `a4_review_pending` case had unescaped parens, so it crashed instead of printing — invisible until that case failed, which my change made happen. Escaped.

### Gates

- Gate 1: `mise test` exit 0, 43 self-tests, gitleaks clean — re-run by haiku after the fix commit.
- Gate 3: this record. Reviewer verification ran in isolated clones; the shared tree was never checked out.

PR #176, opened 2026-07-30
feat(sources): author sources.toml for wave A batch 2 (10 multi-skill plugins)

vinnie357:
Batch 2 of claude-skills-182, after #175 did the 8 single-skill plugins. This migrates the 10 multi-skill wave-A plugins: `core` (13 skills), `elixir` (9), `rust` (8), `ansible` (5), `dagu` (3), and `design` / `pm` / `wasm` / `graphify` / `tweag` (2 each). Corpus goes 10 → 20 sources.toml, 18 → 8 pending.

## The migration found five upstream renames

This is the point of the exercise, and it is not incidental. Prose sources go stale silently; structuring them forces a check that nobody was performing.

| plugin | sources.md says | actual (verified) |
|---|---|---|
| dagu | *no repo named at all* | `dagucloud/dagu` (id 484434975) |
| tweag | `tweag/topiary` | `topiary/topiary` (id 499408736) |
| tweag | `tweag/nickel` | `nickel-lang/nickel` (id 164738063) |
| graphify | `safishamsi/graphify` | `Graphify-Labs/graphify` (id 1200597263) |

Each confirmed by an HTTP 301 from the GitHub API plus a repo-id match, re-verified independently of the authoring agent that reported it. `beads` in #175 was the fifth.

## Version drift recorded, never silently advanced

A migration must not bump a pin — that conflates "we moved the record" with "we re-verified the content." Every drift below is recorded in the entry's `notes` with `current_version` left where the skill content was actually verified:

- **graphify** — pinned 0.8.36, live 0.9.31 (published *during* this PR), and a `v1.0.0` tag already exists. `update_priority = "high"`, `breaking_changes_likely = true`.
- **elixir** — 7 of its 8 entries are behind: elixir-lang, erlang/otp, phoenix, phoenix_live_view, tidewave, ecto, ecto_sql. The 8th (`stream_data`) is `unknown` — sources.md never pinned it, so it is unpinned rather than behind.
- **design** — the WCAG 2.1 quickref now 301s to the WCAG 2.2 quickref; 2.1 is no longer W3C's current recommendation. Not bumped: nobody verified the skill's POUR/ARIA content against 2.2's added success criteria.
- **core** — nushell 0.113.1→0.114.1, mise v2026.3.15→v2026.7.17, apple/container 1.0.0→1.2.0.
- **rust** — clap, anyhow, thiserror, tokio, serde each one patch/minor ahead of the 2026-07-17 snapshot.

## Two upstreams have no release feed at all

[...]

PR #175, opened 2026-07-30
feat(sources): author sources.toml for wave A batch 1 (8 single-skill plugins)

vinnie357:
- Author `skills/sources.toml` for the 8 single-skill plugins in wave A batch 1: `ui` (daisyui), `beads`, `linear`, `playwright`, `proxmox`, `qa`, `rig`, `sbx`
- Add the `Structured tracking: [sources.toml](sources.toml)` index pointer line to each plugin's `sources.md`
- Delete the stale `- **Version**` bullet from `sources.md` where present (`ui`, `playwright`, `qa`, `rig`) — the reviewed-at-version now lives in `sources.toml`'s `[meta].reviewed_at_plugin_version`
- Bump patch versions in `plugin.json` + `marketplace.json` for all 8 touched plugins plus `all-skills`, and sync via `mise update-all-skills`

## Per-plugin notes

- **ui**: 1 entry (`daisyui`), `check_method = "github-releases"` (`saadeghi/daisyui`, confirmed via live GitHub API check 2026-07-30, latest v5.7.8). `current_version = "unknown"` — sources.md never pinned a version the skill content was verified against.
- **beads**: 1 entry (`beads`), `check_method = "github-releases"`. Live check found the repo relocated from `steveyegge/beads` (documented in sources.md, 2026-01-24) to `gastownhall/beads` (current, latest v1.1.2) — recorded in `notes`, `github_repo` points at the new location.
- **linear**: 2 entries — `linear-api` (`check_method = "manual"`, Linear's hosted MCP/GraphQL docs, no versioned release feed) and `internal-vantageex-references` (`check_method = "none"` — first-party cross-repo references: Symphony's Linear skill, VantageEx's Linear client, three ADRs, the epic template/authoring guide, and this marketplace's CLAUDE.md template).
- **playwright**: 1 entry (`playwright-mcp`), `check_method = "manual"` — the schema has no npm check method, so `@playwright/mcp` is tracked by querying the npm registry directly. Live check found v0.0.78 (pre-1.0).
[...]

PR #177, opened 2026-07-30
feat(sources): match c2_md_no_mention against prose only, not URLs

vinnie357:
Closes claude-skills-186, filed out of the wave-A migration (#175, #176) and blocking wave B (#183).

## The rule did not mean what its name said

`c2_md_no_mention` requires every `sources.toml` entry `name` to appear in the plugin's `sources.md` — the point being that the human-readable index actually describes every tracked source. It tested containment against the **whole file**, and a `sources.md` contains URLs. So an entry named `lemonade-server` passed purely because `url = "https://lemonade-server.ai/"` sat in the file. No prose named it.

That gap is not theoretical. During wave A, **four independent authoring agents hit this rule and invented four different workarounds** — an `Entries:` sentence, an inline `_Tracked in sources.toml as_` marker, naming entries after existing prose, and naming entries after URL fragments. The last one produced `doc.rust-lang.org/stable` and `design/wit.html` as values of a field whose schema meaning is "human-readable source identifier" and which `mise sources:check` prints as an output column. A documented-but-unenforced convention is how you get four conventions.

## Measured before deciding

109 entries across the 20 migrated plugins. Exactly **7** passed only via a URL match:

| plugin | entries |
|---|---|
| elixir | `elixir-lang`, `phoenix_live_view`, `stream_data` |
| playwright | `playwright-mcp` |
| rig | `lemonade-server` |
| sbx | `docker-sandboxes` |
| tweag | `json-schema-to-nickel` |

Whole-file failures today: 0. So tightening costs 5 files, not 20 — computed, not estimated, and re-derived independently before acting on it.

## The strip is scheme-only, on purpose

`c2` now strips `https?://[^\s)>"\]]+` before matching. It deliberately does **not** strip bare domains: `m3.material.io` (design) and `cucumber.io` (pm) are legitimate entry names, and prose that mentions them without a scheme must still satisfy the rule. A naive domain-strip would silently break five plugins.

That constraint was pinned by a committed test **before** any implementation existed — `test(sources): specify c2_md_no_mention matching prose only, not URLs` (`327e85f`), written by a separate agent that was forbidden from touching `check-sources`. Five cases: URL-only must fail, prose must pass, **bare-domain-in-prose must pass**, both-present passes, and a URL embedded mid-sentence in a markdown link target must still fail. Two failed red for the right reason before the fix; all 48 self-test cases pass now.

Measured on this corpus (109 entries, 20 plugins): scheme-only leaves **0** failures; a full bare-domain strip fails **9** — the 7 `pm` entries and 2 `design` entries, every one a legitimate bare-domain name. That is the cost of "simplifying" the regex, and it is recorded at the call site so nobody does.

*(An earlier revision of this body and the code comment claimed the two regexes produce an identical failing set. That was false — and it contradicted the very next sentence. Gate 3 caught it; both are corrected. The design conclusion was never in doubt, since the bare-domain case is pinned by a committed test, but the stated justification for it was wrong.)*

## Convention, now enforced rather than described

- **First choice stays: name the source what the prose already calls it.** That is also what `sources-init.nu` generates from sources.md headings.
- **When prose doesn't name it:** append an `Entries:` sentence to the existing index-pointer line. Not a standalone line — that pointer line is byte-identical across all 20 files and all 9 existing usages put `Entries:` as a second sentence on it.
- Documented in `skill-update`'s SKILL.md (206 lines, under the 500 cap) and in the `name` field comment of `templates/sources.toml`, so a wave-B author reads it before authoring rather than discovering it at validation time.
- The failure message now tells the author what to do: *"does not mention entry 'X' in prose — URLs do not count; add it to the Entries: sentence"*.

**Drift is asymmetric and that is deliberate.** Adding or renaming an entry without updating sources.md fails `c2`, so the machine forces that direction. Removing an entry leaves a harmless stale name in prose; no reverse check was built. Recording that as an accepted residual rather than inventing a second checker for a cosmetic case.

## Grandfathered, not normalized

Only the 5 failing plugins were touched. The other 15 pass on genuine prose and were left alone — including `ansible`, whose inline `_Tracked in sources.toml as_` markers satisfy the enforced rule even though they are not the canonical shape. Normalizing 20 files to look uniform would be churn against a rule they already pass.

## Gates

`nu test/validate-sources.nu --self-test` 48 cases pass; corpus clean at 20 validated / 8 pending; **URL-only entries 7 → 0**, re-derived independently; `mise test` exit 0; `mise run test:version-bumps origin/main` exit 0 for all 6 modified plugins; gitleaks clean (exit 0, no leaks found); waivers 65 → 65.

Note for future PRs: `test:version-bumps` is a **separate CI job and is not part of `mise test`** — five missing bumps passed a green `mise test` here and were caught only by running the CI check locally before pushing.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.88",
+      "version": "0.4.89",
       "category": "meta",
       "keywords": [
         "skills",
@@ -119,7 +119,7 @@
       "source": "./plugins/tools/claude-code",
       "strict": true,
       "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
-      "version": "0.2.36",
+      "version": "0.2.37",
       "category": "tools",
       "keywords": [
         "claude-code",
@@ -197,7 +197,7 @@
       "source": "./plugins/languages/elixir",
       "strict": true,
       "description": "Elixir development skills: Phoenix, OTP, Ports, Ecto, Tidewave MCP, testing, configuration, style, and anti-patterns",
-      "version": "0.1.17",
+      "version": "0.1.18",
       "category": "language",
       "keywords": [
         "elixir",
@@ -295,7 +295,7 @@
       "source": "./plugins/tools/playwright",
       "strict": true,
       "description": "Playwright MCP browser automation for Claude Code",
-      "version": "0.1.1",
+      "version": "0.1.2",
       "category": "tools",
       "keywords": [
         "playwright",
@@ -367,7 +367,7 @@
       "source": "./plugins/tools/rig",
       "strict": true,
       "description": "Rig (rig-core): building multi-provider and hybrid LLM clients in Rust",
-      "version": "0.1.1",
+      "version": "0.1.2",
       "category": "tools",
       "keywords": [
         "rig",
@@ -416,7 +416,7 @@
       "source": "./plugins/tools/sbx",
       "strict": true,
       "description": "Docker Sandboxes (sbx CLI): run AI coding agents in isolated microVMs with hypervisor-level isolation",
-      "version": "0.1.2",
+      "version": "0.1.3",
       "category": "tools",
       "keywords": [
         "docker",
@@ -452,7 +452,7 @@
       "source": "./plugins/tools/tweag",
       "strict": true,
       "description": "Tweag tools: Topiary universal code formatter and Nickel configuration language",
-      "version": "0.1.1",
+      "version": "0.1.2",
       "category": "tools",
       "keywords": [
         "tweag",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.88",
+  "version": "0.4.89",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/languages/elixir/.claude-plugin/plugin.json
+++ b/plugins/languages/elixir/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "elixir",
-  "version": "0.1.17",
+  "version": "0.1.18",
   "description": "Elixir development skills: Phoenix, OTP, Ports, Ecto, Tidewave MCP, testing, configuration, style, and anti-patterns",
   "author": {
     "name": "vinnie357"
--- a/plugins/languages/elixir/skills/sources.md
+++ b/plugins/languages/elixir/skills/sources.md
@@ -2,7 +2,7 @@
 
 This file documents the sources used to create the elixir plugin skills.
 
-Structured tracking: [sources.toml](sources.toml) — versions, check methods, and skill coverage live there.
+Structured tracking: [sources.toml](sources.toml) — versions, check methods, and skill coverage live there. Entries: `elixir-lang`, `erlang/otp`, `phoenix`, `phoenix_live_view`, `tidewave`, `stream_data`, `ecto`, `ecto_sql`.
 
 ## Anti-Patterns Skill
 
--- a/plugins/tools/claude-code/.claude-plugin/plugin.json
+++ b/plugins/tools/claude-code/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-code",
-  "version": "0.2.36",
+  "version": "0.2.37",
   "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
   "author": {
     "name": "vinnie357"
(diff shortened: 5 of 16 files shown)


vinnie357:

Verdict: **APPROVE**. No blocking findings. Every gate claim was recomputed, not taken on faith; one body/comment claim could not be reproduced as stated (non-blocking, details below).

### Findings

**[accuracy, non-blocking] The "identical failing set" measurement is under-specified and, as most naturally read, false.**
The PR body and the code comment in `test/validate-sources.nu` both state: *"scheme-only and a more aggressive domain-stripping regex produce the identical failing set."* Reproduced against the main-tree corpus:

- Full bare-domain strip (`[a-z0-9-]+(\.[a-z0-9-]+)+(/[^\s)>"\]]*)?`): **16 failing**, not 7 — it adds 7 pm entries (`cucumber.io`, `agilealliance.org`, `spdx.org`, `owasp.org`, `getdx.com`, `productschool.com`, `addyosmani.com`) and 2 design entries (`m3.material.io`, `oritop.co`). NOT identical.
- Domain-strip restricted to path-carrying tokens (`[a-z0-9-]+(\.[a-z0-9-]+)+/[^\s)>"\]]*`): exactly the same 7, identical set.

So the claim is true for one aggressive regex and false for another, and the body never says which was measured. It also cannot be simultaneously true with the adjacent sentence "the aggressive one … costs the bare-domain case": a regex that costs the bare-domain case produces a *different* failing set by definition. The design conclusion (scheme-only) is correct and independently pinned by the committed bare-domain test, so behavior is unaffected — but the shipped code comment records a measurement that doesn't hold under its natural reading. Recommend tightening that comment's wording in a follow-up commit (state the regex that was compared, or drop the "identical set" sentence and keep the test-pinned rationale).

**[residual gap, non-blocking] Three URL forms are not stripped: schemeless `www.foo.io`, scheme-relative `//foo.io`, and non-http schemes (`ftp://foo.io`).** Reproduced: an entry whose name occurs only in such a token still passes. Grep of all 20+ sources.md files found zero occurrences of any of these forms (every `www.` sits inside an `https?://` URL), so this is theoretical for this corpus — consistent with restraint. Worth remembering if a future sources.md ever writes a schemeless URL.

**[unverified] "gitleaks clean across 487 commits."** Current refs give: origin/main 246, PR head reachable 248, `--all` 496. None is 487. Plausibly the ref state at scan time; could not reproduce, could not falsify.

**[scope note] Bee 186 AC #2 said "decide and record whether c2 should match prose only; if yes, file the follow-up."** The PR implemented the tightening directly instead of filing a follow-up bee. The substance is what the bee wants and it unblocks wave B, so this is a justified judgment call — recording it because it exceeds the AC's letter.

**[process note] The implementation regex (`https?://[^\s)>"\]]+`) differs from the planned `\S+`.** Measured on both trees: zero entries where the two disagree. The char class is the more conservative choice — it stops before `)`, `>`, `"`, `]`, keeping post-URL text as prose instead of consuming it. The substitution is sound; only the "no deviations" self-report was off.

- **Regex, adversarially probed** with 13 constructed inputs: `<https://…>` autolinks, URLs in inline backticks, URLs in fenced code blocks, uppercase `HTTPS://` (handled via the pre-existing downcase), reference-style link definitions, HTML `href="…"`, Wikipedia-style parens in the path (the name inside `Thing_(foo)` IS stripped — `(` is consumed, only `)` terminates), and prose immediately after a link's closing paren (correctly retained). No false positive found: only scheme-prefixed tokens are ever removed, and the two must-pass shapes (`m3.material.io`-style bare-domain prose, link *text* naming an entry) survive.
- **All three headline numbers recomputed** from clean `git archive` snapshots of origin/main and the PR head — no working-tree state involved: **109 entries across 20 plugins; exactly 7 URL-only on main** (sbx 1, tweag 1, playwright 1, rig 1, elixir 3 — matches the body's table plugin-by-plugin); **0 on the branch**.
- **TDD integrity**: `git diff 327e85f..8c82b08 -- test/validate-sources.nu` touches only `check-sources` (comment, `md_prose` binding, one condition, one message) — no test-case record altered. The frozen test commit run standalone fails **exactly** the 2 new-behavior cases (`c2: name appears only inside a scheme-prefixed URL` and the mid-sentence link-target case) and nothing else; the impl commit passes all 48. The five cases genuinely specify: the bare-domain case rejects any domain-stripping implementation, the mid-sentence case rejects any line-anchored or URL-field-only implementation.
- **The 5 normalized sources.md files**: each changed exactly one line (the pointer line). Every `Entries:` list was checked name-by-name against its plugin's `sources.toml` — all 5 lists carry the complete entry set (elixir 8, tweag 4, rig 2, playwright 1, sbx 1), not just the previously failing 7 names.
- **Grandfathering, entry by entry**: all 109 entries across all 20 plugins pass the NEW rule on the branch tree (0 failures) — including ansible, whose `_Tracked in sources.toml as_` markers are prose after the strip and satisfy the rule as claimed. Corpus validator on the branch: `sources.toml validation clean (20 file(s) validated, 8 pending)`.
- **Version bumps**: all 6 plugins + all-skills have plugin.json == marketplace.json (0.4.89 / 0.2.37 / 0.1.18 / 0.1.2 / 0.1.2 / 0.1.3 / 0.1.2); diffs touch version lines only; trailing bytes byte-identical to origin/main for every modified JSON (root plugin.json's missing trailing newline is pre-existing on main); no zero-byte files; quality-baseline byte-identical, `allowed_failures` 65 → 65.
- **Documentation accuracy**: the SKILL.md subsection and template comment describe exactly what the code does (scheme-prefixed strip, prose match, Entries-sentence remedy); `test:sources` is a real mise task inside `mise test`'s depends; `sources-init.nu` does derive names from sources.md headings as claimed; SKILL.md is 206 lines (< 500); no new reference files, depth unchanged.
- **Body bookkeeping numbers**: 9 pre-existing `Entries:` usages on main ✓; the pointer line is byte-identical across all 20 files (1 unique variant after stripping `Entries:` suffixes) ✓; 48 self-test cases ✓; "5 files, not 20" ✓.
- **Hygiene**: no `TODO`/`FIXME`/`XXX`/`HACK` markers in the diff; no `2>&1`, no bare `main` invocation in the changed Nushell.

PR #178, opened 2026-07-30
docs(sources): document that current_version and last_checked accept unknown

vinnie357:
Closes claude-skills-187, the last blocker on wave B (#183) after #177 cleared #186.

## The escape hatch that prevents fabricated dates was undocumented

`sources.toml` permits the literal `"unknown"` when a value cannot be established. That is the anti-fabrication mechanism: it lets an author record *"I could not determine this"* instead of guessing a plausible date or version.

The `[meta]` paragraph in `skill-update`'s SKILL.md documented it. **The per-entry field table did not** — it listed `last_checked` as `YYYY-MM-DD`, required, with no sanctioned way to say you don't know. An author reading that table sees a mandatory date field, and that is precisely the pressure that produces an invented one.

This is not a hypothetical gap. Measured on the current corpus of 109 entries across 20 plugins:

- **64 entries** use `current_version = "unknown"`
- **20 entries** use `last_checked = "unknown"`

A majority convention, relied on by every wave-A author — who only knew it was permitted because their dispatch prompts said so. The skill itself never did.

## The two fields are not enforced the same way, and the doc now says so

This is the part worth getting right, because a doc that overstates a check is the same failure class this epic keeps finding:

- **`last_checked` is enforced.** `test/validate-sources.nu` rule `b3_date` accepts `YYYY-MM-DD` or the literal `unknown` and rejects anything else.
- **`current_version` is not format-checked at all.** It appears only in the allowed-keys, network-required, and `none`-forbidden lists. `current_version = "banana"` passes validation silently.

So `unknown` is enforced in one place and pure convention in the other. The new text states that distinction explicitly rather than implying uniform enforcement. Both claims were verified against the source by two independent readings before the text was written.

## Scope

`templates/sources.toml` was checked and **left alone** — it already documents `unknown` for both fields, including the line "A guessed version is fabrication - use \"unknown\"". The template was more accurate than the SKILL.md it accompanies. Adding anything there would have been duplication.

Three files: the SKILL.md edit (206 → 208 lines, under the 500 cap) and the `claude-code` version bump in its plugin.json and marketplace entry.

## Gates

`mise test` exit 0; `mise run test:version-bumps origin/main` exit 0; `nu test/validate-sources.nu` clean at 20 validated / 8 pending; self-test 48 cases; gitleaks clean; waivers 65 → 65.

## Follow-up, not fixed here

`current_version` accepting any string is a real validator gap, not a doc gap — but a fix has to special-case `unknown` (64 entries depend on it) before validating the rest, so it is out of scope for a field-table correction. Filed separately.


vinnie357:

Reviewed as an adversary: assumed a defect and attacked every claim in the body. All verification was by execution against `origin/feature/claude-skills-187-document-unknown` blobs and the live validator — no checkout of the shared tree.

### Blocking findings

None.

### What was attacked, and the evidence

**1. The enforcement-asymmetry claims (the crux) — verified twice, by reading and by execution.**

Reading `test/validate-sources.nu`: `current_version` appears in rule logic only at lines 27 (allowed keys), 36 (`ENTRY_NETWORK_KEYS`), 38 (`NONE_FORBIDDEN`) — no format branch anywhere. `last_checked` is checked at lines 355–370 by rule `b3_date`: `unknown` or `DATE_RE = '^\d{4}-\d{2}-\d{2}$'`, anything else fails. Exactly what the new paragraph states.

Empirically, calling `check-sources` directly with literal TOML fixtures (same harness shape as the self-test):

| fixture | findings |
|---|---|
| `current_version = "banana"` | 0 |
| `current_version = "https://example.com/releases/v1.0.0"` (pasted URL) | 0 |
| `current_version = "unknown"` | 0 |
| control: `last_checked = "banana"` | 1 — `b3_date` |
| control: `last_checked = "unknown"` | 0 |

The doc's description of the validator is correct in both directions.

**2. The measured counts — recomputed independently.** Nushell sweep over `plugins/**/skills/sources.toml`: 20 files, **109 entries**, **64** with `current_version = "unknown"`, **20** with `last_checked = "unknown"`. Exact match with the PR body.

**3. The scope claim (template left alone) — verified.** `templates/sources.toml` covers BOTH fields: lines 50–51 for `current_version` (including the quoted sentence `A guessed version is fabrication - use "unknown"`) and lines 61–63 for `last_checked` (`YYYY-MM-DD or "unknown"`, example value `"unknown"`). Adding there would have been duplication; leaving it alone was correct.

**4. Does the text close the gap for an author?** Yes. `unknown` now sits in the Values column of both table rows — visible to a skimmer who never reads prose — and the when-to-use paragraph sits immediately after the table, not buried. Bonus verification of the table's Required column against the validator: `last_checked` is in `ENTRY_REQUIRED_ALWAYS` (line 34), so its unconditional "yes" is right; `current_version` "yes unless `none`" matches `ENTRY_NETWORK_KEYS` + `NONE_FORBIDDEN`.

**5. Style.** Branch SKILL.md is 208 lines (cap 500). Frontmatter is `name`/`description`/`license` only — no `allowed-tools`. The added text opens imperative ("Record `unknown` ...") and contains zero hedging verbs (no should/may/might/consider/try to). No new TODO/FIXME markers in the diff.

**6. Version bumps and file integrity.** `plugin.json` and the marketplace entry both read 0.2.38 (verified by `jq` on branch blobs). Both files are byte-identical in size to `origin/main` (14847 / 965 bytes — the version string is same-length), same trailing bytes (`]\n}\n`), valid JSON, nothing truncated, nothing else changed.

**7. Restraint / scope vs bee 187 AC.** The AC asked for the table values plus one line on when to use `unknown`. The PR adds that plus two sentences on enforcement asymmetry. Judged justified, not creep: without them the paragraph would imply uniform enforcement, which is the overstate-a-check failure class this epic exists to kill — and follow-up bee claude-skills-189's AC explicitly depends on the table documenting whatever check shape lands.

**8. Corpus consistency — is `unknown` dodging available data?** Spot-checked tweag, dagu, and ui (3 of 16 files carrying `current_version = "unknown"`): none of their sources.md files or skill bodies state a version, so no plainly-available value is being dodged. The superficially odd pattern of `last_checked = 2026-07-30` (live check today) alongside `current_version = "unknown"` is coherent given the field's definition — the version *the content was verified against*, which a freshness check does not establish.

**9. Gates cross-check.** `nu test/validate-sources.nu --self-test` → "passed (48 cases)", matching the body. Full validator run reports 20 validated / 8 pending, matching the body. Bee claude-skills-189 exists in the bees commit and matches the "filed separately" claim; its exported description is disclosure-clean.

### Non-blocking

- [nit] The when-to-use sentence lists "a live upstream check" among ways to establish the value. For `current_version` a live check establishes the *latest* version, not the version the content was verified against; strictly it only establishes `last_checked`. Reading the three sources as jointly covering both fields resolves it, so wording only.
- [nit] `test/validate-sources.nu` is a repo-internal path invisible to a marketplace install of the plugin — consistent with the rest of this skill (its mise tasks are equally repo-internal), so acceptable as-is.

### Verdict

**APPROVE.** Every falsifiable claim in the PR body was recomputed or executed and held. The two nits do not warrant a round-trip.

PR #179, opened 2026-07-30
feat(sources): author sources.toml for wave B, completing the migration

vinnie357:
Closes claude-skills-183. **This completes the migration: all 28 plugins now have a conforming `sources.toml`, 206 entries, 0 pending.**

Wave B is the 8 plugins whose sources.md is organised *by source* rather than by skill, plus the oddballs: agent-sandboxing (7 skills), claude-code (12), github (6), slidev (8), whamm (2), allium, altana, runex.

## The by-source files needed reading, not slugging

`claude-code` is the inverse of everything migrated so far — 441 lines organised by source, with multi-skill "Used In" bullets. The attribution that a heading-slugger would have lost: **`subagents-docs` covers both `claude-teams` and `claude-agents`**, a link visible nowhere in the Used In bullets. It only surfaces in the Update History section, where the same URL was re-fetched in July to drive a `claude-agents` frontmatter refresh.

`slidev` produced the mirror-image case: sources.md files "Draggable Elements" under the **Export** heading, but `v-drag` is documented only in `syntax/references/syntax.md:376-391` and nowhere in `export/`. Attributed to `syntax` — the one entry where the file's own section placement disagrees with where the content actually lives.

## runex: zero URLs, and no URL invented

`runex`'s sources.md has **no URL at all across 47 lines**, so there was nothing to convert mechanically. It resolved as neither "sources never recorded" nor a `none` entry: it is a first-party **private** repo whose skill is authored by reading the source tree directly (`mix.exs`, `lib/**`, `config/runtime.exs`). The `vinnie357/runex` identifier is transcribed from sources.md line 6 and the `https://github.com/` prefix constructed around it — a derivation from recorded data, not an invention, and worth stating precisely rather than as "transcribed". `current_version = "0.0.7"` traces to the skill's own "Current documented version" line at SKILL.md:11.

It stayed `check_method = "manual"` even though an unauthenticated `git ls-remote` unexpectedly succeeded locally — correctly reasoned as a cached credential helper rather than proof that an unauthenticated CI check would resolve. The unauthenticated API call returns 404, as expected for a private repo.

## A brief I gave was wrong, and the agent overruled it

I told the wave-B agent that "allium's upstream has zero published GitHub Releases." That is true of **`juxt/allium`** (the spec DSL — 404 confirmed) and false of **`juxt/allium-tools`** (the CLI binary the skill actually installs — `releases/latest` returns v3.5.0, with 18+ releases). The agent tested both rather than following me and split them into two honest entries.

[...]

PR #180, opened 2026-07-30
feat(sources): fail on missing sources.toml, retire the substring source check

vinnie357:
Closes claude-skills-184, the final item under the sources epic (#180). Reporting becomes enforcement now that all 28 plugins have a conforming `sources.toml`, and the check that was standing in for coverage is retired.

## A missing `sources.toml` is now a failure

Through PRs #175–#179 a plugin without a `sources.toml` was collected into a `pending` list and reported, never failed — the lever that let the migration land plugin-by-plugin instead of in one commit. All 28 are converted, so the lever is pulled: adding a plugin without one now breaks the build.

**Verified by making it fail**, because a green run on an intact corpus proves nothing about this. `plugins/core/skills/sources.toml` was moved aside:

```
❌ 1 finding(s):
│ missing_sources_toml │ core: missing skills/sources.toml — every plugin
│                      │ must have one (migration complete as of claude-skills-180)
EXIT: 1
```

Restored, the corpus returns to `28 file(s) validated, 0 pending`, exit 0.

## The substring `source` check is retired

Check 11 asked whether a skill's *name* appears anywhere in its plugin's `sources.md` — a whole-file substring test that cannot tell real coverage from an incidental name match. It is replaced by the exact set-difference coverage check in `validate-sources.nu`, which compares the union of every entry's `skills` array against the plugin's declared skill set.

Wave B produced the concrete argument for this. Deleting slidev's stale `## Plugin Information` block — which claimed version 0.2.0 against an actual 0.2.3 — removed the only two occurrences of the word "troubleshooting" from that file, so a correctly-covered skill started failing. The check was reading stale metadata as evidence of coverage.

**Its 7 waivers were retired, not merely waived.** Several belonged to skills whose coverage genuinely exists but is invisible to a substring scan, so that debt partly measured the check's blind spot rather than missing sources:

`claude-code/skill-update`, `core/code-review`, `dagu/rest-api`, `dagu/webui`, `rust/troubleshooting`, `zig/c-interop`, `zig/troubleshooting`

**Waivers 64 → 57.** The bee estimated 8 keys at filing; it is 7 because `github/community-health:source` was already shrunk in #179.

The removal and the shrink land in **one commit** by necessity: removing the check without shrinking leaves 7 stale passing keys, which the ratchet treats as a hard failure; shrinking without removing leaves the check firing.

## `version_pin` is toml-only

The check accepted a version pin if it appeared *either* as `"X (current)"` in `sources.md` **or** in the toml's `current_version` values. The prose path is dropped — only the structured value counts, which is the point of having a structured file.

That made 9 `(current)` annotations dead text, and they are deleted (3 in core, 4 in elixir, 2 in rust). **Every one was confirmed present in its plugin's toml before deletion** — core `v2026.3.15`/`0.113.1`/`1.0.0`, elixir `1.8.5`/`1.1.27`/`1.19.5`/`28.4.1`, rust `1.53.0`/`1.0.228` — so no pin lost its evidence. A comment in `core/skills/sources.toml` that quoted one of the deleted lines was updated so it does not cite text that no longer exists.

Zero `version_pin` findings survive across the corpus.

## Honest testing, and where it ran out

All three behaviors live inline in `main` rather than in callable functions, so none is reachable from the existing case-table harness without first extracting pure functions — production refactoring this PR does not do. The test author reported that rather than writing tests that appear to cover the change while exercising something else.

What was committed first, red, at `8013ee9`: two data assertions on real artifacts — that no `allowed_failures` key ends in `:source`, and that no `(current)` annotation remains. They encode the acceptance criteria directly and fail until the work is done. The behavioral half is covered by the make-it-fail evidence above and the full-corpus runs, stated as integration steps rather than dressed up as unit tests.

That gap is worth naming: extracting `stale_pins` and the missing-file guard into pure functions would make all three unit-testable. Not done here — out of scope for the flip, and worth its own bee.

## Gates

`mise test` exit 0; `mise run test:version-bumps origin/main` exit 0 for core, elixir, rust; `nu test/validate-sources.nu` 28 validated / 0 pending; sources self-test 48 cases; skills-quality self-test all green including the 2 new baseline cases; gitleaks clean.

With this, epic #180 is complete across five PRs: #175, #176 (migration), #177 (`c2` prose-only), #178 (`unknown` documented), #179 (wave B), and this flip.

## Gate 3 findings

**Blocking, fixed here:** `test/README.md` still documented the pre-flip validator in four places — a check count of 17, the retired `source` check as a live row, and `version_pin`'s dropped `"X (current)"` sources.md path. That is this epic's own failure class, in the README for the file this PR changes, and it needs no version bump. Corrected.

Note on the count: the README heading tracks `check_count`, not the table's row count. At main the table listed 18 rows under a heading of 17, because `fm_schema` can fail but is not counted. That off-by-one is preserved exactly rather than silently changed mid-flip — filed as **claude-skills-191**.

**Filed, not fixed:**
- **190** — `benchmark-skills.md` and `claude-skills-benchmark/SKILL.md` still describe the retired check. They live inside the `claude-code` plugin, so fixing them needs a version bump; split out rather than folded in.
- **191** — the `check_count` off-by-one above.
- **192** — extract pure functions so the flipped behaviours become unit-testable. Gate 3 confirmed the disclosed gap is real: an implementation that wrongly kept the sources.md acceptance path would pass every gate silently. Inert today with zero annotations remaining; it reopens the moment a plugin adds one.

**Verified beyond the implementer's own test:** Gate 3 reproduced C1 on a *different* plugin (slidev), and on a plugin with **both** files removed (ui), confirming discovery is marketplace-driven so all 28 are guarded — not just the one that was tested. It also confirmed the quoted failure message matches what the code emits character for character, recomputed 64→57 on both sides of the diff, and verified all 9 deleted annotations against the three tomls independently.


vinnie357:

One blocking finding (stale documentation of the changed validator — the failure class this epic tracks), five non-blocking. Every claim below was verified by execution in a temp clone at `be71d87`; nothing is asserted from reading alone.

### Blocking

**F1 — `test/README.md` still documents the pre-flip validator.** Four spots contradict the shipped behavior:

- Line 170: heading `### Per-skill checks (17)` — the constant is now 16.
- Line ~143: "17 static checks per skill".
- Line 185: the `source` row — "Skill is documented in the plugin's `skills/sources.md`" — documents the check this PR retires.
- Line 192: the `version_pin` row — "matches an 'X (current)' entry in `sources.md`" — documents exactly the acceptance path this PR removes, with no mention of `sources.toml`.

This is the epic's own named failure class ("a doc that describes a check that no longer exists"), it documents the very file the PR modifies, and `test/README.md` is not plugin content so no version bump is needed. Fix in this PR.

### Non-blocking

**F2 — same staleness in plugin content (follow-up bee, needs a claude-code bump):** `plugins/tools/claude-code/commands/benchmark-skills.md:45` ("Skill appears in plugin's `sources.md`") and `plugins/tools/claude-code/skills/claude-skills-benchmark/SKILL.md:23` ("sources documented in the plugin's `sources.md`") both describe the retired check.

**F3 — pre-existing check-count off-by-one, preserved not introduced.** The per-skill loop at `origin/main` can append 18 distinct failure keys with `check_count = 17`; at head it is 17 keys with `check_count = 16` (measured by enumerating `$failed | append` in the loop both sides). `fm_schema` appears to be the un-numbered extra. The PR's −1 is correct relative to main; the underlying discrepancy (score can theoretically go negative) deserves its own bee.

**F4 — the disclosed test gap is real, so file the named follow-up.** Neither the C1 missing-file guard (`missing_sources_toml` appears only in `main`, no self-test case) nor the C2 md-path removal has an automated test. A wrong implementation that deleted the 9 annotations but *kept* the `sources.md` acceptance path would pass every gate silently — inert today (no `(current)` text remains corpus-wide), but the enforcement hole would reopen for any future plugin. The PR body names pure-function extraction as the fix and is honest about the gap; the bee should actually exist before merge is forgotten. (C3's gap is covered: leaving check 11 in place while shrinking the baseline produces 7 new unbaselined `source` failures — the corpus run catches it.)

**F5 — nit:** self-test case 17 watches only core/elixir/rust `sources.md` and matches any `\(current\)`, not the ` (current)` annotation form. Inert either way since check 17 no longer reads `sources.md`.

**F6 — nit:** the hardcoded `0 pending` success line is accurate (non-empty `pending` always produces fail findings and exits 1 first), but the `pending` plumbing is now vestigial phrasing.

- **C1, empirically, beyond the implementer's own test.** Removed `slidev`'s `sources.toml` (a different plugin than the tested `core`): exit 1, `missing_sources_toml`, message names slidev. Removed **both** `sources.toml` and `sources.md` from `ui`: exit 1. Plugin discovery is marketplace-driven (all 28 local-source plugins), so no plugin escapes the guard. Intact corpus: `28 file(s) validated, 0 pending`, exit 0. The failure message emitted by the code matches the PR body quote text.
- **C3 atomicity.** `append "source"` is gone from the loop (not skipped). Baseline diff is removal-only: 0 added lines, 49 removed = exactly the 7 named entries × 7 lines; no other entry touched. Recomputed 64 → 57 from both trees. No `:source` residue anywhere outside the new self-test's own assertions.
- **C2 independently.** All 9 deleted annotation versions confirmed present as `current_version` in their plugin's toml by parsing the tomls (core v2026.3.15/0.113.1/1.0.0, elixir 1.8.5/1.1.27/1.19.5/28.4.1, rust 1.53.0/1.0.228). Full corpus run: exit 0, zero `version_pin` findings, zero `(current)` strings in any `sources.md`. No dangling `sources_content`/`sources_path` references.
- **TDD integrity.** `git diff 8013ee9..be71d87 -- test/validate-skills-quality.nu` touches only `main`'s inline logic and comments — no self-test case records. Red proof reproduced: at `8013ee9` the self-test exits 1 with both new cases failing, naming exactly the 7 `:source` keys and counting exactly 9 `(current)` lines. At head: 17 cases pass, exit 0. The data-assertion approach is an honest call given the inline-`main` structure (see F4 for its limit).
- **Numbers in the body.** 28 validated ✓, 7 keys ✓, 9 annotations (3 core + 4 elixir + 2 rust) ✓, 64 → 57 ✓, sources self-test 48 cases ✓, `check_count` 16 ✓ relative to main (F3 caveat).
- **Version bumps.** core 0.2.43, elixir 0.1.19, rust 0.1.10, all-skills 0.4.91 — each `plugin.json` equals its marketplace entry, diffs are version-line-only, trailing 16 bytes identical to `origin/main`, `check-version-bumps.nu --base origin/main` exit 0.
- **Scope vs bee 184.** Diff matches the bee's scope exactly (the bee's "3 core SKILL.md files" wording is off — they are `sources.md` files — and the PR edited the right ones; "8 keys at filing" → 7 is explained by #179, verified).

vinnie357:
## Gate 3 follow-up — APPROVE

`c5e6103` resolves F1, verified at `447f9b4` in a temp clone:

- All four stale spots in `test/README.md` fixed: "16 static checks per skill", heading `(16)`, `source` row deleted, `version_pin` row rewritten as toml-only citing claude-skills-184.
- Repo-wide re-sweep for pre-flip descriptions ("migration pending", "flips it to a failure", `X (current)`, substring-check language): zero hits beyond the two plugin-content locations already tracked in claude-skills-190. No third location found.
- No new inconsistency introduced: 17 table rows under a heading of 16 preserves the pre-existing convention (heading tracks `check_count`; `fm_schema` is the un-numbered key) — same shape as main's 18-rows/17-heading. Agreed that reconciling mid-flip would have been the wrong move; claude-skills-191 is the right home.
- Bees 190/191/192 confirmed filed in `447f9b4`; the added `issues.jsonl` lines scan clean for disclosure (no op:// paths, RFC1918 literals, or estate hostnames).
- Both validators re-run at `447f9b4`: quality self-test exit 0 (17 cases), `validate-sources.nu` exit 0 (28 validated).

PR #181, opened 2026-07-31
refactor(sources): extract missing-sources-findings and find-stale-version-pins as pure functions

vinnie357:
Closes claude-skills-192, filed out of the Gate 3 review on #180.

## The gap this closes

PR #180 flipped three checks from reporting to enforcement. None of them was reachable from the existing case-table harnesses, because all three lived inline inside `main` rather than in callable functions — `check-sources` receives an already-parsed toml and never sees the missing-file case at all.

The test author on that PR reported this instead of writing tests that appear to cover the change while exercising something else, and Gate 3 confirmed the gap was real: **an implementation that wrongly kept the `sources.md` acceptance path in `version_pin` would have passed every gate silently.** Inert today because no `(current)` annotation remains — and live again the moment a plugin adds one.

## Two extractions, nothing else

- `missing-sources-findings [pending: list<string>] -> list<record>` in `validate-sources.nu`, sorting internally so `main` calls it directly on `pending`.
- `find-stale-version-pins [content: string, toml_versions: list] -> list<string>` in `validate-skills-quality.nu`, returning exactly today's `stale_pins`.

The test author specified both names and signatures **before** any implementation existed; the implementer built to that contract. `stale_pins` feeds three downstream consumers — the `version_pin` failure append, the `detail_count` entry, and the `details` string join — and all three still receive the same value.

## 10 cases, written red first

Committed at `a9b25ba` before the functions existed:

**`missing-sources-findings`** (3) — empty list yields nothing; one missing plugin yields one finding with the exact message asserted; two unsorted names yield two separate findings in sorted order, proving sorting is the function's job and not one merged finding.

**`find-stale-version-pins`** (7) — a pin absent from `toml_versions` is stale; a pin present is not; no pin yields empty (the soft-check requirement); the `Currently at version X` form works standalone; a trailing dot is trimmed before comparison; two different versions across both regex forms resolve independently; **the same version pinned via both forms dedupes to one entry** — that last one is the real `uniq` regression catch, since without dedup the result comes back length-2 against a length-1 expectation.

## The red had to be made safe first

Nushell resolves a custom-command call at runtime, so calling a not-yet-existing function raises an unhandled error that **aborts the whole script**. In `validate-skills-quality.nu` that mattered: `main --self-test` runs 8 suites sequentially, and an unguarded abort in suite 2 would have silently skipped the 6 queued behind it with no output at all — a failure that reads as "everything else passed."

Each call is wrapped in `try`/`catch` returning a type-checked sentinel, so every case reports its own red line and the downstream suites still run. Verified: 7 suites green alongside 7 red cases while the functions were missing. The scaffolding needs no removal — once the real function exists, the `try` block simply returns its value.

## Behaviour-preserving, and proved that way

A refactor's evidence is that nothing moved:

- corpus **28 validated / 0 pending**
- waivers **57** — unchanged, not 56 or 58
- `mise test` exit 0
- the `missing_sources_toml` message verified **byte-identical** by making it fail again (moving whamm's toml aside) and comparing the rendered text against the frozen commit — it travels through nushell interpolation with escaped parentheses, exactly the kind of thing that shifts silently during a move

## One correction

My instruction to the implementer predicted the sources self-test would report 51 cases. It reports 48, and that is correct: the printed count covers the `$cases` table only, while the 3 new cases run in their own loop with inline reporting — the same pattern the pre-existing `a4_review_pending` severity assertion has always used. The count has never been the total number of assertions in that file. Flagged rather than "fixed", since changing it would mean editing frozen tests to match a wrong expectation.

That said, a printed count that excludes real assertions is the same species as the `check_count` off-by-one already filed as **claude-skills-191**, and belongs in that bee's scope rather than here.

## Gate 3

**APPROVE, nothing blocking.** The review did the two things that make a refactor claim falsifiable:

**A differential run.** Both validators executed at `origin/main` and at HEAD over the real corpus — byte-identical output, equal exit codes. Since a clean corpus exercises neither changed path, it forced both: removing `github` and `whamm`'s `sources.toml` at each revision (identical failure text and sorted ordering, exit 1 both), and injecting `Current stable: 99.0.0. Currently at version 99.0.0. Current stable: 88.8.8` to exercise dedup, the trailing-dot trim, a multi-pin line, and all three `stale_pins` consumers at once — including the `detail_count` classification shifting 16/16 → 15/16 identically.

**Five deliberate mutations against the sentinel scaffolding**, the highest-risk part of this change since it installs `try`/`catch` permanently in a test harness. A thrown error, a dropped `uniq`, a dropped trim, an em-dash typo in the message, and a wrong-typed return were **all caught**, exit 1 each. The throw stayed isolated — downstream suites still ran — so the guard contains failures without absorbing them.

**Fixed from the review:** the catch label printed `not yet implemented (<msg>)` for any runtime error, permanently — accurate while the functions were missing, actively misleading afterwards (it rendered "not yet implemented (deliberate boom)" during the mutation test). Now reads `call raised (<msg>)`.

**Not taken:** the suggestion to `| sort` the got value in the pin cases so the frozen tests stop pinning output order. Order is not incidental here — it feeds `main`'s `details` string join, which the differential showed renders in insertion order. Pinning it is the stronger test, so the cases stay order-sensitive deliberately.


vinnie357:

Reviewed as an adversarial pass with the goal of defeating the change. Every claim below is from tool execution in a scratch clone at `e2c8404` (never the shared tree); nothing is asserted from reading alone.

### Verdict

**APPROVE.** No blocking findings. The refactor is behaviour-preserving by differential execution, not just by eye, and the frozen tests catch every mutation I threw at the extracted functions.

### What I attacked and what happened

**1. Behaviour equivalence (old inline vs new functions), by execution**

- Full-corpus differential: `validate-sources.nu` and `validate-skills-quality.nu` run at `origin/main` (058141f) and at HEAD — stdout+stderr **byte-identical** both scripts, exit 0 both.
- The clean corpus never fires either changed path, so I forced both:
  - Moved `github` and `whamm` `sources.toml` aside at both revisions: failure output **byte-identical** (exact `missing_sources_toml` message text incl. the escaped-paren interpolation, sorted two-plugin ordering, exit 1 both).
  - Injected `Current stable: 99.0.0. Currently at version 99.0.0. Current stable: 88.8.8` into a SKILL.md at both revisions: output **byte-identical**, exit 1 both. This exercises all three `stale_pins` consumers at once — the `version_pin` failure key (`github/act:version_pin`), the ratchet classification (act drops 16/16 → 15/16), and the details join (`99.0.0 88.8.8`) — and confirms dedup (99.0.0 once despite both forms), trailing-dot trim, and multi-pin-per-line capture are unchanged.
- Direct edge probes via `nu -c 'source ...'`: duplicate names in `pending` produce duplicate findings (1:1 with the old loop's semantics); names containing parens and non-ASCII interpolate cleanly; a bare version with no phrase prefix is not captured; empty `toml_versions` handled (also pinned by the dedup unit case).
- The `find-stale-version-pins` body is textually identical to the removed inline block; `missing-sources-findings` replaces `for`+`append` with `sort | each` and the call site appends the returned list — nushell `append` flattens, verified by the injected-failure differential above. The call site does **not** double-sort (plain `append`, no `sort`).

**2. TDD integrity**

- `git diff a9b25ba HEAD -- test/`: the implementation commit adds exactly two `def` blocks and swaps the two inline call sites. The frozen case tables and loops are untouched.
- At `a9b25ba` (functions absent): sources self-test exits 1 with 3 red cases; quality self-test exits 1 with 7 red pin cases **while the other 7 suites still run and pass** — matching the PR body's "7 suites green alongside 7 red cases" claim exactly.

**3. Error-swallowing attack on the try/catch scaffolding (highest risk)**

Five mutations at HEAD, each restored after:

| Mutation | Result |
|---|---|
| `find-stale-version-pins` throws (`error make`) | exit 1, 7 red lines, 7 downstream suites still ran — isolated, not swallowed |
| drop `uniq` | exit 1 — caught by the dedup case (`want [1.0.0], got [1.0.0, 1.0.0]`) |
| drop trailing-dot trim | exit 1 — 3 cases red |
| em-dash → hyphen in the finding message | exit 1 — both non-empty message cases red |
| return `"oops"` (wrong type) | exit 1 — `nu::parser::output_type_mismatch`, hard loud abort |

The sentinel path sets `$failed = true` on any non-list result, so the scaffolding converts errors into counted reds rather than hiding them.

**4. The numbers (all re-measured)**

- Waivers: `jq` on `test/quality-baseline.json` → `allowed_failures` length **57**; baseline file has a **zero-line diff** vs main.
- Corpus: **28 validated / 0 pending**, exit 0.
- Sources self-test: **48 cases** at HEAD — and **48 at main too**: the counted `$cases` table is unchanged; the 3 new assertions run in their own loop outside it, the same shape as the pre-existing `a4_review_pending` inline assertion (main `test/validate-sources.nu:1653-1674`). The PR body's self-correction is accurate, and deferring the count question to claude-skills-191 is the right call.
- Baseline suite: **18 cases**, all 8 quality suites green.
- Remote CI: both checks pass.

**5. Scope / restraint**

Three files total: the two validators plus a `.bees` label dedup in its own `chore` commit. `find-stale-version-pins` sits after `find-orphan-files` as stated. Nothing else moved, renamed, or "tidied". No new `TODO`/`FIXME` markers, no `2>&1`, no bare `main` call.

### Non-blocking observations

1. **[nit]** The catch label prints `not yet implemented (<msg>)` for *any* runtime error, permanently — a real post-implementation throw reads as "not yet implemented (deliberate boom)". The genuine error message is included and the case still goes red, so this is cosmetic; worth relabeling to something neutral ("call failed") next time the file is touched.
2. **[suggestion]** The pin unit cases apply `| sort` to `got` before comparing, so the frozen tests do not pin the function's output *order* — which feeds `main`'s `details` join. The differential above proves order is preserved by this implementation (details rendered `99.0.0 88.8.8`, insertion order), so no behaviour change ships here; it is a latent test-strength gap if someone later rewrites the function body.

Attack surface covered: full-corpus differential, two fault-injection differentials, direct edge-input probes, frozen-test diff audit, red-first verification, five mutations, all four headline numbers, CI. Nothing blocking found.

PR #183, opened 2026-07-31
docs(tweag): make three reference files reachable from their SKILL.md

vinnie357:
Closes claude-skills-140. First of the two burn-down items the sources epic was clearing the way for.

## Three reference files nobody could reach

`tweag/nickel` and `tweag/topiary` carried three reference files that exist on disk but are mentioned nowhere in their SKILL.md — so under progressive disclosure an agent has no path to them. Neither SKILL.md contained the token `references/` at all.

They are not dead content: **383, 420, and 373 lines** respectively. So the fix is navigation, not deletion.

- `nickel` gains a `## References` section pointing at `references/contract-patterns.md` (contract syntax and schema-validation patterns) and `references/import-export.md` (format conversion and validating imported data).
- `topiary` gains one pointing at `references/query-syntax.md` (the S-expression query syntax for `.scm` files).

Each line says what the file contains **and when to read it**, so the pointer earns its place in a body that loads on every activation rather than just satisfying the `orphans` check.

## Waivers 57 → 55

Both `tweag/nickel:orphans` and `tweag/topiary:orphans` now pass, and the shrink-only baseline requires removing a fixed entry to lock the fix in. The diff is **removal-only: 0 insertions**.

These two entries were also load-bearing for **claude-skills-132**, which requires every waiver to carry a linked issue — they pointed here.

## One adjacent fix, disclosed

`nickel/SKILL.md:412` cited `https://github.com/tweag/nickel`. That repo **301-redirects to `nickel-lang/nickel`** — a rename this migration caught and verified (repo id 164738063), and one the plugin's own `sources.toml` already records. A skill body disagreeing with its plugin's structured source-of-truth is the exact drift this epic exists to remove, and it sat one line from an edit I was already making, so it is corrected here.

**Deliberately not touched:** `tweag/skills/sources.md` still shows `tweag/topiary` and `tweag/nickel` as the URLs accessed at the time. That is the intended division — `sources.md` records what was read and when, `sources.toml` records how to check the source now, carrying the current `github_repo` and the rename in its `notes`. Rewriting the historical record to match today's paths would erase the evidence that the rename happened.

## Gates

`mise test` exit 0; `mise run test:version-bumps origin/main` exit 0 with `tweag` bumped 0.1.2 → 0.1.3; corpus 28 validated / 0 pending; all 8 skills-quality self-test suites green; waivers **57 → 55**, baseline diff removal-only; gitleaks clean.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -452,7 +452,7 @@
       "source": "./plugins/tools/tweag",
       "strict": true,
       "description": "Tweag tools: Topiary universal code formatter and Nickel configuration language",
-      "version": "0.1.2",
+      "version": "0.1.3",
       "category": "tools",
       "keywords": [
         "tweag",
--- a/plugins/tools/tweag/.claude-plugin/plugin.json
+++ b/plugins/tools/tweag/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "tweag",
-  "version": "0.1.2",
+  "version": "0.1.3",
   "description": "Tweag tools: Topiary universal code formatter and Nickel configuration language",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/tweag/skills/nickel/SKILL.md
+++ b/plugins/tools/tweag/skills/nickel/SKILL.md
@@ -399,9 +399,14 @@ nickel export config.ncl --format toml --output config.toml
 """
 ```
 
+## References
+
+- `references/contract-patterns.md` — contract annotation syntax and the validation patterns used for configuration schemas. Read when writing or debugging a contract.
+- `references/import-export.md` — converting between config formats and validating imported data. Read when moving data in or out of Nickel.
+
 ## Additional Resources
 
 - [Nickel Official Documentation](https://nickel-lang.org)
 - [Contracts Guide](https://nickel-lang.org/user-manual/contracts/)
 - [Typing Guide](https://nickel-lang.org/user-manual/typing/)
-- [GitHub Repository](https://github.com/tweag/nickel)
+- [GitHub Repository](https://github.com/nickel-lang/nickel)
--- a/plugins/tools/tweag/skills/topiary/SKILL.md
+++ b/plugins/tools/tweag/skills/topiary/SKILL.md
@@ -366,6 +366,10 @@ mise run fmt:check
 mise run fmt:debug -- script.ncl
 ```
 
+## References
+
+- `references/query-syntax.md` — the S-expression query syntax Topiary uses to match code patterns and apply formatting rules. Read when writing or debugging a `.scm` query file.
+
 ## Anti-Fabrication Note
 
 Before claiming that Topiary formats a language or query correctly, verify behavior by running `topiary fmt` on actual code files and inspecting the output. Do not assume query behavior without testing; use `topiary visualise` to inspect the parse tree and confirm node names and structure match the query. Test idempotency by running the formatter twice and confirming output is identical.
--- a/test/quality-baseline.json
+++ b/test/quality-baseline.json
@@ -317,20 +317,6 @@
       "detail_count": null
     },
     {
-      "key": "tweag/nickel:orphans",
-      "class": "DEBT",
-      "issue": "claude-skills-140",
-      "first_seen": "migrated",
-      "detail_count": 2
-    },
-    {
-      "key": "tweag/topiary:orphans",
-      "class": "DEBT",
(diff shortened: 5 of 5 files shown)


vinnie357:

Reviewed as adversarial Gate 3: assumed a defect and attacked every claim in the body. Verified by execution against branch commit `179f26a8` in a scratch clone (no shared-tree checkout). No blocking findings; one non-blocking follow-up.

### Verified by execution

**Validator, for the right reasons.** Full `mise test` on a clean clone at `179f26a8`: exit 0. Both `tweag/nickel` and `tweag/topiary` score **17/17** with an empty failed column — `orphans`, `links`, and `ref_depth` all genuinely pass, not suppressed. All 8 skills-quality self-test suites green (frontmatter 5, ratchet 18, check-fix 19, dupes 15, vocab 10, links 8, orphans 6, schema 12); corpus 28 validated / 0 pending; `mise run test:version-bumps origin/main` exit 0 with `tweag: 0.1.2 → 0.1.3`.

**Latent ref_depth exposure — attacked, held.** The three reference files were unreachable before this PR, so a sibling-link violation could have been hiding. Grepped all three (branch tip) for `references/`, `.md`, and `SKILL`: **0 hits each**. No reference→reference links exist; making them reachable exposes nothing.

**Pointer descriptions vs actual content.** Read all three files' full heading trees:
- `contract-patterns.md` (383 lines): contract annotations, built-in/custom/combined contracts, validation patterns, blame handling, schema examples — matches "contract annotation syntax and the validation patterns used for configuration schemas."
- `import-export.md` (420 lines): YAML/JSON/TOML import, validation during import, export, merge/transform workflows, bulk conversion — matches "converting between config formats and validating imported data."
- `query-syntax.md` (373 lines): S-expression basics, capture names/formatting effects, predicates, formatting patterns, query debugging — matches "the S-expression query syntax … `.scm` query file."

No pointer misdescribes its target.

**Baseline shrink.** `allowed_failures`: 57 on `origin/main` → 55 on branch. `git diff --numstat`: **0 insertions, 14 deletions** — exactly the `tweag/nickel:orphans` and `tweag/topiary:orphans` entries (both `issue: claude-skills-140`), no other entry touched, zero tweag waivers remain, no new key added.

**URL rename, independently.** `https://github.com/tweag/nickel` → **301** → `https://github.com/nickel-lang/nickel`; GitHub API returns `full_name: nickel-lang/nickel`, `id: 164738063` from **both** paths — matches the body's cited id. Scope check on the sibling: `tweag/topiary` also renamed (per `sources.toml`, → `topiary/topiary`), but topiary's SKILL.md contains **no** repo URL at all (only placeholder example URLs at lines 169/205), so nickel's was the only stale repo link and the one-line fix covered the whole class. Not scope creep — it corrected drift against the plugin's own `sources.toml` in a file already being edited. The `sources.md` historical/`sources.toml` current division the body claims is real: `sources.toml` records both renames with dated API-verified notes; `sources.md` keeps the as-accessed URLs.

**Body numbers.** 383 / 420 / 373 line counts: exact (`wc -l` on branch). SKILL.md sizes 412 and 375 — both under the 500-line cap. Frontmatter byte-identical to main (diffed first 10 lines of each). "Neither SKILL.md contained the token `references/` at all" on main: confirmed, 0 hits in each.

**JSON integrity.** Only the two `version` lines changed across `marketplace.json` and `plugin.json`; versions match each other (0.1.3); trailing bytes identical to main (`0a 7d 0a`); both parse as valid JSON.

**Deletion vs. link.** Attacked the premise that linking was right: the duplicate-block detector reports no tweag groups, and the files carry substantive non-duplicated syntax detail — deletion would have discarded real content. Three pointer lines (~3 body lines per activation) each state content + when-to-read; none is padding.

**Other.** No TODO/FIXME/HACK markers in the diff. Reference files byte-unchanged vs main. bees `claude-skills-140` matches the PR's scope and its claude-skills-132 linkage claim.

### Non-blocking

1. **[non-blocking] Stale anchor inside newly-reachable `query-syntax.md:372`.** `https://tree-sitter.github.io/tree-sitter/using-parsers#query-syntax` 301s to `using-parsers/` — a landing page where the `#query-syntax` fragment no longer resolves; the query docs moved to `using-parsers/queries/` (probe of `using-parsers/queries/index.html` returns 200). Pre-existing content, untouched by this diff, and only surfaced because the file is now reachable. Suggest a follow-up bee rather than expanding this PR. The other two URLs in the file return 200 directly.

### Verdict

**APPROVE.** Every claim in the PR body replicated under independent verification; the diff is minimal, removal-only where it must be, and the one adjacent fix is disclosed and correct.

PR #186, opened 2026-07-31
docs(skills-quality): document the ref_depth order requirement and surface nested filenames

vinnie357:
Closes claude-skills-195, filed during claude-skills-145 when a brief I wrote about this check turned out to be wrong.

## The residual

`ref_depth`'s cross-skill exemption is evaluated against the text **before** the `references/` token only, so it fires only when the qualifier precedes the path:

- ``Per `/core:tdd`'s `references/ci-discipline.md`:`` → exempt
- ``Per `references/ci-discipline.md` in `/core:tdd`:`` → **flagged**

Both name the same real file. This is a **false positive** — the check silently fails legitimate content — which differs in kind from the two residuals already documented on that function, since those silently *pass* things that should fail.

## The bee offered widening the check. Measurement says don't.

The plan measured four things before recommending, and the headline finding is that whole-line matching is **not a widening at all**:

`cross-skill-qualified`'s second qualifier form is `parse --regex 'skills/(?P<skill>[a-z][a-z0-9-]*)/$'` — **anchored to end-of-string**. It only ever matches because it is handed a prefix ending exactly at the token. Feed it the whole line and that anchor can never match, so the repo-path qualifier form stops working entirely. One corpus file (`pm/spec-harvest/references/user-story-format.md`) flips from exempt to **failing**.

Beyond that, whole-line matching converts a false positive into a false **negative**, in two corpus-plausible shapes:

- **Unknown namespace mentioned after the path.** `lookup-skill-dir` returns empty for a name outside the local marketplace and the function then returns `true` unconditionally — so any external plugin mention later on the line exempts a genuine same-skill link.
- **A sibling skill owning the same basename.** This corpus has **19** reference basenames owned by more than one skill — `examples.md` by three, `installation.md` by four, `commands.md` by three.

And the false positive it would fix bites **zero** corpus lines today; the 145 cleanup left no residue. In a lint, trading a false positive for a false negative is strictly worse: the false positive asks an author to reword, the false negative lets the violation land silently.

The helper is shared by four call sites (`ref_depth`, check 14 `links`, and two Pass-2 paths), so the blast radius was real — measurement found 2 further false-negative diffs in check 14 and 1 in Pass-2.

## What landed instead

- **A third accepted-residual bullet** on `has-unqualified-references-token`, giving both example forms, stating that the fix is to reorder the sentence, that this residual differs in kind from the other two, and why it is not widened — with the measured figures attached.
- **One sentence** on `preceding-line` noting the same truncation makes the exemption order-sensitive for checks 14 and Pass-2. Deliberately one sentence, not a copied block: the corpus-wide `duplicate_block` scan uses an 8-line window and the shrink-only baseline cannot accept a new key at all. Verified — still 4 groups, unchanged.
- **Four regression tests** — three committed first at `b4a5369`, a fourth added after review (see the Gate 3 section below): qualifier-before (exempt), qualifier-after (flagged), a same-skill link whose trailing sibling mention names a skill owning the same basename (flagged), and one naming an unknown external namespace (flagged). They pass on arrival — **they are regression guards, not red-first TDD**, and their purpose is to make a future silent widening impossible. The fixture is a real `skill_dir_map` built via `mktemp`, not the existing cases' `[] []`, which would have made every qualifier resolve as "unknown → exempt" and the cases vacuous.
- **The offending filename now reaches the author.** `ref_depth` previously reported only its own name; the `details` column now carries `nested:<filename>`, matching the existing `frontmatter:<key>` convention.

## One item cut from the plan

The plan also proposed printing an advisory line whenever a `ref_depth` failure occurs. **Dropped.** No skill currently fails that check, so it is an unexercisable print path, and the test author confirmed it cannot be unit-tested without extracting a new function from `main` or hand-breaking a skill. The ordering rule belongs in the documentation a maintainer reaches from the check's own source.

`details` survived the same cut because it earns its place differently: it fires for **every** `ref_depth` failure, not only the ordering case.

Since nothing in a clean corpus exercises it, it was proved by hand — a sibling link was temporarily added to `core/tdd/references/ci-discipline.md`, the run showed `ref_depth │ nested:ci-discipline.md` in the details column with the expected unbaselined failure, and reverting returned the corpus to 49 waivers with no FAIL line.

## Gates

`mise test` exit 0; all 8 skills-quality self-test suites green with check-fix at 23 cases; waivers **49 — unchanged**; duplicate blocks **4 groups — unchanged**; corpus 28 validated / 0 pending; `test/quality-baseline.json` diff empty; `test/` is not a plugin so no version bump; gitleaks clean.

## Gate 3 caught a decorative guard — the PR's own deliverable

The three regression cases pass on arrival by design, which makes them impossible to distinguish from decorative ones by looking. So the reviewer was asked to apply the widening they exist to prevent and see which actually go red. **Only one did.**

Case 3 claimed whole-line matching would wrongly exempt a same-skill path via a trailing sibling mention. It did not reproduce: under widening `lookup-skill-dir` resolved the sibling, but `path exists` then returned false because the fixture only ever created `foo.md`. The line stayed flagged **for a reason unrelated to the residual being pinned**. A true assertion, and not the guard its comment described.

The fixture was one line short of real. Adding `bar.md` to the sibling skill — the exact 19-colliding-basenames shape the header comment cites — makes it bite. A fourth case was added for the second documented false-negative shape (unknown namespace after the path), which had no coverage at all and which the reviewer confirmed is load-bearing. A leaked `mktemp -d` was also cleaned up (temp-dir count now holds across runs).

Re-verified independently by applying the widening to a scratch copy and running the suite:

```
                                     shipped        widened
case 1  qualifier-before             not flagged    not flagged   control, must survive either way
case 2  qualifier-after              flagged        NOT flagged   catches it
case 3  same-basename sibling        flagged        NOT flagged   catches it (was decorative)
case 4  unknown namespace            flagged        NOT flagged   catches it (was absent)
```

Three guards fire where one did before. A test that cannot fail under its own target change is worse than no test — it manufactures confidence that the behaviour is pinned.

Check-fix suite 19 → **23 cases**. Waivers **49**, dupe groups **4**, corpus 28/0 — all unchanged.

### DIFF
--- a/test/validate-skills-quality.nu
+++ b/test/validate-skills-quality.nu
@@ -133,6 +133,8 @@ def cross-skill-qualified [prefix_line: string, dir_name: string, path: string, 
 
 # Text preceding the first occurrence of `path` in `content`, truncated to
 # just its own line (so cross-skill-qualified only sees same-line context).
+# The same truncation makes the exemption order-sensitive for checks 14 and
+# Pass-2 too — see has-unqualified-references-token's header for the rationale.
 def preceding-line [content: string, path: string]: nothing -> string {
     let idx = ($content | str index-of $path)
     if $idx < 0 {
@@ -261,6 +263,25 @@ def pass2-dir-in-scope [top: string, own_dir: string, plugin_dir: string]: nothi
 #   concrete sibling file. Accepted residual: a sibling pointer written
 #   without a file extension escapes (check 14's link regex shares this
 #   extension requirement).
+# - The cross-skill exemption is evaluated against the text BEFORE the
+#   `references/` token only (the prefix/rest split just below), so it
+#   fires only when the `/ns:skill` (or `plugins/<...>/skills/<other>/`)
+#   qualifier PRECEDES the path on the same line. This residual differs in
+#   kind from the two above: those silently PASS things that should fail;
+#   this one silently FAILS something legitimate. Flagged:
+#   ``Per `references/x.md` in `/core:tdd`: ...``. Exempt:
+#   ``Per `/core:tdd`'s `references/x.md`: ...``. Both name the same real
+#   file — fix by reordering the sentence. Not widened to whole-line
+#   matching: measured against this corpus (claude-skills-195 plan), that
+#   breaks the `plugins/<...>/skills/<other>/` qualifier form (its regex is
+#   `$`-anchored, so it can only match a line-ending prefix), costing 1
+#   corpus exemption, while a later mention of an unknown namespace or of a
+#   sibling skill sharing the same reference basename (19 such basenames in
+#   this corpus) would silently exempt a genuine same-skill link — against
+#   0 current false positives from the order requirement. A gated design
+#   (prefix-first, whole-line fallback only when the citing skill does not
+#   own the cited path) was evaluated and deferred until a real instance
+#   needs it.
 def has-unqualified-references-token [content: string, dir_name: string, skill_dir_map: list, known_plugins: list]: nothing -> bool {
     ($content | lines | any {|line|
         if not ($line | str contains "references/") {
@@ -2314,7 +2335,7 @@ def main [--update-baseline, --self-test] {
                 score: $"($score)/($check_count)"
                 failed: ($failed | str join " ")
                 new: ($new_failures | each {|k| $k | split row ":" | last} | str join " ")
-                details: ($broken_links | append $bad_invocations | append $stale_pins | append ($orphans | each {|f| $f | path basename}) | append ($fm_unknown | each {|k| $"frontmatter:($k)"}) | str join " ")
+                details: ($broken_links | append $bad_invocations | append $stale_pins | append ($orphans | each {|f| $f | path basename}) | append ($fm_unknown | each {|k| $"frontmatter:($k)"}) | append ($nested | each {|r| $"nested:($r.name)"}) | str join " ")
             })
 
             if ($failed | is-empty) {
--- a/test/validate-skills-quality.nu
+++ b/test/validate-skills-quality.nu
@@ -1033,6 +1033,50 @@ def run-check-fixes-self-test [] {
         $failed = true
     }
 
+    # --- ref_depth cross-skill exemption is word-order sensitive (claude-skills-195) ---
+    # These three cases pin the documented residual on
+    # has-unqualified-references-token's header comment (accepted-residual
+    # bullet three): the exemption fires only when the /ns:skill qualifier
+    # PRECEDES the references/ path on the same line. Unlike the cases
+    # above, they build a REAL skill_dir_map fixture — the existing ref_depth
+    # cases all pass `[] []`, which makes every qualifier resolve as
+    # "unknown -> exempt" and would make an order check vacuous. This
+    # fixture (a synthetic mktemp tree, never a real corpus skill, so a
+    # future reference-file rename in this repo can't break it) makes the
+    # `path exists` branch of cross-skill-qualified genuinely fire.
+    let rd_root = (mktemp -d)
+    mkdir ($rd_root | path join "otherns" "skills" "otherskill" "references")
+    "content" | save ($rd_root | path join "otherns" "skills" "otherskill" "references" "foo.md")
+    let rd_map = [{skill: "otherskill", dir: ($rd_root | path join "otherns" "skills" "otherskill"), plugin: "otherns"}]
+    let rd_plugins = ["otherns"]
+
+    # Case: qualifier BEFORE the path — a real cross-skill pointer whose
+    # target file exists in the other skill's own tree. NOT flagged.
+    if (has-unqualified-references-token "Per `/otherns:otherskill`'s `references/foo.md`: see there for detail." "myskill" $rd_map $rd_plugins) {
+        print $"(ansi red_bold)❌ check-fix self-test: qualifier-before cross-skill pointer wrongly flagged(ansi reset)"
+        $failed = true
+    }
+
+    # Case: the SAME pointer, reordered so the qualifier comes AFTER the
+    # path. This is the documented residual, not a bug — the test's purpose
+    # is to make a silent future widening (to whole-line matching)
+    # impossible without a failing test alerting the author.
+    if not (has-unqualified-references-token "Per `references/foo.md` in `/otherns:otherskill`: see there for detail." "myskill" $rd_map $rd_plugins) {
+        print $"(ansi red_bold)❌ check-fix self-test: qualifier-after cross-skill pointer no longer flagged \(documented residual regressed\)(ansi reset)"
+        $failed = true
+    }
+
+    # Case: a genuine SAME-skill link (no cross-skill qualifier applies to
+    # it at all) with an unrelated but real, resolvable plugin:skill mention
+    # AFTER the path on the line. Regression guard against widening to
+    # whole-line matching: the plan's adversarial measurement shows
+    # whole-line matching would wrongly treat the trailing mention as an
+    # exemption qualifier for the earlier, unrelated path. Must stay flagged.
+    if not (has-unqualified-references-token "See references/bar.md for detail, mirroring the layout in /otherns:otherskill." "myskill" $rd_map $rd_plugins) {
+        print $"(ansi red_bold)❌ check-fix self-test: same-skill link with trailing unrelated plugin mention wrongly exempted(ansi reset)"
+        $failed = true
+    }
+
     # --- invocations (enumerated upstream commands, no namespace exemption) ---
     let allium_cmds = ($UPSTREAM_COMMANDS | where ns == "allium" | first | get commands)
     let reg = [
@@ -1057,7 +1101,7 @@ def run-check-fixes-self-test [] {
     }
 
     if not $failed {
-        print $"(ansi green_bold)✅ Check-fix self-test passed \(19 cases\)(ansi reset)"
+        print $"(ansi green_bold)✅ Check-fix self-test passed \(22 cases\)(ansi reset)"
     }
     $failed
 }
--- a/test/validate-skills-quality.nu
+++ b/test/validate-skills-quality.nu
@@ -1055,7 +1055,7 @@ def run-check-fixes-self-test [] {
     }
 
     # --- ref_depth cross-skill exemption is word-order sensitive (claude-skills-195) ---
-    # These three cases pin the documented residual on
+    # These four cases pin the documented residual on
     # has-unqualified-references-token's header comment (accepted-residual
     # bullet three): the exemption fires only when the /ns:skill qualifier
     # PRECEDES the references/ path on the same line. Unlike the cases
@@ -1064,10 +1064,13 @@ def run-check-fixes-self-test [] {
     # "unknown -> exempt" and would make an order check vacuous. This
     # fixture (a synthetic mktemp tree, never a real corpus skill, so a
     # future reference-file rename in this repo can't break it) makes the
-    # `path exists` branch of cross-skill-qualified genuinely fire.
+    # `path exists` branch of cross-skill-qualified genuinely fire. otherskill
+    # owns BOTH foo.md and bar.md — the bar.md duplicate basename is what
+    # makes case 3 below load-bearing (the 19-colliding-basenames shape).
     let rd_root = (mktemp -d)
     mkdir ($rd_root | path join "otherns" "skills" "otherskill" "references")
     "content" | save ($rd_root | path join "otherns" "skills" "otherskill" "references" "foo.md")
+    "content" | save ($rd_root | path join "otherns" "skills" "otherskill" "references" "bar.md")
     let rd_map = [{skill: "otherskill", dir: ($rd_root | path join "otherns" "skills" "otherskill"), plugin: "otherns"}]
     let rd_plugins = ["otherns"]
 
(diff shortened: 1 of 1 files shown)


vinnie357:

**Verdict: REQUEST CHANGES** — one blocking finding, two non-blocking. Everything else I attacked held up, including the central measured claim, which reproduces exactly.

All work done in a throwaway clone (`git clone` + `git checkout -B rev origin/<branch>`); the shared working tree was never checked out, staged, or committed.

---

### [blocking] Regression case 3 does not fail under the widening its own comment names

The three new cases exist to make a silent widening to whole-line matching impossible. I applied that widening — the one-character-scope change at line 293:

```diff
-            let prefix = ($line | str substring 0..<$idx)
+            let prefix = $line
```

and ran the self-test. **Exactly one of the three cases went red:**

```
❌ check-fix self-test: qualifier-after cross-skill pointer no longer flagged (documented residual regressed)
```

Isolating all three under both variants:

| case | prefix (shipped) | whole-line (widened) | catches widening? |
|---|---|---|---|
| 1 — qualifier before | `flagged=false` | `flagged=false` | no (control) |
| 2 — qualifier after | `flagged=true` | `flagged=false` | **yes** |
| 3 — trailing unrelated mention | `flagged=true` | `flagged=true` | **no** |

Case 3's comment claims:

> the plan's adversarial measurement shows whole-line matching would wrongly treat the trailing mention as an exemption qualifier for the earlier, unrelated path.

That does not reproduce. Under whole-line matching the line *is* handed `/otherns:otherskill`, `lookup-skill-dir` *does* resolve it — and then `($target_dir | path join "references/bar.md") | path exists` returns **false**, because the fixture only ever creates `foo.md`. The line stays flagged for a reason unrelated to the residual being pinned. The case asserts a true invariant, but it is not the guard it is documented as.

**The fixture is one line short of making it real.** Adding `bar.md` to the sibling skill — which is precisely the 19-basename collision shape the header comment cites — flips it:

```
[prefix]    case3_AS_SHIPPED_bar_absent    flagged=true
[prefix]    case3_WITH_bar_in_sibling      flagged=true
[wholeline] case3_AS_SHIPPED_bar_absent    flagged=true    ← green, no guard
[wholeline] case3_WITH_bar_in_sibling      flagged=false   ← red, guards correctly
```

Fix: add one line beside the existing `foo.md` save.

```nu
"content" | save ($rd_root | path join "otherns" "skills" "otherskill" "references" "bar.md")
```

This is blocking because it is the PR's own deliverable — a regression guard whose stated measurement is falsified by execution — and because the repo's anti-fabrication posture applies to code comments asserting measured behaviour. Case 2 does carry the widening, so the *suite* is not worthless; case 3 specifically is decorative as written.

**Related, non-blocking:** the second documented false-negative shape (unknown namespace after the path) has no case at all. I verified it would be load-bearing — `flagged=true` → `flagged=false` under the widening. Worth adding as case 4 while the fixture is open.

### [nit] The new fixture leaks a temp dir on every `--self-test`

`let rd_root = (mktemp -d)` at line 1068 is never removed. The file's own sibling fixture (line 1912) cleans up with `rm -rf $root` at line 1995. Measured: temp-dir count `52 → 53` across one `--self-test` run.

### [suggestion] Dropping the advisory footer was the right call

Agreed, and I tried to argue the other side. The print path is unexercisable on a clean corpus, cannot be unit-tested without extracting a function from `main`, and `/core:restraint` rung 1 puts it squarely in "skip it, say so in one line". The author who trips the check now sees `ref_depth` plus `nested:<filename>`, and the header comment names both the two example forms and the fix ("reorder the sentence"). The one place I'd push back mildly: because this residual is a *false positive*, the author's instinct is "the lint is wrong", not "open the validator". If a real instance ever appears, the footer earns its place then. It does not now.

---

## What I attacked that held up

**1. The central measured claim — reproduces exactly.** Sourced the real functions and drove the actual corpus line (`pm/spec-harvest/references/user-story-format.md:7`):

```
PREFIX_ONLY: <This dialect mirrors the qa plugin's Gherkin parser at `plugins/tools/qa/skills/qa/>
prefix     vs skills/X/$ : [{skill: qa}]
whole line vs skills/X/$ : []
prefix     vs /ns:skill  : []
whole line vs /ns:skill  : []
```

End-to-end through `has-unqualified-references-token`:

```
VARIANT=prefix     user-story-format.md flagged? false
VARIANT=wholeline  user-story-format.md flagged? true
```

The `$`-anchor argument is correct: widening does not merely broaden the match, it destroys the `plugins/<...>/skills/<other>/` qualifier form outright. The rationale for the chosen route stands.

**2. Both false-negative shapes — confirmed, and they are stronger than the body claims.** Constructed against the real registry (citing skill `bees`, its own `references/installation.md`):

```
prefix    shapeA_unknown_ns        flagged? true    control_no_mention flagged? true
prefix    shapeB_sibling_basename  flagged? true
wholeline shapeA_unknown_ns        flagged? false   control_no_mention flagged? true
wholeline shapeB_sibling_basename  flagged? false
```

The control stays flagged in both, proving the trailing mention is what does the exempting.

**19 colliding basenames — independently recomputed**, over 218 reference files: `installation.md` ×4; `language.md`, `cli.md`, `commands.md`, `patterns.md`, `examples.md` ×3. Matches the body.

**Correction in the PR's favour:** the body calls these shapes "corpus-plausible". They are not merely plausible — they are **live**. Scanning every reference file, `SKILL.md`, `agents/*.md` and `commands/*.md` for prefix-vs-whole-line divergence found 4 lines, three of which are real false negatives widening would introduce:

| file | path | prefix | whole | citing skill owns path |
|---|---|---|---|---|
| `spec-harvest/references/user-story-format.md` | `references/gherkin-format.md` | exempt | **flagged** | no |
| `spec-harvest/SKILL.md` | `references/gherkin-format.md` | exempt | **flagged** | no |
| `spec-harvest/SKILL.md` | `references/sdlc-checklist.md` | flagged | **exempt** | yes |
| `allium/SKILL.md` | `references/distilling-legacy-code.md` | flagged | **exempt** | yes |

The `allium` one is a nice specimen: `/allium:distill` is a *command*, not a skill, so `lookup-skill-dir` misses, and the `return true` short-circuit exempts a genuine same-skill nested link.

**3. "Zero corpus lines hit the false positive today" — confirmed.** The false positive is `prefix=flagged, whole=exempt, citing skill does NOT own the path`. Across the whole corpus that set is **empty** — both `whole=exempt` rows above have `owns=yes`, i.e. they are the false negatives, not false positives. "Fixes nothing currently broken" is accurate.

**4. The `details` addition — made it fire.** Appended a sibling link to `core/tdd/references/ci-discipline.md`:

```
│ 31 │ tdd │ core │ 344/500 │ 16/17 │ ref_depth │ ref_depth │ nested:ci-discipline.md │
```

Correct filename (`$r.name` is a `path basename`), correct `frontmatter:`-style convention, and it lands in `new` as an unbaselined failure. Reverted cleanly.

**Non-disturbance proved the strong way:** I ran `origin/main`'s validator and the PR's validator over the identical corpus and diffed the full output — **byte-identical**. No change to `frontmatter:`, orphan basenames, broken links, or any `detail_count`. On a clean corpus this PR is a provable no-op.

**5. The fixture genuinely exercises `path exists`.** Verified rather than assumed — pointing the map at a directory where `foo.md` is absent flips case 1 from `flagged=false` to `flagged=true`, so the `path exists` branch is what produces the result, not the `lookup empty → true` short-circuit that `[] []` would give. The test comment's claim here is accurate.

**6. Duplicate-block risk — clear.** Still 4 groups, same four keys (`1182fd06`, `22d40d6a`, `28ccd32d`, `639413bb`), same files, same window counts. Implied by the byte-identical output above.

**7. TDD integrity — clean.** `git diff b4a5369..HEAD -- test/` contains only the two comment blocks plus the single `details` line. No test case was touched by the implementation commit; the `19 → 22 cases` counter moved in the test commit where it belongs. No behaviour change to the check itself.

**8. Every figure in the PR body — all correct.** 19 basenames ✓ · 4 call sites (`ref_depth`, check 14 `links`, Pass-2 agents, Pass-2 commands, all reaching `cross-skill-qualified`) ✓ · 2 check-14 diffs ✓ · 1 Pass-2 diff ✓ (`pm-sdlc-assessor.md` → `references/sdlc-checklist.md`, the only one) · 22 check-fix cases ✓ · 49 waivers ✓ · 4 dupe groups ✓ · 28 validated / 0 pending ✓ · baseline diff empty ✓ · `mise test` exit 0 ✓ · all 8 self-test suites green ✓ · no new `TODO`/`FIXME`/`XXX`/`HACK` markers ✓.

No wrong number in this body — which breaks the epic's streak.

---

Fix case 3's fixture (one line) and the comment that describes it, and this is an approve. The reasoning, the measurement, and the restraint call are all sound; the guard just needs to actually guard.

vinnie357:
## Gate 3 — re-review of `3e91af1`

**Verdict: APPROVE.** The blocking finding is fixed, both non-blocking items are fixed, and I re-ran every experiment against the new head rather than trusting the reported result. Scratch clone only; the shared working tree was never checked out, staged, or committed.

---

### 1. Widening experiment — three cases go red, case 1 correctly stays green

Applied the same one-line widening to a scratch copy of the new head:

```diff
-            let prefix = ($line | str substring 0..<$idx)
+            let prefix = $line
```

```
❌ check-fix self-test: qualifier-after cross-skill pointer no longer flagged (documented residual regressed)
❌ check-fix self-test: same-skill link with trailing same-basename sibling mention wrongly exempted
❌ check-fix self-test: same-skill link with trailing unknown-namespace mention wrongly exempted
```

Isolating all four cases under both variants:

| case | shipped (prefix) | want | widened (whole-line) | role |
|---|---|---|---|---|
| 1 — qualifier before | `false` | `false` ✓ | `false` | **control, correctly green** |
| 2 — qualifier after | `true` | `true` ✓ | `false` | **red** |
| 3 — trailing same-basename sibling | `true` | `true` ✓ | `false` | **red** |
| 4 — trailing unknown namespace | `true` | `true` ✓ | `false` | **red** |

Case 1 stays green in both columns, which is exactly what a control should do — it proves the widening does not break the legitimate qualifier-before exemption, so the three reds are attributable to the residual and not to collateral damage.

### 2. The fixture change weakens nothing — ablation, not inspection

I ran the full 4-case matrix twice, once with `bar.md` present and once absent, under both variants:

```
[prefix]    bar_PRESENT  case1=false case2=true  case3=true  case4=true
[prefix]    bar_ABSENT   case1=false case2=true  case3=true  case4=true
[wholeline] bar_PRESENT  case1=false case2=false case3=false case4=false
[wholeline] bar_ABSENT   case1=false case2=false case3=true  case4=false
```

Two things fall out:

- **Under the shipped code the two rows are identical.** Adding `bar.md` changes no case's outcome, so it cannot make anything pass for the wrong reason. That is structural, not coincidental: with prefix-only matching the trailing qualifier is never examined, so `bar.md`'s existence is inert on the shipped path. Its *only* effect is under the widened variant.
- **`bar_ABSENT` → `case3=true` under whole-line** is the clean ablation: without that one file case 3 would still be green under the widening, exactly as I found last round. `bar.md` is precisely and solely what makes case 3 load-bearing.

Pre-existing `[] []` ref_depth cases re-checked and unmoved (`heading=false`, `plain=true`) — they don't touch the fixture.

**Corpus-level non-disturbance proved the strong way again:** ran `origin/main`'s validator and this head's validator over the identical corpus and diffed full output — **byte-identical**. On a clean corpus this PR remains a provable no-op.

### 3. Cleanup ordering is correct

`rm -rf $rd_root` sits at line 1125; the four cases that need the fixture are at 1079, 1088, 1103, 1120 — all before it. I also grepped for any `rd_root`/`rd_map`/`rd_plugins` use *after* the cleanup: none. The `invocations` block that follows uses only `$allium_cmds`/`$reg`.

Leak measured empirically rather than assumed: temp-dir count **56 → 56, delta 0** across a `--self-test` run (was delta 1 before the fix).

### Both new case rationales verified against the code

- **Case 3**: comment now describes the 19-colliding-basenames shape, and the ablation above confirms the stated mechanism (widening resolves `otherskill`, finds *its* `bar.md`, wrongly exempts a path that never named it).
- **Case 4**: the empty-target branch it cites is real (`if ($target_dir | is-empty) { return true }`, lines 128–130), and the premise is genuinely stable — `openai` is not a plugin in `marketplace.json` and `codex` is not a skill anywhere in the repo, so `lookup-skill-dir` returns `""` for the right reason. Confirmed it needs no fixture file: `case4=true` under prefix in both ablation rows.

### TDD integrity of the fix commit

Every hunk in `3e91af1` lands inside `run-check-fixes-self-test`:

```
@@ -1058 +1058 @@ def run-check-fixes-self-test [] {
@@ -1067 +1067,3 @@ ...   @@ -1070,0 +1073 @@ ...   @@ -1090,6 +1093,10 @@ ...
@@ -1097 +1104 @@ ...   @@ -1100,0 +1108,19 @@ ...   @@ -1125 +1151 @@ ...
```

All five production functions hash-identical to `52d0ba9` (`has-unqualified-references-token`, `cross-skill-qualified`, `preceding-line`, `lookup-skill-dir`, `resolve-pass2-path`), and the `details` line is untouched. A `test:`-scoped commit that is genuinely test-scoped.

### Gates

`mise test` **exit 0** · all 8 self-test suites green · check-fix **23 cases** · waivers **49** · duplicate blocks **4 groups**, same four keys · corpus **28 validated / 0 pending** · baseline diff empty · diff still confined to `test/validate-skills-quality.nu` so no plugin version bump · no new `TODO`/`FIXME`/`XXX`/`HACK` markers.

### One nit, not worth a round-trip

Case 3's comment says `references/bar.md` is "owned by THIS skill, `myskill`" — `myskill` has no fixture directory, and the check never verifies that the citing skill owns the path. It's a description of the shape being modeled rather than a filesystem fact, and case 4's "no fixture file needed" makes the convention clear, so it reads fine in context. Mentioning only for completeness.

---

Nothing left blocking. The guards now fail under the change they exist to prevent, the second documented false-negative shape is covered, the fixture is inert on the shipped path, and the corpus behaviour is unchanged. Good fix.

PR #187, opened 2026-07-31
feat(sources): add b3_version_shape validation for current_version

vinnie357:
Closes claude-skills-189, filed from the Gate 3 review on #178 and proven there with a fixture and a control.

## The gap

`current_version` had **no format validation at all**. It appeared in `test/validate-sources.nu` only in the allowed-keys list, `ENTRY_NETWORK_KEYS`, and `NONE_FORBIDDEN` — no regex, no check. `current_version = "banana"` passed silently, as would a typo, a stray space, or a pasted URL. `last_checked` has been properly checked by `b3_date` the whole time; #178 documented that asymmetry, and documenting a gap is not closing it.

## Why a semver regex would have been wrong

The corpus carries **193 entries with `current_version`, 40 distinct values**. The 39 non-`unknown` ones are genuinely heterogeneous:

```
0.0.7  0.11.0  0.113.1  1.0.103  1.1  2.1  2025  28.4.1  3  3.0
v0.0.0  v0.1.0  v0.4.6  v2026.3.15   … and 25 more
```

Bare majors (`3`), a year (`2025`), major.minor (`1.1`, `3.0`), and `v`-prefixed CalVer (`v2026.3.15`). A strict semver check rejects all of those. And **152 of the 193 entries — 79% — use the literal `unknown`**, the sanctioned way to record "could not establish this" rather than guessing. Any rule must special-case it or break the corpus at once.

## The contract

`b3_version_shape` accepts `unknown`, or:

```
^[A-Za-z]{0,10}[-.]?\d+(\.\d+){0,4}([-+][0-9A-Za-z.]+)?$
```

Optional short alpha prefix, 1–5 dot-separated numeric groups, optional `-prerelease` or `+build` suffix. Shape-based, deliberately not semver-conformant.

**Verified against the live corpus before implementation and again after: 0 of 39 distinct values rejected.** Rejected: `banana`, `1.0.0 beta`, a pasted URL, the empty string, `see changelog`.

**The contract is deliberately wider than the data.** Accept-cases were written for shapes **absent** from today's corpus — `OTP-29.0.4`, `1.0.0-rc1`, `1.0.0+build.5` — because a rule fitted to exactly the 39 present values would pass CI today and reject the next honest entry, with nobody able to tell whether that rejection was intended.

## A claim in the bee was wrong

The bee cites `OTP-29.0.4` as a corpus shape justifying a permissive check. It is not one — that entry stores `28.4.1`, and the `OTP-` prefix lives in `notes`. The test author checked rather than designing around it, and added it as an accept-case anyway on its own merits, since a future entry may legitimately use that shape.

## TOML numerics, resolved by measurement

`current_version = 2025` parses as an **int** and `1.0` as a **float** — not strings. The file already carries a `scalar-str` helper for exactly this. Measured: `1.0` coerces to `"1"` (nushell drops the trailing `.0`). Both are pinned as passing rather than treated as type errors, consistent with how every other field in the file behaves.

## Verified by making it fail

A green run on a clean corpus does not prove a new rule fires. The implementer corrupted `core/nushell`'s pin to `banana`; I independently corrupted a different entry, `qa/gherkin-cucumber`, to `see changelog`:

```
❌ 1 finding(s):
│ b3_version_shape │ qa/gherkin-cucumber: current_version 'see changelog' …
```

Both reverted clean, no residue.

## Doc and check now agree

The `current_version` row in `skill-update`'s field table read `string | unknown`. It now names the shape, the rule, and gives both the accepted forms (CalVer, OTP-style, bare majors, pre-release/build) and the rejected ones (whitespace, URLs, prose). The bee required this explicitly — a check the documentation misdescribes is the failure mode this epic has hit repeatedly.

## Gates

`mise test` exit 0; corpus 28 validated / 0 pending; sources self-test green; all 8 skills-quality suites green; waivers **49 — unchanged**; `test/quality-baseline.json` diff empty; `mise run test:version-bumps origin/main` exit 0 with `claude-code` bumped 0.2.41 → 0.2.42; gitleaks clean.

## Gate 3 — approved, two findings addressed

**A wrong number in this body, corrected.** It claimed 64 entries use `unknown`. The real figure is **152** (confirmed two ways: structural parse and grep). 64 was correct when bee 189 was filed — the corpus was 109 entries across 20 plugins then, and is 193 across 28 now. I carried the figure forward instead of re-measuring. The error understated my own argument: at 79% rather than 59%, special-casing the literal is more load-bearing, not less. Fourth wrong count an epic PR body has shipped, and the reason reviewers keep being asked to recompute rather than accept.

**A missing test, added.** `b3_version_shape` has two reachable paths — a `scalar-str` null guard for non-coercible types, and the shape check. All 13 original cases exercised only the shape path, while the rule it was modeled on (`b3_date`) carries two wrong-type cases, both added after this repo was bitten by that class.

Two cases added. The mutation testing behind them turned up a distinction worth recording:

- `current_version = ["1.0"]` routes through the **null guard**. Neutralizing that guard does not merely turn the case red — it **crashes the whole self-test** (`'=~' does not work on values of type 'nothing'`). I reproduced this independently. That guard is a crash guard, and this case is what holds it in place.
- `current_version = 2026-01-01` (unquoted) does **not** reach the null guard. `datetime` *is* in `scalar-str`'s coercible set, so it becomes a string and fails the shape check instead. The test author reported this rather than claiming both cases pin the null branch, and verified by isolated probe that deleting the null guard has zero effect on it. It still earns its place — deleting the whole check turns it red — and it is the sneakier shape, since an unquoted date is a plausible thing to type meaning a version.

Sources self-test 61 → **63 cases**.


vinnie357:

**Verdict: APPROVE.** No blocking findings. Two should-fixes below, neither requiring re-review.

I went in assuming a defect existed. What follows is what I attacked and what survived, so the negative result carries weight. All work done in a throwaway clone (`gh pr diff` / `git show` / a scratch clone) — the shared working tree was never checked out or mutated.

---

### Blocking

None.

---

### Findings

**F1 — [should-fix, non-blocking] The body's `unknown` count is wrong: 64 → actually 152.**

> "**64 entries use the literal `unknown`**"

Measured two independent ways against the branch:

- structural parse of all 28 `sources.toml` → `current_version == "unknown"` in **152** entries
- `grep -c 'current_version *= *"unknown"'` → **152**

No other metric equals 64 either: `last_checked == "unknown"` is 35, `check_method == "none"` is 13, entries with a *non*-`unknown` `current_version` is 41, entries lacking the key is 13. So it is not a mislabeled figure — it is simply wrong.

Worth stressing: **the error understates the evidence for the design.** 152 of 193 entries (79%) carry `unknown`, so special-casing the literal is even more load-bearing than the body argues. The conclusion is unaffected; only the number needs correcting. Given this epic's own thesis is that the written record must match the measured reality, please fix the body before merge.

**F2 — [should-fix, non-blocking] The `$cv == null` branch has no test, and its sibling rule has two.**

`b3_version_shape` guards non-string TOML values via `scalar-str` returning null. That guard is correct — I verified both reachable paths by execution against a corrupted entry:

- `current_version = ["1.0"]` → `rust/official-rust-documentation: current_version is a list<string>, expected …`
- `current_version = 2025-01-01` (bare TOML date → `datetime`) → fires

But **no frozen test case exercises it.** All 13 new cases produce a non-null `scalar-str` (int and float both coerce). Meanwhile the rule it was modeled on, `b3_date`, has *two* such cases — line 821 `"entry last_checked as a list must report, not crash =~ (claude-skills-185 rd2)"` and line 1639 `"A-F4.3: TOML-native unquoted date must not crash =~"` — both filed *after* this repo was bitten by exactly that class.

The consequence is concrete. Dropping the guard is not a soft failure:

```
Error: nu::parser::operator_unsupported_type
  x The '=~' operator does not work on values of type 'list<string>'.
```

That aborts the whole validator, and today nothing would catch the regression. One case mirroring line 821 closes it. Not blocking — the shipped behavior is right — but this is the one gap I'd actually want closed.

**F3 — [non-blocking] False acceptances, quantified — and forced, not lazy.**

Junk that still passes: `version1`, `changelog1`, `unknown1`, `TBD1`, `a1`, `x-1`, `0`, `00000`, plus the degenerate leading-separator forms `-1` and `.1`.

I tried to make this a blocking finding by proposing the obvious tightening — require the separator when a prefix is present, `^([A-Za-z]{1,10}[-.])?\d+(\.\d+){0,4}([-+][0-9A-Za-z.]+)?$` — and it fails: that variant **rejects real corpus values** `v2026.3.15`, `v0.1.0`, `v0.0.0`, `v0.4.6`. The bare-alpha-prefix allowance is precisely what those `v`-prefixed pins need, and prose-with-a-trailing-digit is the unavoidable price. The permissiveness is data-driven. `-1` and `.1` could be closed by making `[-.]?` conditional on a non-empty prefix, but neither is a plausible typo. Leaving this as taste.

**F4 — [non-blocking] False rejections, judged individually.**

Rejected: `R2023b` (MATLAB), `1.2.3rc1` (PEP440), `8u402` (Java), `12.0.0.RELEASE` and `1.0.0.Final` (Spring/JBoss), `1.2.3_4`, 6-group `1.0.0.0.0.0`, a 40-char git sha, and `latest` / `main` / `HEAD` / `unstable`.

Most are correctly rejected or irrelevant — and rejecting `latest`/`HEAD` is *right*, since `unknown` is the sanctioned way to say that. The one I'd flag as realistic is **`1.2.3rc1`**: PEP440 permits a pre-release with no separator, and the corpus does track a PyPI upstream (`plugins/tools/graphify` — package `graphifyy`, currently pinned `0.8.36`, notes record live PyPI at `0.9.31`). A future `0.9.31rc1` pin would be rejected and would need a *rule* change, not an entry edit. Not blocking: nothing in the corpus uses that shape today and `unknown` + a note is an honest fallback. Just worth knowing the ceiling.

**F5 — [non-blocking] `1.0` → `"1"` is right on consistency grounds, but the float path is lossy.**

All three body claims verified by execution: `current_version = 2025` parses `int`; `1.0` parses `float`; `scalar-str` yields `"1"`; both pass.

Going further than the body did: `1.10` → `"1.1"` and `0.30` → `"0.3"`. An unquoted float silently becomes a **different version**.

Was pinning `1.0` as passing the correct call? **Yes.** `scalar-str` is what every other field already uses, and the corpus has **zero** non-string `current_version` values (measured), so this is latent, not active. But note the exposure differs from `b3_date`: an unquoted date becomes a `datetime`, hits the null path, and *fails loudly* — an unquoted float passes quietly. If it ever bites, the fix is a "must be a quoted string" rule, not a regex change. Out of scope here.

**F6 — [nit] `{0,10}` is arbitrary.** The only prefix in the corpus is `v` (1 char); accept-cases add `OTP` (3). The bound's sole observable effect is the meaningless line between `abcdefghij1` (accepted) and `abcdefghijk1` (rejected). Harmless — excluding whitespace is what actually blocks prose.

---

### What I attacked that held up

- **Rule fires where it should.** Corrupted a **third** plugin, untouched by prior verification — `plugins/languages/rust`, entry `official-rust-documentation` → `"release notes"`. `b3_version_shape` named it. Restored; `git status` clean.
- **Does not fire where it shouldn't.** 13 `check_method = "none"` entries legitimately omit `current_version`; clean baseline of 28 files / 0 pending confirms no misfire on absence.
- **No interference with `b4_conditional` / `b4_none_forbidden`.** Adding a bad `current_version` to a `none` entry produced exactly **2** findings — one per rule, no duplication, no masking. Correct: two genuinely distinct problems.
- **Corpus claim re-derived independently:** 193 entries with `current_version`, 40 distinct, 39 non-`unknown`, **0 rejected**. Matches the body.
- **TDD boundary intact.** The implementation commit's only hunks on the test file are at lines 40 and 374 (the const and the check block); the 13 new cases live at 1661–1972 and were not touched. Test and implementation share a file, so I verified by hunk position rather than by an empty diff.
- **Red proof.** Ran the frozen test file from `139697f` standalone: exactly the 4 negative cases fail (`want [b3_version_shape], got []`), accept-cases pass. The negatives genuinely drove the implementation; the accept-cases are regression guards, which is the right split.
- **Doc/check agreement — verified by byte comparison, not by eye.** Extracted the regex from `SKILL.md` and from `validate-sources.nu` and compared: **byte-identical**. Every prose claim checked against the probe: 1–5 numeric groups ✓, accepts CalVer / OTP-style / bare majors / pre-release / build ✓, rejects whitespace / URLs / prose ✓. The stale "`current_version` has no format check" sentence is correctly removed.
- **Integrity.** Waivers **49**; `test/quality-baseline.json` untouched; the only JSON change is `0.2.41 → 0.2.42` in *both* `plugin.json` and the marketplace entry, in sync; trailing bytes on both JSON files identical to `origin/main`, both text files end `0a`; `mise test` **exit 0**; disclosure lint clean (614 tracked files); `mise run test:version-bumps origin/main` passes.
- **Body figures.** 193 ✓, 40 ✓, 39 ✓, 0 rejected ✓, 61 self-test cases ✓ (the runner prints 61; the 3 extra `label:` lines belong to the pre-existing `missing_cases` group, not this PR), 28 validated / 0 pending ✓, 49 waivers ✓. Only F1 is wrong.
- **Restraint.** One const plus one 17-line block that mirrors the adjacent `b3_date` block exactly, reusing the existing `scalar-str` helper. No new abstraction, no new dependency. Stops at the right rung.

The PR's self-criticism also checks out: the bee's `OTP-29.0.4` claim really is absent from the corpus (that entry stores `28.4.1`, with `OTP-` living in `notes`), and the body says so rather than designing around it.

vinnie357:
## Gate 3 delta confirmation — APPROVE

Narrow re-check of the delta only (`581d8fe` + the body correction) against the existing Gate 3 approval. Not a re-review; the earlier verdict stands. Run in a scratch clone at `581d8fe` — the shared working tree was never checked out.

**1. The list case holds the null guard — confirmed, and it is a crash guard.**

Neutralizing only line 379 (`if $cv == null {` → `if false {`):

```
Error: nu::shell::operator_unsupported_type
  x The '=~' operator does not work on values of type 'nothing'.
 385 |  } else if not ($cv == "unknown" or ($cv =~ $VERSION_SHAPE_RE)) {
     :                                       ^|^ ^|
     :                                        |   `-- does not support 'nothing'
```

The whole self-test aborts — no case count is printed at all, so this is a hard failure, not a red case. Attribution checked: with the guard still neutralized, changing *only* that case's value from `["1.0"]` to `"banana"` restores `✅ sources self-test passed (63 cases)`. The `["1.0"]` case is what holds the guard, and nothing else in the suite does.

**2. The datetime case routes to the shape branch, not the null branch — confirmed.**

Isolated `check-sources` probe:

```
RAW TYPE: datetime
scalar-str => string: Thu Jan  1 00:00:00 2026
demo/demo-source: current_version 'Thu Jan  1 00:00:00 2026' is neither a recognizable version string nor 'unknown'
```

`datetime` is in `scalar-str`'s coercible set, so it returns a string and the null branch is never reached. Contrast, same probe on the list case:

```
RAW TYPE: list<string>
scalar-str => nothing
demo/demo-source: current_version is a list<string>, expected a version string or 'unknown'
```

Two distinct messages, two distinct branches. Independently corroborated by (1): with the null guard neutralized the datetime case still passes, so it has no dependency on it. The body's account is accurate.

**3. The corrected figure — confirmed, 152 of 193.**

Re-derived two ways, templates excluded (`skill-update/templates/sources.toml` carries 1 entry and is not corpus — it is the source of a naive 194):

| method | files | with `current_version` | `unknown` | non-`unknown` |
|---|---|---|---|---|
| structural TOML parse | 28 | 193 | 152 | 41 |
| grep `^current_version *=` | 28 | 193 | 152 | 41 |

78.8% (body says 79%), 39 distinct non-`unknown` values, 40 including the literal. Every figure in the body checks out.

**4. No regression from the delta — confirmed.**

`git diff 2fd2dd1..581d8fe` is `test/validate-sources.nu | 60 ++++++`, insertions only, entirely inside the case table — no production logic touched. `mise test` exit **0**; `sources.toml validation clean (28 file(s) validated, 0 pending)`; sources self-test 61 (`2fd2dd1`) → **63** (`581d8fe`), matching the claim; `49 debt/bug entries to burn down` — unchanged; `test/quality-baseline.json` diff empty across the whole PR and `git status` clean after the run, so nothing was rewritten in place.

### On the `{0,10}` bound and the false-acceptance surface

Probed the regex directly rather than leaving it as a judgement call.

**Junk it admits — taste, not a bug.** `banana1`, `version1`, `see1`, `TBD1` all pass: any alpha token ending in a digit satisfies `[A-Za-z]{0,10}\d+`. But every realistic authoring mistake is still caught — `banana`, `see changelog`, `1.0.0 beta`, `""`, a pasted URL, and the plausible `unknown` fat-finger `unkown` are all rejected. Nothing a person actually types slips through. Not worth a rule change.

**The `{0,10}` bound — taste.** It rejects `elasticsearch-8.11` (13-char prefix) and accepts `PostgreSQL-16.1` (exactly 10). Product-name prefixes belong in `name`, not `current_version`, so the bound is not load-bearing.

**One thing marginally above taste, for the record only.** The suffix group requires a `-`/`+` separator, so PEP 440 forms without one are rejected:

| input | accepted |
|---|---|
| `1.0.0-rc1` | yes |
| `3.13.0rc1` | **no** |
| `2.0b1` | **no** |
| `1.0.post1` | **no** |
| `1.0.0-rc.1+build.5` | **no** |

`3.13.0rc1` is a genuine CPython release string, and one hyphen is the only thing separating it from the accepted form — an author who hits it would reasonably be surprised. Mitigating, and why this is a note rather than a finding: no corpus entry pins a pre-release; the check is `severity: "fail"` so it blocks loudly rather than corrupting silently; and the SKILL.md row now publishes the regex verbatim, so a rejected author reads the rule instead of reverse-engineering it. `1.0.0-rc.1+build.5` being rejected is consistent with the body's explicit "deliberately not semver-conformant".

No change requested. Recording it so a future rejection is a known edge rather than a mystery.

**Verdict: APPROVE.** All four delta claims hold under execution.

vinnie357:
## Gate 3 — round 2 (commit `581d8fe`)

**Verdict: APPROVE.** Both round-1 findings closed, verified by my own mutation testing rather than by report. One follow-up worth filing, plus a correction to my own round-1 review.

---

### F1 — closed, and the historical account checks out

Body now reads "**152 of the 193 entries — 79% — use the literal `unknown`**". Matches my measurement (152/193 = 78.8%).

I also verified the new historical claim rather than taking it on faith. Checked out `c3b5706` (PR #178 — the commit where bee 189 was filed) and re-ran the count:

```
files: 20 | entries: 109 | with current_version: 99 | == unknown: 64
```

**Exactly 20 plugins, 109 entries, 64.** The stated cause — figure accurate when filed, corpus grew via Wave B (#179/#180) to 28/193, carried forward without re-measuring — is correct in every particular. Good account.

### F2 — closed. Both cases are load-bearing; here is the mutation evidence

**The list case holds the null guard — confirmed, and isolated.**

- Deleted the `$cv == null` branch, leaving only the shape check → the run does not go red, it **crashes**: `The '=~' operator does not work on values of type 'nothing'`. Same as reported.
- Then the decisive test: deleted the null guard **and** the list case → `✅ sources self-test passed (62 cases)`.

So the list case is precisely what holds that guard — nothing else in 62 cases does. That is the strongest form of the claim and it holds.

**The datetime case is load-bearing for the shape branch.** Deleted the shape branch, keeping the null branch → the datetime case goes red alongside the four pre-existing negatives (`banana`, whitespace, URL, empty).

**One honest qualification, non-blocking.** I mutated `scalar-str` to drop `datetime` from its coercible set — which would reroute the case from branch B to branch A — and the suite **survives** (`63 cases` pass). That is a harness limitation, not an authoring miss: the comparison is on rule *names*, and both branches emit `b3_version_shape`, so no test can distinguish the routes. The comment documents the routing; the suite cannot pin it. Worth knowing, nothing to fix here.

### The datetime routing — confirmed, and it corrects my round-1 review

The test author is right, and I was wrong in round 1.

In round 1 I probed with a **reimplementation** of `scalar-str` rather than the real one, and gave it a narrower coercible set (`int`/`float`/`bool`). The real helper at line 64 is:

```
["int" "bool" "float" "datetime" "filesize" "duration"]
```

`datetime` is in it. Re-run against the actual helper:

```
2026-01-01 -> describe=datetime  scalar-str="Thu Jan  1 00:00:00 2026"  null?=false
             shape-regex accepts: false | date-regex accepts: false
["1.0"]    -> describe=list<string>  scalar-str=null  null?=true
```

So an unquoted date coerces to `"Thu Jan  1 00:00:00 2026"` and fails the **shape** check — branch B, exactly as the test comment states, down to the rendered string.

This corrects a statement in my round-1 F5, where I wrote that an unquoted date "parses as `datetime` → `scalar-str` null → `b3_date` fails loudly". The conclusion (fails loudly) stands; the mechanism I gave did not. `b3_date`'s unquoted-date case also travels branch B, not the null branch. My round-1 F5 recommendation is unaffected — a quoted-string rule is still the fix if the lossy float path ever matters.

### Does anything in the regex rise above taste?

Asked directly, so answering directly.

**Yes, one thing: `1.2.3rc1` (PEP440).** This is a realistic version string the rule rejects. PEP440 permits a pre-release with no separator (`1.2.3rc1`, `1.0a1`, `2.0b2`), and the corpus **does** track a PyPI upstream — `plugins/tools/graphify`, package `graphifyy`, pinned `0.8.36`, with notes recording live PyPI at `0.9.31`. A future pre-release pin there would be rejected, and the author would have to misrecord it or fall back to `unknown`. Worth a follow-up bee. Still not blocking: no corpus entry uses that shape today, the failure is a loud CI stop rather than silent corruption, and widening the regex to admit a trailing alpha run costs false-acceptance surface — a real trade-off that deserves its own ticket, not a rushed edit here.

**No, on `-1` / `.1`.** Junk, but not plausible authoring mistakes, so it stays taste. That said, if you want it closed cheaply, I tested a one-character-class change that makes the separator conditional on the prefix:

```
^([A-Za-z]{1,10}[-.]?)?\d+(\.\d+){0,4}([-+][0-9A-Za-z.]+)?$
```

Measured against the live corpus and the existing cases: **rejects `-1` and `.1`; zero regressions** — 0 of 39 corpus values wrongly rejected, 0 of 8 accept-cases wrongly rejected, 0 of 5 reject-cases wrongly accepted. Strictly tighter at no cost. Optional — it would mean re-freezing tests for two new negatives, and the current rule is not wrong, just marginally loose.

**No, on `{0,10}`.** Still arbitrary, still harmless. Its only observable effect is the meaningless boundary between `abcdefghij1` and `abcdefghijk1`. Excluding whitespace is what actually blocks prose. Not worth a ticket.

### Gates re-verified on `581d8fe`

`mise test` **exit 0**; sources self-test **63 cases**; corpus **28 validated / 0 pending**; waivers **49**; disclosure lint clean (614 files); `test/quality-baseline.json` untouched; changed-file set still exactly four (two JSON version bumps, `SKILL.md`, `validate-sources.nu`).

**Test/implementation boundary intact:** `git diff 2fd2dd1..581d8fe` is `test/validate-sources.nu | 60 ++++`, insertions only, single hunk at line 1964 — inside the case list, nowhere near the production block at 377–392. No production logic touched by the test commit, as stated.

vinnie357:
## Gate 3 — third pass (post-merge note)

I was dispatched to review this PR at `2fd2dd1` and worked from that revision before noticing it had already merged as `b16b5f5`, with two Gate 3 rounds and the wrong-type commit `581d8fe` already landed. So most of what I attacked was ground the earlier rounds had covered. Recording only what is **new and still live on `main`**, plus the negative results, so the coverage is on record.

### New finding — the annotated template documents `current_version` values the merged rule rejects

Not covered by claude-skills-196 or -197. Verified against `origin/main` (`b16b5f5`), not the branch.

`SKILL.md:24` sends authors to the template: *"See `templates/sources.toml` for the complete annotated template."* That template (`plugins/tools/claude-code/skills/skill-update/templates/sources.toml:48-52`, unchanged by this PR) says:

```
# The upstream version this skill's content was last verified or documented
# against. Exact version string (e.g. "0.10.0"), or "stable"/"rolling" for
# those constraints, or "unknown" when nobody has measured it yet.
# A guessed version is fabrication - use "unknown".
current_version = ""
```

Measured against the merged `VERSION_SHAPE_RE`:

```
'stable'  -> REJECTED by b3_version_shape
'rolling' -> REJECTED by b3_version_shape
''        -> REJECTED by b3_version_shape
'unknown' -> ACCEPT
'0.10.0'  -> ACCEPT
```

Three things in one file:

1. `"stable"` / `"rolling"` are a sanctioned convention — they pair with `VERSION_CONSTRAINTS = ["pre-1.0" "semver" "rolling" "stable"]`, and the comment explicitly says to use them "for those constraints". Both now fail validation.
2. The template ships `current_version = ""` as its **default**, and this PR pins the empty string as a deliberate reject case. Copy the template, fill in everything else, and `mise test` fails on a line the template gave you.
3. The template is now internally inconsistent: `last_checked = "unknown"` *is* a value `b3_date` accepts, so the date field uses the sanctioned literal while the version field ships one its own validator rejects.

Corpus impact today is zero — no entry uses `stable`/`rolling` as a `current_version`, so CI is green. That is what makes it latent rather than visible. Nothing catches it automatically either: `validate-sources.nu` globs `plugins/**/skills/sources.toml`, and the template's parent directory is `templates/`, so it is not among the 28 files validated.

This is the failure class the PR body names — *"a check the documentation misdescribes is the failure mode this epic has hit repeatedly."* The field table in `SKILL.md` was updated; the template it points at was not.

Fix is a choice, not a defect hunt: either change the template prose + default to `unknown`, or accept `stable`/`rolling` as literals alongside `unknown`. Worth a bee alongside 196/197.

### Increment to claude-skills-196

196 frames the gap as PEP 440 forms carrying no separator (`3.13.0rc1`, `2.0b1`, `1.0.post1`) and lists `1.0.0-rc.1+build.5`. There is a second, distinct cause worth adding, because these are **valid strict SemVer** rather than PEP 440:

| value | b3_version_shape | strict SemVer? |
|---|---|---|
| `1.0.0-alpha-1` | reject | yes |
| `1.0.0-rc-2` | reject | yes |
| `2.0.0-rc.1+exp.sha.5114f85` | reject | yes (semver.org's own example) |

Cause: the suffix class `[0-9A-Za-z.]` omits `-`, so a hyphen *inside* a prerelease identifier fails. `-rc-1` / `-alpha-1` / `-beta-2` are common tags.

The sharp part: `SEMVER_RE`, declared one line above `VERSION_SHAPE_RE` in the same file, uses `[0-9A-Za-z.-]` — it *includes* the hyphen. A rule introduced as "deliberately wider than semver" is, on this axis, narrower than the strict-semver const sitting directly above it. `([-+][0-9A-Za-z.-]+)?` closes it.

### Negative results — what I attacked that held up

Independent of the earlier rounds, on the merged rule:

- **Fires where it should.** Corrupted `rust/tokio` — a third target, neither `core/nushell` (implementer) nor `qa/gherkin-cucumber` (round-1 reviewer): `b3_version_shape │ rust/tokio: current_version '1.53 (see changelog)' is …`. Reverted clean.
- **No false positive on absent `current_version`, mechanism verified.** The risk was nushell unifying columns across heterogeneous `[[sources]]` records, giving `none` entries a null `current_version`. It doesn't — `let cols = ($entry | columns)` is per-record. Probed `claude-code/sources.toml` (34 mixed entries): *none entries exposing a current_version column: 0*.
- **Clean b1/b3/b4 separation,** each by mutation: key deleted → `b1_entry_missing` only; valid version on a `none` entry → `b4_none_forbidden` only; invalid version on a `none` entry → both (two true statements about one line; noise, not a defect).
- **TDD integrity.** `git diff 139697f..2fd2dd1 -- test/validate-sources.nu` is +21 lines, entirely const + rule block, **zero test-case lines changed**. The frozen test commit runs red with exactly the 4 negative cases failing, each naming `b3_version_shape`; 9 accept-cases green. Genuine red→green.
- **Corpus re-derived independently:** 193 entries, 40 distinct, 39 distinct non-unknown, **0 of 39 rejected**, 152 `unknown` — all matching the corrected body.
- **Doc/check regex agreement is byte-exact:** extracted the const from the code and the regex from the SKILL.md prose, compared → identical. (The prose *around* it is what overstates.)
- **`{0,10}` prefix bound is adequate:** `abcdefghij1` (10) accepts, `abcdefghijk1` (11) rejects; covers `v`, `OTP`, `release`, `jdk`, `python`, `R`.
- **Residual false acceptances** — `version1`, `banana1`, `a1`, `x-1`, `0`, `00000`, and (from the empty-alpha-prefix path) bare `-1` and `.1`. All require a digit and no whitespace, so the targeted class — prose, whitespace, URLs, empty — is genuinely caught. Not worth tightening.
- **Integrity on the merged tree:** `mise test` exit 0; sources 28 validated / 0 pending; self-test 61 cases; waivers 49 unchanged; `quality-baseline.json` diff empty; `test:version-bumps origin/main` exit 0 with `claude-code 0.2.41 → 0.2.42`; plugin.json == marketplace entry; `jq del(.version)` diff empty on both JSON files; trailing bytes identical to `origin/main` on all four files; gitleaks `no leaks found` across 308 commits.
- **Restraint:** the new block mirrors the adjacent `b3_date` block and reuses the existing `scalar-str` helper instead of adding a parallel coercion path. Rung-2 reuse, correctly taken.

The implementation is sound and the TDD discipline is clean. The template disagreement is the one thing I'd still act on.

PR #188, opened 2026-08-01
feat(sources): align the version-shape suffix with semver, require a quoted string, and guard the template

vinnie357:
Closes claude-skills-196, 197, and 198 — three bees that are one question: **what exactly is a valid `current_version`?** They arrived as "too narrow", "too permissive about types", and "disagrees with its own docs", and they interact, so they were planned and decided together rather than in three rounds of gates.

## Two premises in the bees were wrong

**I wrote one of them.** Bee 198 says `validate-sources.nu` globs `plugins/**/skills/sources.toml`. It does not — `main` reads `marketplace.json` and opens each plugin's file, so the file set is marketplace-driven and **"extend the glob" was never an available move**. I transcribed that from a reviewer's report without checking it.

Second: `sources-init.nu` already emits `current_version = "unknown"`, a value the rule accepts. So the sanctioned generator produces a valid file today, and 198's defect is reachable only by hand-copying the template — real, but narrower than the bee implies, and it argues for a doc fix over new enforcement.

## Five decisions, four measured, two of them "no"

**D1b — hyphen inside a prerelease: FIXED**, then corrected after review (see the Gate 3 section — the first attempt admitted calendar dates). This is not a policy widening: `SEMVER_RE` sits four lines above with `[0-9A-Za-z.-]` and **already accepts** `1.0.0-alpha-1` and `1.0.0-rc-2`. The rule documented as "deliberately wider than semver" was narrower than the strict-semver constant above it. Measured: 0 of 39 corpus values change acceptance.

**D1a — PEP 440 separator-less (`3.13.0rc1`, `2.0b1`, `1.0.post1`): WON'T FIX.** The cost curve is sharply non-linear — D1b admits nothing extra; the PEP 440 alternative admits **six** junk strings including `1.x`, which is a version *constraint*, not a version, and precisely the silently-wrong pin this rule exists to block. And the demand is absent rather than dormant: `CHECK_METHODS` has **no PyPI method at all**, **0** corpus entries name a Python upstream, and the one PyPI-adjacent entry (`graphify`) uses plain versions. Recorded in the docs so a future rejection is a known edge.

**D1c — dual `-prerelease`+`+build`: WON'T FIX.** Rejected by `SEMVER_RE` too, which permits only one suffix group. Accepting it in `VERSION_SHAPE_RE` alone would recreate the D1b contradiction inverted. If ever wanted, both constants move together.

**D2 — are `stable`/`rolling` legal `current_version` values? No.** **0 of 193** entries use them there, and the positive convention already exists: entries whose `version_constraint` is `rolling` record `current_version = "unknown"` in **117 of 119** cases. `unknown` *is* the sanctioned way to say "no discrete version to pin". The template was simply wrong.

**D3 — unquoted numerics now rejected.** An unquoted TOML numeric was passing *silently while recording a different version*: `1.10` → `"1.1"`, `0.30` → `"0.3"`, `1.0` → `"1"`.

The scoping argument is the interesting part. `current_version` is the **only** `scalar-str` consumer whose check silently accepts a coerced numeric — `DATE_RE` needs `YYYY-MM-DD`, `SEMVER_RE` needs three numeric components (so `"1.1"` fails), and both enums need exact membership. Every other field gets a type guard for free. So diverging this one field from the shared helper is not an inconsistency; it is the one field missing the guard the others already have. `scalar-str` itself is untouched — it has 7 call sites on this branch (8 on `main`; this PR removed one).

All non-strings are rejected, not just floats: `0x10` → `"16"`, `1_000` → `"1000"`, `1e3` → `"1000"` are lossy too, and the published schema already declares the field a string.

**D4 — the template.** Fixed (`current_version = ""` → `"unknown"`, the `stable`/`rolling` clause removed, quoting noted), plus a new `b6_template_drift` rule that reads the **real template file** and checks its format-ruled fields against the existing consts. Running `check-sources` against it was rejected as the wrong instrument — the template has no `plugin.json` and no adjacent `sources.md`, so A-group and C-group checks would fail on placeholder values a skeleton is supposed to carry.

## The contract change, made by the test author

Two frozen cases had to **flip** from pass to fail (`current_version = 2025` and `= 1.0`, both unquoted). That is a deliberate contract change, not a test weakened to fit an implementation, so it went to the test author rather than the implementer. Their labels called them "real corpus value" — misleading, since the corpus value is the *quoted* `"2025"` — and were rewritten to state what they now assert.

The won't-fix decisions are pinned as reject-cases too, so a future widening is a deliberate act rather than a drift.

Sources self-test 63 → **74 cases** (70 after the first pass, 74 after the review fix below).

## Verified by making things fail

A green run proves nothing about a new rule. Two probes, both reverted clean:

- `b6_template_drift` — restored `current_version = ""` in the template → `templates/sources.toml: sources.0.current_version = '' fails its own format rule`
- the type rule — set a real entry to an unquoted `1.10` → `core/apple-container: current_version is a float, expected a QUOTED version string ... (1.10 becomes 1.1)`

## A borrowed figure, corrected

The implementer's code comment asserted "0 of 31 adversarial junk strings newly accepted" — citing the planner's set, which it told me plainly it did not have and had not run. It had measured its own delta instead. The substance held; the attribution did not.

My own 18-probe set found the claim is **set-dependent**: the widening also newly admits `1.0.0--` and `a-1-2-3`. Neither is a plausible authoring mistake, and `1.0.0--` is **already accepted by `SEMVER_RE`** — so tolerating it is exactly the agreement D1b set out to achieve. The decision stands; the comment now states what was measured rather than what was inherited.

## Gates

`mise test` exit 0; corpus 28 validated / 0 pending; sources self-test 74 cases; all 8 skills-quality suites green; waivers **49 — unchanged**; `test/quality-baseline.json` diff empty; `mise run test:version-bumps origin/main` exit 0 with `claude-code` 0.2.42 → 0.2.43; SKILL.md 217 lines; gitleaks clean.

## Gate 3 — REQUEST CHANGES, both blockers fixed

Full record: the review comment above. The reviewer completed its work but died to an API error twice before it could post, so I posted its findings on its behalf. Both blockers were real and both were mine to answer.

**The widening admitted bare calendar dates.** `2024-01-01`, `2026-07-30`, `10-11-2025`, `1-800-flowers`, `v1-2-3`, `1--2` — all rejected on `main`, all passing on the first version of this PR. `SEMVER_RE` rejects them too, so "we're only agreeing with `SEMVER_RE`" covered 7 of the 18-string delta and **not the other 11**. A date here is a plausible mistake, not a constructed one: `last_checked` sits two lines away in every entry and takes exactly that format.

**And the chain that produced it is the point.** The planner measured a 31-item junk set. The implementer cited "0 of 31" without having that set. I caught the borrowed attribution and replaced it with my own claim — from an 18-probe set that contained no dates. Gate 3's 68-probe set found the family none of us had looked for. Three successive coverage claims, each more confident than its evidence, and the last one was mine while I was fixing the first.

The fix is principled rather than patched: a hyphenated suffix now requires a **dotted base**, because a hyphenated prerelease is semver-shaped and semver requires dotted numeric components. Measured against three candidates — leaving it leaks 6; requiring a letter in the suffix still leaks `1-800-flowers`; requiring a dotted base leaks 0 and rejects 0 of 39 corpus values.

Four reject-cases now pin the family, each justified. `1-800-flowers` earns its place specifically by ruling out the letter-in-suffix alternative.

**`b6_template_drift` crashed the run and masked every other finding.** A missing, unparseable, or malformed template died with a raw Nushell error. The reviewer proved the masking: with a genuine `whamm` finding present *and* the template moved aside, the validator printed its banner and died — the real finding never reported. Now guarded; the combined probe shows a corrupted entry **and** a missing template reported together, exit 1 preserved.

**Also corrected:** the PR body said `scalar-str` has 8 call sites — 8 on `main`, **7 on head**, since this PR removed one. And the implementer found a *third* stale reproduction of the pre-196 regex in a self-test section comment, wrong since `5649061`.

Sources self-test 70 → **74 cases**.

## Gate 3 round 2 — the guard was honest but incomplete

The date fix verified clean on an **independent 113-probe set** (deliberately not the prior reviewer's 68 or my 18, since two people had already under-measured this exact regex): 18 strings now rejected, **zero regressions**, and the "agrees with `SEMVER_RE` on suffix shape" framing now holds across the whole delta — which is precisely what the first blocker said it did not.

But `b6_template_drift` still masked. The first guard covered the four **container** shapes — null, no `[meta]`, no `[[sources]]`, empty `sources` — and its comment enumerated exactly those and claimed nothing more. It did not cover a template that parses into a well-formed record whose first entry is **missing a checked field or holds a wrong type**; the `else` branch indexed those unguarded.

Five shapes crashed and masked, each verified with a real finding planted so the masking was directly observable:

```
[[sources]] missing current_version   -> column_not_found
[[sources]] missing update_priority   -> column_not_found
[meta] missing last_full_check        -> column_not_found
sources = "oops"                      -> only_supports_this_input_type
current_version = 2025 (unquoted)     -> operator_unsupported_type
```

The last one is D3 of this very PR: a real entry gets a clean diagnostic for an unquoted numeric, while the **template** crashed on identical input. A drift-prevention rule that could not survive the drift it exists to catch.

Now guarded one level deeper, and re-verified on all five — both the real finding and a `b6_template_drift` finding appear together, no crash, exit 1 preserved. I reproduced one shape independently.

**Worth distinguishing:** this was an *incomplete* guard, not an *overstated* one. Its comment described its actual coverage honestly. After the borrowed-figure chain earlier in this PR, that difference is the whole point — the failure was reach, not truthfulness.

Self-test holds at **74 cases**; corpus 28/0; waivers 49; baseline diff empty.


vinnie357:
## Gate 3 — REQUEST CHANGES

Posted by the lead on the reviewer's behalf: the reviewing agent completed its work but died to an API error twice before it could post, once at 13:36 and again at 13:42. Its findings reached me in full via status message and are reproduced here with my independent verification. Credit for both blocking findings is the reviewer's.

### BLOCKING 1 — the widened suffix admits bare calendar dates, and two in-repo claims deny it

The reviewer built a **68-probe** adversarial set against my 18 and found an **18-string delta**. The PR body enumerates 4 of them. The unenumerated family is dates:

```
2024-01-01   2026-07-30   10-11-2025   2024-1-1
1-800-flowers   v1-2-3   1--2
```

Independently reproduced — all reject on `main`, all pass on head:

| value | old regex | new regex | SEMVER_RE |
|---|---|---|---|
| `2024-01-01` | reject | **accept** | reject |
| `2026-07-30` | reject | **accept** | reject |
| `10-11-2025` | reject | **accept** | reject |
| `1-800-flowers` | reject | **accept** | reject |
| `1.0.0-alpha-1` | reject | accept | accept |

Only 7 of the 18 are accepted by `SEMVER_RE`, so the "we are merely agreeing with `SEMVER_RE`" justification covers 7 and **not the other 11**.

A bare ISO date in `current_version` is a plausible authoring mistake, not a constructed one: `last_checked` sits two lines away in every entry and takes exactly that format, and the self-test already treats a date in this field as an error worth catching.

**Two durable in-repo claims are falsified by this**, which is this epic's own defect class:

- `test/validate-sources.nu`: "No plausible authoring mistake became acceptable."
- `skill-update/SKILL.md`: the widening is "self-consistent rather than looser".

The first sentence is mine. I wrote it while correcting the implementer's borrowed "0 of 31" figure — replacing a claim it had not measured with one I had under-measured, from an 18-probe set that never contained a date. Catching the provenance error and committing the substance error in the same edit.

**Fix, measured against three candidates.** Permit a hyphen-containing suffix only when the base has at least one dot-group:

```
^[A-Za-z]{0,10}[-.]?\d+((\.\d+){1,4}([-+][0-9A-Za-z.-]+)?|([-+][0-9A-Za-z.]+)?)$
```

| candidate | must-accept broken | date family leaking |
|---|---|---|
| head (no fix) | 0 | **6** |
| suffix must contain a letter | 0 | 1 (`1-800-flowers`) |
| **base needs a dot-group** | **0** | **0** |

Verified: 0 of 39 corpus values rejected; `1.0.0-alpha-1` and `1.0.0-rc-2` still accepted. The rule is principled — a hyphenated prerelease is a semver-shaped thing and semver requires dotted numeric components, so a bare-integer base has no business carrying one.

### BLOCKING 2 — `b6_template_drift` aborts the run and masks every other finding

All six checked fields fire correctly. But a missing, unparseable, `[[sources]]`-less, or `[meta]`-less template crashes with a raw Nushell error. The reviewer proved the masking: with a genuine `whamm` finding present **and** the template moved aside, the validator prints its banner and dies — **the real finding is never reported**.

It fails closed (exit 1) and cwd-independence passes (tested from four directories), so the blast radius is bounded. Fix is roughly three lines: guard existence and parse, and report a finding instead of crashing.

### Non-blocking correction to the PR body

"`scalar-str` … has 8 call sites" — 8 on `main`, **7 on head**; this PR removed one.

- **D1a re-derived independently**: the reviewer's own 31-item junk set gives **31/31** newly accepted under the plan's PEP 440 variant, against the plan's 6/31. The rejection is *stronger* than the PR claimed, not weaker. Demand claim holds — no `pypi` in `CHECK_METHODS`, 0 Python upstreams by url/name scan.
- **Type check, both directions**: int, float, bool, datetime, list, inline table, `0x10`, `1_000`, `1e3` all rejected; `"1.10"`, `"unknown"`, `"1.0.0-alpha-1"` accepted; `stable`/`rolling`/`""` rejected; `none` entries unaffected; `b4_conditional` undisturbed.
- **TDD integrity**: implementation commit changed zero `want:` / `label:` / `toml:` lines; the suite at `80ad746` is red on exactly the four expected cases.
- **Doc/check**: the regex in SKILL.md is byte-identical to the code const.
- **Integrity**: waivers 49 → 49, baseline byte-identical to main, corpus 28/0, self-test 63 → 70, SKILL.md 217 lines, `0.2.43` == `0.2.43`, trailing bytes match.
- **Body figures**: 117/119, 8/12, 0/193 all confirmed; "0 of 39" correct as distinct-non-unknown.

vinnie357:
## Gate 3 re-review (second reviewer) — REQUEST CHANGES

Narrow re-review of the two fixes at `8c79886`, against the prior Gate 3 record in [#issuecomment-5151684337](https://github.com/vinnie357/claude-skills/pull/188#issuecomment-5151684337). Everything was verified by execution in a throwaway clone; the shared working tree was never checked out.

**Blocking 1 (date family) is fully fixed.** **Blocking 2 (`b6_template_drift` masking) is fixed for the four shapes it names and still masks for a fifth.** That fifth is the one finding below.

---

### BLOCKING — `b6_template_drift` still crashes and still masks, for a *parseable but structurally incomplete* template

The guard added in `8c79886` covers `null` / no `meta` / no `sources` / empty `sources`. It does **not** cover a template that parses into a well-formed record whose first `[[sources]]` entry is missing a checked field, or whose values are the wrong type. The `else` branch then indexes `$tmpl_src.current_version` (line 610) unguarded.

I ran every probe with a **genuine `whamm` finding planted** (`current_version = "2024-01-01"`), so masking is directly observable. `MASKS = true` means the real finding never printed:

| template shape | exit | crash | `b6` emitted | whamm finding reported | MASKS |
|---|---|---|---|---|---|
| `[[sources]]` missing `current_version` | 1 | `column_not_found` | no | no | **yes** |
| `[[sources]]` missing `update_priority` | 1 | `column_not_found` | no | no | **yes** |
| `[meta]` missing `last_full_check` | 1 | `column_not_found` | no | no | **yes** |
| top-level `sources = "oops"` | 1 | `only_supports_this_input_type` | no | no | **yes** |
| `current_version = 2025` (unquoted) | 1 | `operator_unsupported_type` | no | no | **yes** |

Verbatim, first row:

```
🔍 Validating sources.toml across 28 plugins...

Error: nu::shell::column_not_found

  x Cannot find column 'current_version'
     ,-[test/validate-sources.nu:610:42]
 610 |             ["sources.0.current_version" $tmpl_src.current_version (...)]
```

Banner, then death — the same signature as the prior blocker's reproduction. (In three of the five rows my first-pass harness scored `b6 = true`; that was a false positive from the string `b6_template_drift` appearing in the *error trace's source context* at line 601. Re-checked: no b6 finding is emitted in any of the five.)

Two reasons this is blocking rather than a nit:

1. **A dropped or unquoted field is the single most likely form of template drift** — it is precisely what the rule was added to detect. The rule crashes on the case it exists for.
2. **`current_version = 2025` unquoted in the template crashes**, and D3 of this very PR is the rule that makes an unquoted `current_version` a reportable finding in real entries. The template gets a crash where an entry gets a diagnostic.

The in-repo claim is honest and correctly scoped — the comment enumerates exactly the four shapes it handles, and does not overreach. But `— fail closed, but report` reads as a general property of the rule, and it is not one.

**Fix** is the same shape as the one already applied — wrap the check construction rather than only the `open`:

```nu
} else {
    let checks = (try {
        let tmpl_src = ($tmpl.sources | first)
        [ ...the six rows... ]
    } catch { null })
    if $checks == null {
        $findings = ($findings | append {
            rule: "b6_template_drift"
            severity: "fail"
            message: $"templates/sources.toml: [[sources]]/[meta] missing a checked field or holding a non-string value at ($tmpl_path)"
        })
    } else {
        for row in $checks { ... }
    }
}
```

Exit 1 is preserved in all five shapes, so the blast radius stays bounded — but "fails closed" was never the complaint; masking was.

---

### Confirmed — Blocking 1 is genuinely fixed

I built an **independent 113-probe set** (not the prior 68, not the 18) and ran all of it through four regexes: `main`, the buggy `5649061`, `head`, and `SEMVER_RE`.

**The whole date/junk family is closed.** 18 strings that `5649061` accepted are now rejected:

```
2024-01-01  2026-07-30  2026-08-01  10-11-2025  2024-1-1  01-01-2024
99-99-9999  v2024-01-01  1-800-flowers  1-800-FLOWERS  1--2  a-1-2-3
v1-2-3  1-a-2  0-0-0  1-alpha-1  3-beta-1  1+build-1
```

**Zero regressions.** No probe accepted on `main` is rejected on head — the `{0,4}` → `{1,4}` split did not cost the bare-major/no-dot forms (`3`, `10`, `v1`, `0`, `1-alpha`, `1.21-bookworm` all still accepted).

**Newly accepted vs `main`: 11**, of which **9 are also accepted by `SEMVER_RE`**:

| newly accepted | `SEMVER_RE` |
|---|---|
| `1.0.0-alpha-1`, `1.0.0-rc-2`, `1.0.0-alpha-beta-gamma` | accept (the intent) |
| `1.0.0--`, `1.0.0+a-b`, `1.0.0-01-02`, `2.5.0-2024-01-01`, `2.7.18-final-0`, `5.15.0-91-generic` | accept |
| `1.0+build-1`, `2.5-2024-01-01` | **reject** |

The two `SEMVER_RE` rejects fail it *only* on component count (two groups, not three) — the exact dimension `VERSION_SHAPE_RE` is documented to widen (one to five groups). So the "agreeing with `SEMVER_RE`" framing now holds on suffix shape for the full delta, which is what the prior blocker said it did not.

Attacked in both directions and held: constraint strings (`1.x`, `1.*`, `>=1.0.0`, `^1.0.0`, `~1.2`), prose (`banana`, `stable`, `rolling`, `latest`, `main`, `HEAD`, `see changelog`, empty), URLs and paths, all four whitespace forms, embedded/leading/trailing newlines (Rust's `$` does not admit a trailing `\n` — `"1.0.0\n"` rejects), the 10/11-char prefix boundary, dot-group count boundary (`1.2.3.4.5` accept / `1.0.0.0.0.0` reject), degenerate suffixes (`1.0.0-`, `1.0.0+`, `1.0.0++`), D1c duals (`1.0.0-a+b`, `1.0.0-alpha-1+build-2` both reject), underscores, and `0x10` / `1_000` / `1e3` as strings.

### Confirmed — corpus untouched (item 2)

Measured across all 28 `sources.toml`: **206 entries, 193 with `current_version`, every one a TOML string**, 152 `unknown`, **39 distinct non-unknown values**.

- **0 of 39 rejected** by the corrected regex.
- **0 of 39 changed acceptance** between `main` and head.
- `1.0.0-alpha-1` and `1.0.0-rc-2` both still accepted — the D1b fix survived the correction.

### Confirmed — the four new cases are load-bearing (item 4)

Self-test at `6320505` (new cases, pre-fix regex) fails on **exactly the four**, and only those:

```
❌ current_version = 2024-01-01 ...: want [b3_version_shape], got []
❌ current_version = 2026-07-30 ...: want [b3_version_shape], got []
❌ current_version = 10-11-2025 ...: want [b3_version_shape], got []
❌ current_version = 1-800-flowers ...: want [b3_version_shape], got []
```

Green at head: `✅ sources self-test passed (74 cases)`. Red-then-green on the real defect, not cases that were already passing.

**TDD boundary intact**: `git show 8c79886 -- test/validate-sources.nu | grep -E '^[+-]\s*(want:|label:|toml:)'` is empty — the implementation commit changed no case data.

### Confirmed — doc/code agreement (item 5)

Three copies of the regex exist; all **byte-identical** (compared programmatically, not by eye):

- `test/validate-sources.nu:67` (the const)
- `test/validate-sources.nu:1741` (self-test section comment — the third stale copy the implementer reported fixing)
- `SKILL.md:45`

`grep -rn '{0,4}'` over `*.nu` / `*.md`: **no stale pre-196 regex remains anywhere.**

All three falsified claims are gone from head: "No plausible authoring mistake became acceptable", "self-consistent rather than looser", "already accepted this shape". The prose now states the dotted-base rule and names the rejected family.

### Confirmed — every PR body figure

| claim | measured | |
|---|---|---|
| 39 corpus values | 39 distinct non-unknown | ✓ |
| 0 rejected | 0 | ✓ |
| 74 self-test cases | `✅ sources self-test passed (74 cases)` | ✓ |
| 49 waivers | `allowed_failures` length 49 | ✓ |
| baseline diff empty | `git diff origin/main -- test/quality-baseline.json` empty | ✓ |
| corpus 28 / 0 pending | `28 file(s) validated, 0 pending` | ✓ |
| `scalar-str` 7 on head | `(scalar-str ` → 7 head, 8 main | ✓ |
| `claude-code` 0.2.43 | plugin.json and marketplace.json both 0.2.43; main 0.2.42 | ✓ |
| SKILL.md 217 lines | 217 | ✓ |

No stale figure remains.

---

### Non-blocking

**1. `2.5-2024-01-01` is newly accepted and is the one string in my set that `SEMVER_RE` rejects on suffix-adjacent grounds.** A dotted base licenses a hyphenated suffix, so `<version>-<date>` passes. Not a bare date, not confusable with `last_checked`, and low plausibility as an authoring slip — recording it so the delta is fully enumerated rather than proposing a change.

**2. Date-shaped strings that pass on head but also passed on `main` — untouched by this PR, listed so "rejecting bare calendar dates" isn't read wider than it is:**

`2024-01`, `2024-1` (bare year-month), `2024.01-01`, `2024-01.01`, `2026-07.30` (mixed separators), `2026.08.01` (dotted date — indistinguishable from CalVer by construction; correctly unfixable), plus `555-1234`, `1-2`, `-1.0.0`, `.1.0.0`, `.1`.

`2024-01` is the one I'd call mildly awkward against the new SKILL.md sentence "rejecting bare calendar dates … like `2024-01-01`". The `like` keeps it exemplary and the shape was always accepted, so this is a wording note, not a defect.

**3. Rust's `\d` is Unicode-aware**, so `١.٢.٣`, `1.٢.3`, and `１.０.０` all pass. Accepted on `main` too — pre-existing, no action implied.

---

### Verdict

**REQUEST CHANGES** on the single `b6_template_drift` residual. The date-family fix is correct, principled, measured, and cost nothing in corpus acceptance; the tests genuinely pin it; the docs and every PR-body number now match the code. Once the `else` branch is guarded the same way the `open` was, this is an approve from me.

vinnie357:
## Gate 3 round 3 (second reviewer) — APPROVE

Re-review of `2ce4781` against my round-2 blocker in [#issuecomment-5151788776](https://github.com/vinnie357/claude-skills/pull/188#issuecomment-5151788776). Verified by execution in a throwaway clone; the shared working tree was never checked out.

**The blocker is fixed.** The `else` branch is now guarded the same way the `open` was, and I could not find a sixth shape.

---

### 1. All shapes degrade — 27 template corruptions, every one clean

Every run had a **genuine `whamm` finding planted** (`current_version = "2024-01-01"`), template mutated one shape at a time and restored between each, so masking is directly observable. `MASKS` is derived from whether the planted finding actually printed.

| group | shapes | exit | crash | `b6` emitted | planted finding reported | MASKS |
|---|---|---|---|---|---|---|
| **guard 1** — container | missing, unparseable, empty file, no `[meta]`, no `[[sources]]`, `sources = []` | 1 | none | yes | yes | **no** |
| **guard 2** — field-level (my round-2 blocker) | missing `current_version`, missing `update_priority`, missing `last_full_check`, `sources = "oops"`, `current_version = 2025` | 1 | none | yes | yes | **no** |
| **sixth-shape hunt** | `sources` as int / bool / TOML date / float, `sources = ["a","b"]`, `meta` as string / int, `current_version` as inline table / array, `last_checked` as unquoted TOML date, `version_constraint = 3`, `reviewed_at_plugin_version = 1.0`, template path is a **directory** | 1 | none | yes | yes | **no** |
| **exotic file** | dangling symlink, invalid UTF-8 bytes, BOM-only file | 1 | none | yes | yes | **no** |

**27 of 27: exit 1, zero `Error: nu::` traces, a real `b6_template_drift` finding, and the planted finding still reported.** The two container shapes that were already passing in round 2 still pass — the deeper nesting did not break them.

Verbatim, on the shape that was the round-2 reproduction:

```
│ 0 │ b3_version_shape  │ whamm/whamm: current_version '2024-01-01' is neither │
│   │                   │  a recognizable version string nor 'unknown'         │
│ 1 │ b6_template_drift │ templates/sources.toml: a checked field is missing   │
│   │                   │ or the wrong type (Cannot find column                │
│   │                   │ 'current_version')                                   │
```

Both findings, run intact, and the underlying Nushell error is carried into the message as diagnostics rather than discarded — better than the minimum I asked for.

(Round-2 methodology note, since it matters for reading this table: my first-pass detector scored `b6 = true` on three shapes where no finding was emitted, because the string `b6_template_drift` appears in the *error trace's source context*. This round's detector excludes trace lines, and `crash = none` everywhere independently confirms there are no traces to be fooled by.)

**cwd-independence holds** — exit 0 from the repo root, `test/`, `plugins/tools/whamm/`, and `plugins/core/skills/`.

### 2. Nothing else regressed

| check | measured | |
|---|---|---|
| self-test | `✅ sources self-test passed (74 cases)` | ✓ |
| corpus | `28 file(s) validated, 0 pending` | ✓ |
| corpus values | 206 entries, 193 with `current_version`, 152 `unknown`, **39 distinct non-unknown, 0 rejected** | ✓ |
| waivers | `allowed_failures` length **49** | ✓ |
| baseline | `git diff origin/main -- test/quality-baseline.json` empty | ✓ |
| `VERSION_SHAPE_RE` | **not present in the `2ce4781` diff**; const at `:67` still the corrected value | ✓ |

**The date-family fix is intact.** I re-ran my full round-2 113-probe set against `2ce4781`: byte-identical results — 18 junk/date strings still closed, **zero regressions vs `main`**, 11 newly accepted of which 9 are also accepted by `SEMVER_RE`, and the same two (`1.0+build-1`, `2.5-2024-01-01`) failing it only on component count.

### 3. The new comment is accurate — checked against the code, not taken on faith

Every claim in it verified:

- **"caught by `.sources | first` failing, not by `is-empty`"** — the load-bearing detail, and correct. Executed directly:

  | value | `is-empty` | `first` |
  |---|---|---|
  | `"oops"` | `false` | error |
  | `5` | `false` | error |
  | `true` | `false` | error |
  | `1.5` | `false` | error |

  So a non-list `sources` genuinely passes the outer guard and is genuinely caught by the inner one. The comment explains the right mechanism.

- **The three named error types are the three that actually arise**, at the shapes named: `column_not_found` (missing field), `only_supports_this_input_type` (`sources = "oops"`), `operator_unsupported_type` (unquoted numeric `current_version`). All three observed in round 2 at exactly those shapes.
- **"all six checked fields"** — exactly six rows sit inside the `try` (`meta.reviewed_at_plugin_version`, `meta.last_full_check`, `sources.0.current_version`, `sources.0.last_checked`, `sources.0.version_constraint`, `sources.0.update_priority`).
- **"from INSIDE the check construction, not from `open`"** and **"one level in"** — both structurally accurate.

**No overreach.** The comment describes what it catches and by what mechanism; it does not claim exhaustiveness, and every specific it asserts is true. Given this PR's history I looked specifically for a sentence that reads wider than the code, and there isn't one — the round-2 phrasing I flagged (`— fail closed, but report`, which read as a general property) is now bounded by the enumeration that follows it.

### 4. No sixth shape

I attacked the two remaining unguarded surfaces specifically:

- **The outer guard's own expressions.** `"meta" in $tmpl`, `"sources" in $tmpl`, and `$tmpl.sources | is-empty` are evaluated *outside* any `try`, so a value that makes `is-empty` itself raise would crash before either guard. Tested `sources` as int, bool, float, TOML date, and list-of-strings, and `meta` as string and int — `is-empty` returns `false` for all non-list scalars rather than raising, so the outer guard falls through to the inner `try`, which catches. No crash.
- **The `open` boundary.** Directory in place of the file, dangling symlink, invalid UTF-8, BOM-only — all caught by the existing `open` try.
- **Post-construction interpolation.** `$"($value)"` in the finding message can only be reached by a value on which `== "unknown"` and `=~` already succeeded, so exotic types are filtered out before they get there.

Both guards together now cover every corruption I could construct.

---

### Verdict

**APPROVE.** The round-2 blocker is resolved with the fix shape I suggested, the two previously-passing container shapes are undamaged, no regression anywhere, the date-family work from round 2 is intact, and the comment is accurate.

Prior non-blocking notes stand unchanged and need no action: `2.5-2024-01-01` newly accepted; the date-shaped strings that pass on head but also passed on `main` (`2024-01`, `2024.01-01`, `2026.08.01`, …); and Unicode-aware `\d`.

PR #182, opened 2026-07-31
docs(skills): describe the sources.toml coverage check and correct check_count to 17

vinnie357:
Closes claude-skills-190 and claude-skills-191, both filed out of the Gate 3 review on #180. Same species — documentation disagreeing with the checker it describes — so they batch into one PR and one CI cycle.

## 190 — two docs still described a retired check

PR #180 removed the substring `source` check: it asked whether a skill's *name* appeared anywhere in its plugin's `sources.md`, a whole-file test that could not tell real coverage from an incidental name match. Coverage is now recorded in `sources.toml` and verified by set difference against the plugin's declared skill set.

Two files inside the `claude-code` plugin still advertised the old behaviour:

- `commands/benchmark-skills.md:45` — "**Source documentation**: Skill appears in plugin's `sources.md`" → "**Source coverage**: Skill is listed in one of the plugin's `sources.toml` entries' `skills` array"
- `skills/claude-skills-benchmark/SKILL.md:23` — "...and sources documented in the plugin's `sources.md`" → "...and coverage recorded in the plugin's `skills/sources.toml` (each entry's `skills` array), verified by set difference against the plugin's declared skill set in `test/validate-sources.nu`"

They were split out of #180 rather than folded in because they live inside a plugin and therefore need a version bump, which `test/README.md` did not.

## 191 — `check_count` was off by one, and the fix is not the one the bee guessed

The bee framed this as an open question: either count `fm_schema` and make `check_count` 17, or explain why it is excluded and rename the constant. Measuring settles it — there is nothing to exclude.

The per-skill loop can append exactly **17** distinct failure keys:

`allowed_tools, anti_fab, desc, examples, fm_schema, invocations, kebab, lines, links, name_dir, name_len, orphans, ref_depth, reserved, third_person, use_when, version_pin`

`test/README.md`'s per-skill table lists exactly those same 17. The sets are **identical** — `fm_schema` was never deliberately un-numbered, it was simply uncounted. So `check_count = 16` was wrong, and a skill failing every check would have scored `-1/16`.

Applied: `check_count` 16 → 17, and the README's heading and prose to match. The heading now agrees with the constant **and** with its own table's row count for the first time — previously the table has always carried one more row than the heading claimed (18 rows under 17 at main before #180; 17 under 16 after it).

## Blast radius checked before changing it

[...]

PR #189, opened 2026-08-01
docs(skills): defer the benchmark score to the validator and fix the tree-sitter queries link

vinnie357:
Closes claude-skills-193 and 194 — two independent one-liners in different plugins, batched into one PR and one CI cycle.

## 193 — a denominator that matched nothing, removed rather than corrected

`commands/benchmark-skills.md` instructed "Calculate score (X/11 checks passing)". Three numbers, no two agreeing: the command's own bullet list has **6** grouped bullets covering roughly 14 named sub-checks, and the validator scores each skill out of `check_count`, currently **17**.

The bee offered either fixing the number or dropping it. **Dropped**, for a reason stronger than the one I gave the implementer: `claude-skills-benchmark/SKILL.md:21` already states that `test/validate-skills-quality.nu` "is the single source of truth — do not restate its checks here." The command was violating a principle its own sibling skill declares, not merely carrying a stale figure.

Writing `17` would fix today's mismatch and schedule tomorrow's. That number was already wrong *before* `check_count` moved 16 → 17 in #182, which is evidence it drifts faster than anyone maintains it. Step 5 now says to run `mise run test:skills-quality` and report the score exactly as printed; the example scorecard shows `X/N` with a note that `N` is the validator's `check_count`, not a fixed number.

The bullet list stays as a category summary the agent narrates. It no longer contradicts a count, because there is no second count left to disagree with.

## 194 — a stale link in a file that only recently became reachable

`topiary/references/query-syntax.md:372` pointed at `tree-sitter.github.io/tree-sitter/using-parsers#query-syntax`. That **301s to `using-parsers/` and drops the `#query-syntax` fragment**, so a reader landed on a page that does not answer the question.

This mattered more than when the bee was filed: **#183 made this file reachable from `topiary/SKILL.md` for the first time**, putting it on the progressive-disclosure path.

Now points at `tree-sitter.github.io/tree-sitter/using-parsers/queries/1-syntax.html` — verified 200, no redirects, `<title>Basic Syntax</title>`, and the S-expression material the citing file is about.

**That target changed after review, and the reviewer was right.** My first fix pointed at the queries *section index* on a stability argument: a numbered filename could be renumbered, and the link text read "Documentation". Gate 3 fetched both and the trade was worse than I allowed — the index's entire `<main>` is a one-paragraph stub with **zero** S-expression content, while `1-syntax.html` opens directly on `(binary_expression (number_literal) (number_literal))`, fields, and negated fields. The content gap was actual; the renumbering risk was hypothetical. My own earlier probe had shown 0 keyword hits against 13 and I read it as acceptable rather than disqualifying.

Worth recording either way: **no CI check would catch either link rotting.** Check 14 (`links`) validates local asset paths, not external HTTP — which is both the reason to prefer a stable URL and the reason the original rotted unnoticed.

**Sibling links checked in the same pass**, per the bee's own AC that two earlier sweeps in this epic were incomplete: `topiary.tweag.io/book/reference/capture-names/index.html` → 200, `tweag.io/blog/2025-01-30-topiary-tutorial-part-1/` → 200. Neither needs changing. Notably `topiary.tweag.io` still resolves despite the `tweag/topiary` → `topiary/topiary` repo rename this migration caught earlier.

## Gates

`mise test` exit 0; `mise run test:version-bumps origin/main` exit 0 with `claude-code` 0.2.43 → 0.2.44 and `tweag` 0.1.3 → 0.1.4; corpus 28 validated / 0 pending; sources self-test 74 cases; all 8 skills-quality suites green; waivers **49 — unchanged**; `test/quality-baseline.json` diff empty; gitleaks clean.

## Gate 3 — APPROVE, two findings applied and three filed

**Applied here:** the link target above, and the score wording. "Report the score exactly as it prints (`X/N`)" read as if the validator emits the literal string `X/N`; it prints concrete values like `17/17`. Now says `X/N` is a placeholder, names a real example, and notes the validator takes no `--plugin` filter — so the full suite runs and you read the rows you need.

**Filed as claude-skills-199,** because the review found this fix is half-applied and widening the PR would repeat the mistake of fixing what is adjacent rather than what is scoped:

- The `500` denominator in the example scorecard survives the exact argument that removed `11` — it is validator-owned too (`:2221` gates on it, `:2360` hardcodes the display string).
- The surviving step-3 bullets carry hard numbers (`≤1024`, `≤64`, `≤500`) where the sibling skill deliberately states intent without figures. That withholding is what makes "do not restate its checks here" survivable, and the command breaks it one screen above the line this PR fixed.
- **The bullets and the score now measure different sets** — newly load-bearing, because making the printed score authoritative removed the coincidence that previously hid it. Step 3 names two things the validator does not check and omits seven it does, and one bullet belongs to a different validator entirely.

The reviewer verified what held: `mise run test:skills-quality` exists and prints the promised form, `claude-skills-benchmark/SKILL.md:21` says what the body claims, both sibling links return 200, and every version and figure checks out.


vinnie357:

**Verdict: APPROVE.** No blocking findings. Five non-blocking items below, then the list of what I attacked that held — the negative is only worth what the attack was.

I went after the two things most likely to be wrong: that step 5 tells an agent to report a form the tool never prints, and that the body's own "single source of truth" argument cuts against the bullet list it spared. The first does not land. The second partly does.

---

### Non-blocking findings

**1. [suggestion] The `500` denominator survives the exact argument that killed `11`.**

The example row still reads `120/500` / `380/500`. `500` is validator-owned in the same way `check_count` is — `test/validate-skills-quality.nu:2221` gates on `> 500` and `:2360` hardcodes the display string `$"($line_count)/500"`. The new note claims the change means the score "will not go stale as the check list grows"; true for the Score column, but the same two-column-over table still embeds a validator-owned threshold. If the drift argument is strong enough to remove a denominator, it applies here too. Currently correct, so nothing is broken — but the fix is internally inconsistent about its own principle.

**2. [suggestion] The fix is half-applied: the surviving bullets restate more than the sibling skill's pattern allows.**

`claude-skills-benchmark/SKILL.md:23` states check *intent* while deliberately withholding numbers — "length caps", "bounded body size". That is the pattern that makes SKILL.md:21's "do not restate its checks here" survivable: intent without figures. The command's step-3 bullets instead carry hard numbers: `≤1024 chars`, `≤64 chars`, `≤500 lines`. All three match the validator today (verified at `:2194`, `:2215`, `:2221`), so this is not a live defect — but it is precisely the drift class this PR just removed, one screen above the line it removed it from. The body's framing ("the command was violating a principle its own sibling skill declares") is therefore still partly true after the fix. Worth a follow-up bee; out of scope to widen this PR.

**3. [question] The bullet list and the Score column now measure different sets.**

Making the score authoritative newly exposes a mismatch that was previously hidden by the agent hand-computing both from the same bullets. The validator's 17 keys are `desc, use_when, third_person, kebab, name_len, reserved, lines, fm_schema, examples, ref_depth, anti_fab, allowed_tools, name_dir, links, orphans, invocations, version_pin`. Against that, step 3 names two things it does **not** check — `imperative language`, and `Has references/` (`ref_depth` tests nesting only; orphan detection is a separate key) — and omits seven it does: `fm_schema`, `allowed_tools`, `name_dir`, `links`, `orphans`, `invocations`, `version_pin`. Bullet 6 `Source coverage` is not in this validator at all; it lives in `test/validate-sources.nu`, as SKILL.md:23 itself says. So the agent narrates one check set and reports a score computed from another. Pre-existing in the bullets, newly load-bearing.

**4. [suggestion] I would have pointed at `1-syntax.html`, and the evidence is more lopsided than the body allows.**

The body concedes the index is not the syntax page and calls it precision traded for stability. Having fetched both, the trade is worse than that reads. The new target's entire `<main>` body is a stub:

> Pattern Matching with Queries — Code analysis often requires finding specific patterns in source code. Tree-sitter provides a simple pattern-matching language for this purpose, similar to what's used in its unit test system. This allows you to express and search for code structures without writing complex parsing logic.

That is the whole page (11,961 bytes, nearly all mdbook chrome). Zero S-expression material. `queries/1-syntax.html` is titled **"Query Syntax"** and opens on exactly what the citing file is about — `(binary_expression (number_literal) (number_literal))`, fields, negated fields — and returns **200 with 0 redirects**, so the stability objection is currently hypothetical while the content gap is actual. One extra sidebar click from a stub, in a reference file whose own title is `query-syntax.md`.

Not blocking: the link is live, topically correct, and the link text does read "Documentation" rather than "Syntax". And the renumbering concern is legitimate — worth noting that **no CI check would catch either link rotting**, since check 14 (`links`) validates local asset paths, not external HTTP. That cuts both ways: it is the reason to prefer the stable URL, and the reason the old one rotted unnoticed for so long.

**5. [nit] "exactly as `mise run test:skills-quality` prints it" is ambiguous.**

The validator prints `17/17`, never the literal string `X/N`. The first clause reads as if the tool emits `X/N`; the trailing clause rescues it. Also `--plugin=<name>` has no validator support (`def main [--update-baseline, --self-test]`), so that path requires running the full 102-skill suite and filtering rows — fine, but unstated.

---

### What I attacked that held

The primary suspicion — that removing the denominator left step 5 promising output the tool does not produce — **does not land.** I ran it.

- `mise run test:skills-quality` exists (`mise.toml:242`) and prints a per-skill table with `skill | plugin | lines | score | failed | new | details`. The `score` column emits `17/17`, `16/17`. The doc promises a form the tool genuinely produces, and step 5 is runnable as written.
- `check_count` really is 17: `let check_count = 17` at `:2354`, and I counted 17 distinct `$failed | append` keys in the scoring block rather than trusting the constant.
- **SKILL.md:21 says verbatim what the body quotes.** Not paraphrased.
- All four link claims are exact. Old URL: 1 redirect → `.../using-parsers/`, fragment dropped. New URL: 200, 0 redirects. **Both siblings verified myself**, not taken on report: `topiary.tweag.io/book/reference/capture-names/index.html` → 200/0 redirects; `tweag.io/blog/2025-01-30-topiary-tutorial-part-1/` → 200/0 redirects.
- The history checks out: `check_count` moved 16 → 17 in #182 (`2577581`), so `11` was indeed already wrong beforehand.
- "roughly 14 named sub-checks" is exactly 14 by count (4+3+3+2+1+1).
- **Scope is clean.** Exactly 5 files, +10/−8. The only JSON hunks are the four version lines — no adjacent key drift. `test/quality-baseline.json` diff is **0 lines**. Trailing bytes byte-identical to `origin/main` on all five files (compared as hex).
- `mise run test:version-bumps origin/main` exit 0, reporting exactly `claude-code: 0.2.43 → 0.2.44` and `tweag: 0.1.3 → 0.1.4`, each matching its marketplace entry.
- Waivers: my own run printed `49 debt/bug entries to burn down` — matches the claimed 49, independently reproduced.
- No skill exceeds the 500-line cap; highest observed are `wit` 496 and `interactive` 485. `topiary` sits at 375/500 and the one-line link edit does not move it.

Every figure in the PR body that I checked was accurate, including the ones easiest to get wrong. Findings 1–3 are about a principle applied to one line and not to the lines above it; none of them makes the command wrong today.

vinnie357:
## Gate 3 — re-review of `02dbc8b`, sign-off

**Verdict: APPROVE.** Findings 4 and 5 are correctly applied, nothing regressed, and no new inaccuracy was introduced. Verified independently rather than on report.

**1. Link.** `queries/1-syntax.html` → **200, 0 redirects.** Right target. You also changed the link *text* from "Tree-sitter Queries **Documentation**" to "Tree-sitter Query **Syntax**", which I did not ask for and which matters: "Documentation" was the one honest argument for pointing at the index, so leaving it would have left the label describing the old choice. Text and target now agree, and both agree with the file they sit in.

**2. Reworded note and step 5 — unambiguous, and accurate.** Checked each new assertion against the source rather than reading for tone:

- "`X/N` above is a placeholder" — no longer readable as tool output.
- "prints concrete values such as `17/17`" — `17/17` is a real printed value; most rows emit it.
- "`N` is the validator's `check_count`" — `test/validate-skills-quality.nu:2354`, and I counted 17 distinct `$failed | append` keys rather than trusting the constant.
- "The validator takes no `--plugin` filter" — `def main [--update-baseline, --self-test]` at `:2029`. Correct, and it closes the gap I raised.
- Step 5's "Read each skill's score ... rather than computing one" is stronger than the original "Calculate score by running", which still framed the agent as the computing party. That was the actual defect in the first wording, not just the `X/N` ambiguity.

One carried-over phrase, not new and not worth a change: "not a fixed number" — `check_count` **is** a fixed constant in the script; the sentence means "not fixed *in this doc*", which context makes clear.

**3. No regressions.** `mise test` **exit 0**. Waivers **49 — unchanged**. Corpus **28 validated / 0 pending**. Sources self-test 74 cases; disclosure lint 9 bad + 11 good. `test:version-bumps origin/main` exit 0 with `claude-code 0.2.43 → 0.2.44` and `tweag 0.1.3 → 0.1.4` intact — `02dbc8b` did not disturb them. `test/quality-baseline.json` diff still **0 lines**. Trailing bytes on both edited files still byte-identical to `origin/main`. Branch scope is 6 files; the sixth is `.bees/issues.jsonl` (+1 line) from `8fc2d16`, which touches nothing else.

Since bee 199's description exports to the tracked `issues.jsonl`, I specifically confirmed `test:disclosure` passes with it present — no paths, IPs, or hostnames of the kind that caused the #123 incident. It is clean.

**4. Bee 199 states findings 1–3 faithfully.** Every line reference in it is correct (`:2221`, `:2360`, `:2194`, `:2215`, `SKILL.md:23`), and it preserves the nuance that made finding 3 worth filing: the bullets/score divergence became load-bearing *only* once #189 made the printed score authoritative — previously both came from the same hand-count and could not disagree. The AC also asks the right open question rather than presuming the fix ("decide whether the score column example should show a literal number at all"), which is the correct handling for finding 1, since removing `120/500` and stating why `500` is safe are both defensible and that is a decision, not an edit.

Deferring 1–3 rather than chasing them here is the right call. Finding 2 in particular is a judgment about what the step-3 bullets are *for* — a reading aid or a check list — and answering it inside a link-fix PR would have been the same reflex that put the stale denominator there.

PR #190, opened 2026-08-01
fix(agent-sandboxing): use bare angle-bracket form for CLAUDE_PLUGIN_ROOT

vinnie357:
Closes claude-skills-178 and claude-skills-137.

## 178 — braced `CLAUDE_PLUGIN_ROOT` expanded at load time in four command bodies

Eight occurrences of the braced form sat in the BODY of four agent-sandboxing command files (`bootstrap.md`, `build-image.md`, `provision.md`, `status.md`). Command files load as skill content, so the braced form is substituted before the model sees it — replacing each with one machine's absolute cache path, pinned to whatever plugin version happens to be installed. Same defect class as claude-skills-169 and 172, hand-fixed for the third time here.

Fixed to the bare angle-bracket form, which is safe by construction: there is no dollar sign, so there is nothing to expand.

All eight sites verified to want `CLAUDE_PLUGIN_ROOT` rather than `CLAUDE_SKILL_DIR` — `scripts/` and `templates/` live at the plugin root, and `skills/sandbox/` contains only SKILL.md. All referenced artifacts confirmed to exist, so the second-order defect from claude-skills-128 (a path naming a `scripts/` dir that did not exist) is not repeated.

The braced form is deliberately left alone in two places where it is correct: YAML frontmatter, where the harness expands it at hook runtime, and reference files, which `Read` returns as raw bytes.

**Correction, per Gate 3:** an earlier draft of this body claimed the angle-bracket form is "the house style already used elsewhere in the claude-code plugin." That is false. `git grep -c '<CLAUDE_PLUGIN_ROOT>' -- plugins/tools/claude-code/` returns **zero** matches. That plugin uses `CLAUDE_PLUGIN_ROOT` **bare**, and documents the choice twice (`skills/claude-hooks/SKILL.md:37`, `skills/claude-plugins/SKILL.md:71`). Angle brackets ARE house style there — for `CLAUDE_SKILL_DIR`, a different variable. The code is safe either way, so the notation is not being churned; the justification was simply wrong. Settling `CLAUDE_PLUGIN_ROOT` notation repo-wide is filed as a follow-up.

## 137 — two follow-up nits from the PR #132 / #134 reviews

The `skills/sandbox/SKILL.md` bullet list wrote `templates/SandboxTemplate.*.yaml` and `mise.toml` skill-relative, though both are plugin-root artifacts — the same ambiguity PR #132 fixed one line above. Now consistent with the rest of the list.

`skills/webui/SKILL.md` asserted "All executions are preserved" and "History preservation: All executions are logged and accessible". Dagu can delete history, so the absolute was wrong.

**The first attempt at this fix replaced one unverified absolute with another** — it said retention is "configurable", but nothing in the repo evidences a retention setting. Grep across the whole dagu plugin returns only the lines this PR adds; `references/monitoring.md:179` is a bare bullet ("Archive old executions: Keep history manageable") with no mechanism. Both lines now claim only what the evidence supports: history is not unconditionally preserved, because `DELETE /dags/{dagName}/history/{requestId}` exists. Deletable, not configurable.

## Gates

Gate 1 `mise test` exit 0, 49 pre-existing debt entries, no new baseline key. `mise run test:version-bumps origin/main` exit 0 (agent-sandboxing 0.2.8→0.2.9, dagu 0.1.8→0.1.9). gitleaks clean via the native binary.

Gate 2 both remote checks pass; `mergeStateStatus` CLEAN after rebasing onto `e56823f`.

Gate 3 APPROVE — https://github.com/vinnie357/claude-skills/pull/190#issuecomment-5152798062. The reviewer's sharpest attack was that `<...>` inside a bash fence is a redirect pair, so `nu <CLAUDE_PLUGIN_ROOT>/scripts/x.nu` might input-redirect and create a file at an absolute path; tested, bash hits the failing input redirect first, exits 1, and creates nothing. It also confirmed `CLAUDE_PLUGIN_ROOT` is not in the Bash tool environment at all, so the old braced form was expanding to a bare `/templates/...` — this change is a net improvement, not a notation preference. Repo-wide completeness sweep found zero missed sites.

## Follow-ups filed rather than widened into this PR

- Settle `CLAUDE_PLUGIN_ROOT` notation repo-wide; bare vs angle is currently split across plugins.
- Add a lint catching braced `CLAUDE_*` in SKILL.md and `commands/*.md` bodies, allowed in frontmatter, `references/`, `templates/` and manifests. Three hand-fixes of one defect class with no check guarding it; the reviewer hand-classified that exact rule over 614 files with zero false positives.
- Upstream documents `outputStyles`, `lspServers` and `experimental` as plugin.json fields; the `claude-plugins` skill mentions none of them.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -166,7 +166,7 @@
       "source": "./plugins/tools/dagu",
       "strict": true,
       "description": "Dagu workflow orchestration: workflow authoring, web UI, and REST API",
-      "version": "0.1.8",
+      "version": "0.1.9",
       "category": "tools",
       "keywords": [
         "dagu",
@@ -256,7 +256,7 @@
       "source": "./plugins/tools/agent-sandboxing",
       "strict": true,
       "description": "Agent sandboxing for AI coding agents: microVM and kernel-isolated execution environments. Index skill (sandbox) routes to substrate-specific paths. Currently covers the kubernetes-sigs/agent-sandbox path on kind, GKE, and kina (Kubernetes-in-Apple-Containers), with Kata, gVisor, and Apple Container microVM substrates. Ships 7 skills + 6 slash commands + nushell scripts. Designed to expand to non-k8s sandboxing paths.",
-      "version": "0.2.8",
+      "version": "0.2.9",
       "category": "tools",
       "keywords": [
         "agent-sandbox",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.92",
+      "version": "0.4.93",
       "category": "meta",
       "keywords": [
         "skills",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.92",
+  "version": "0.4.93",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/agent-sandboxing/.claude-plugin/plugin.json
+++ b/plugins/tools/agent-sandboxing/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "agent-sandboxing",
-  "version": "0.2.8",
+  "version": "0.2.9",
   "description": "Agent sandboxing for AI coding agents: microVM and kernel-isolated execution environments. Index skill (sandbox) routes to substrate-specific paths. Currently covers the kubernetes-sigs/agent-sandbox path on kind, GKE, and kina (Kubernetes-in-Apple-Containers), with Kata, gVisor, and Apple Container microVM substrates. Ships 7 skills + 6 slash commands + nushell scripts. Designed to expand to non-k8s sandboxing paths.",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/agent-sandboxing/commands/bootstrap.md
+++ b/plugins/tools/agent-sandboxing/commands/bootstrap.md
@@ -87,7 +87,7 @@ Pick the template matching the substrate:
 Override the placeholder image:
 
 ```bash
-TEMPLATE=${CLAUDE_PLUGIN_ROOT}/templates/SandboxTemplate.kata.yaml
+TEMPLATE=<CLAUDE_PLUGIN_ROOT>/templates/SandboxTemplate.kata.yaml
 REGISTRY=${REGISTRY:-ghcr.io/$USER}
 TAG=${TAG:-latest}
 sed -e "s|REGISTRY/claude-code:TAG|${REGISTRY}/claude-code:${TAG}|" $TEMPLATE | kubectl apply -f -
--- a/plugins/tools/agent-sandboxing/commands/build-image.md
+++ b/plugins/tools/agent-sandboxing/commands/build-image.md
@@ -24,8 +24,8 @@ Stage them into a build context:
 
 ```bash
 BUILD_DIR=$(mktemp -d)
-cp ${CLAUDE_PLUGIN_ROOT}/templates/Dockerfile.claude-code $BUILD_DIR/Dockerfile
-cp ${CLAUDE_PLUGIN_ROOT}/templates/mise.toml.claude-code $BUILD_DIR/mise.toml
+cp <CLAUDE_PLUGIN_ROOT>/templates/Dockerfile.claude-code $BUILD_DIR/Dockerfile
+cp <CLAUDE_PLUGIN_ROOT>/templates/mise.toml.claude-code $BUILD_DIR/mise.toml
 ```
 
 ### 2. Build
@@ -65,7 +65,7 @@ Or re-apply from the plugin template with placeholders substituted:
 
 ```bash
 sed -e "s|REGISTRY/claude-code:TAG|${REGISTRY}/claude-code:${TAG}|" \
-  ${CLAUDE_PLUGIN_ROOT}/templates/SandboxTemplate.kata.yaml | kubectl apply -f -
+  <CLAUDE_PLUGIN_ROOT>/templates/SandboxTemplate.kata.yaml | kubectl apply -f -
 ```
 
 ### 5. Clean up
--- a/plugins/tools/agent-sandboxing/commands/provision.md
+++ b/plugins/tools/agent-sandboxing/commands/provision.md
@@ -35,7 +35,7 @@ Typical names: `claude-code-kata`, `claude-code-gvisor`, `claude-code-kina`.
 ### 3. Render the claim
 
 ```bash
-nu ${CLAUDE_PLUGIN_ROOT}/scripts/render-claim.nu \
+nu <CLAUDE_PLUGIN_ROOT>/scripts/render-claim.nu \
   --session-id $SESSION_ID \
   --template $TEMPLATE_NAME \
   ${SHUTDOWN_TIME:+--shutdown-time $SHUTDOWN_TIME} \
@@ -47,7 +47,7 @@ The script reads `templates/SandboxClaim.session.yaml`, substitutes placeholders
 ### 4. Wait for Bound
 
 ```bash
-nu ${CLAUDE_PLUGIN_ROOT}/scripts/wait-bound.nu --session-id $SESSION_ID --timeout 120
+nu <CLAUDE_PLUGIN_ROOT>/scripts/wait-bound.nu --session-id $SESSION_ID --timeout 120
 ```
 
 The script polls `kubectl get sandboxclaim` every 2 seconds until `status.phase == Bound` or the timeout elapses. Reports the bound pod name on success; reports controller log excerpts on timeout.
--- a/plugins/tools/agent-sandboxing/commands/status.md
+++ b/plugins/tools/agent-sandboxing/commands/status.md
@@ -16,7 +16,7 @@ Show the live state of all SandboxClaims for at-a-glance triage.
 ### 1. Render the table
 
 ```bash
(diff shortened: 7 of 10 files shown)


vinnie357:
# Gate 3 — adversarial review

**Verdict: APPROVE.** Four findings, none blocking. F1 is a factually wrong sentence in the PR body and should be corrected before merge; F2–F4 are follow-ups.

I was briefed to defeat this change. Below is what I attacked, what survived, and the one attack that produced a measurement worth keeping.

---

## The premise — does the braced form actually expand at load time?

The author flagged this as their own gap. Probing it:

- **Confirmed for skill bodies**, by two independent traces. `claude-skills-172` records a reproduction (loading `claude-hooks` returned every token already replaced by an absolute cache path). The surviving remediation is still readable in this machine's installed cache — `claude-code/0.2.34/skills/claude-hooks/SKILL.md:32`: *"Written bare here (not as `${...}`) because the braced form itself expands when this skill loads."*
- **The replacement form is proven safe by direct observation in this session.** The `claude-skills` SKILL.md body loaded into my context contains `` bash <CLAUDE_SKILL_DIR>/scripts/foo.sh `` and it arrived with `<CLAUDE_SKILL_DIR>` intact. No dollar sign means no substitution rule fires. Safe by construction, and empirically safe.
- **Not re-derived for `commands/*.md` specifically** — see "Could not verify" below.

## Attack that failed, and what it measured

**Hypothesis: `<...>` breaks the bash blocks these commands instruct Claude to run.** In bash, `nu <CLAUDE_PLUGIN_ROOT>/scripts/x.nu` parses as an input redirect from a file named `CLAUDE_PLUGIN_ROOT` *plus an output redirect to `/scripts/x.nu`* — so verbatim execution might create a file at an absolute path.

Tested in a sandboxed dir. Bash processes the failing input redirect first, aborts with exit 1, and **does not create the output target** (`ls` confirmed nothing created). Verbatim execution fails cleanly. Attack failed.

Two measurements that came out of it, both favouring the change:

- `CLAUDE_PLUGIN_ROOT` is **not** an env var in the Bash tool environment (`${CLAUDE_PLUGIN_ROOT:-NOT_SET}` → `NOT_SET`). So the pre-existing braced form would have expanded to a bare `/templates/...` — a plausible-looking absolute path that silently points nowhere. The change is a net improvement, not a regression.
- Residue, pre-existing and not introduced here: `commands/bootstrap.md:90` is an *assignment* (`TEMPLATE=<CLAUDE_PLUGIN_ROOT>/templates/...`). It fails **and leaves `TEMPLATE` empty**; the next line's `sed ... $TEMPLATE | kubectl apply -f -` would then read stdin and hang. The braced form failed the same way.

## Everything else I attacked, and what survived

**Completeness — survived, zero misses.** Repo-wide grep on the branch for `${CLAUDE_PLUGIN_ROOT}` / `${CLAUDE_SKILL_DIR}` / `${CLAUDE_PROJECT_DIR}` over all tracked files. Every remaining hit classified:
- *(a) correctly-excluded manifest/frontmatter* — `core/skills/security/SKILL.md:10`, `core/hooks/hooks.json:10`, `claude-code/hooks/hooks.json:10`
- *(b) correctly-excluded reference/template/script* — `claude-hooks/references/hook-examples.md` (5), `claude-hooks/templates/plugin-hook.md` (2), `plugin-marketplace/references/marketplace-examples.md` (9), `plugin-marketplace/references/schema-specification.md` (7), `claude-code/skills/sources.md` (2), `claude-code/hooks/session-start.sh:35`
- *(c) misses* — **none**

Set I ran: every tracked `SKILL.md` and `commands/*.md` in the repo, frontmatter stripped at the second `---`. Body-only census, main → branch: **9 bare + 8 braced → 8 bare + 14 angle, 0 braced.** Also checked the four command bodies for other load-time hazards: no bang-injection lines, no `$ARGUMENTS`, no `$N`, and no `arguments:` frontmatter — so `$USER` / `$TEMPLATE` / `$SESSION_ID` are undeclared and render unchanged. Clean.

**Correct variable per site — survived.** `skills/sandbox/` contains exactly one file, `SKILL.md`. No `scripts/` or `templates/` under the skill dir, so `CLAUDE_SKILL_DIR` would be wrong at every site. `CLAUDE_PLUGIN_ROOT` is right at all 14.

**Referenced paths exist — survived, 11/11.** Confirmed on the branch with `git cat-file -e`: all four `templates/SandboxTemplate.*` / `SandboxClaim.session.yaml` / `NetworkPolicy.anthropic.yaml` / `Dockerfile.claude-code` / `mise.toml.claude-code`, all three `scripts/*.nu`, and plugin-root `mise.toml`. This is the exact second-order defect `claude-skills-128` hit (a `scripts/` dir that did not exist). Not repeated here.

**Version bumps — survived, 6/6.** All verified main→branch in both files. A diff of the full marketplace version list shows exactly three lines changed, no stray bumps. `description` matches between marketplace.json and plugin.json for both changed plugins.

**Baseline — survived.** `test/quality-baseline.json` byte-identical to main. Debt count 49 on **both** — I ran `mise test` against main and against the branch in my own clone rather than trusting the number.

**Gates re-run independently** in a scratchpad clone (shared working tree never checked out): `mise test` **EXIT=0**, `mise run test:version-bumps origin/main` **EXIT=0**, `mise test:disclosure` **EXIT=0** (614 tracked files). Remote: both checks SUCCESS, `mergeStateStatus: CLEAN`.

---

## Findings

### F1 [medium] The PR body's house-style justification is false for this token

> "The bare angle-bracket form (`<CLAUDE_PLUGIN_ROOT>`) ... is the house style already used elsewhere in the claude-code plugin."

Measured: `git grep -c '<CLAUDE_PLUGIN_ROOT>' -- plugins/tools/claude-code/` returns **zero matches, across every file type**. The claude-code plugin uses `CLAUDE_PLUGIN_ROOT` **bare** at all 8 body sites, and documents that choice explicitly twice:

- `plugins/tools/claude-code/skills/claude-hooks/SKILL.md:37` — "Written bare here (not as `${...}`) because the braced form itself expands when this skill loads"
- `plugins/tools/claude-code/skills/claude-plugins/SKILL.md:71` — "It is shown bare here because the braced form expands when *this skill* loads ... That is the same reason script commands on this page use `CLAUDE_SKILL_DIR` bare."

The angle form *is* house style in the claude-code plugin — for `CLAUDE_SKILL_DIR` (9 angle / 5 bare), not for `CLAUDE_PLUGIN_ROOT`. The code is fine either way; the justification is not. Correct the sentence before merge.

### F2 [medium] The change makes angle the majority notation for `CLAUDE_PLUGIN_ROOT` and splits it along plugin lines, unexplained

Body-only counts, main → branch: 9 bare + 8 braced → **8 bare + 14 angle**. Beyond the 8 defect sites, the PR also converts the one pre-existing **bare** site (`skills/sandbox/SKILL.md` scripts bullet — correct under the incumbent convention) to angle and deletes the footnote that explained it.

Result: `plugins/tools/claude-code/**` says "bare, and here is why"; `plugins/tools/agent-sandboxing/**` says angle, silently. Both forms are safe, so this is not blocking — but it should be settled repo-wide in one pass and written down, not left as a per-plugin coin flip.

On the dropped footnote specifically: deleting it is defensible, since `sandbox/SKILL.md` retains no bare form for it to explain, and the other bare sites all carry their own explanations. The notation switch is the issue, not the deletion.

Pre-existing and not introduced here, but worth folding into the same pass: `claude-plugins/SKILL.md:71` contradicts itself — it says "script commands on this page use `CLAUDE_SKILL_DIR` bare" while the page uses the angle form 4× and bare 1×.

### F3 [minor] The dagu softening swaps an unverified absolute for a differently-unverified claim, and its cross-reference does not deliver

Both new lines assert retention is **configurable** (`webui/SKILL.md:60` and `:99`). Checking both cited sources:

- `webui/references/monitoring.md:179` reads, in full: `5. **Archive old executions**: Keep history manageable` — one line in a Best Practices list. No setting, no mechanism, no pruning instructions. A reader sent there "for pruning old runs" finds nothing.
- `rest-api/references/api-endpoints.md:248` — `DELETE /dags/{dagName}/history/{requestId}` **does** exist. It establishes history is *deletable via the API*, which is enough to disprove "all executions are preserved". It is not evidence of *configurable retention*.
- Grepping `histretention|retentiondays|retention` across the entire dagu plugin: the **only two hits are the two lines this PR adds.** Zero prior grounding.

The direction is right and the old absolute was worse, so this is minor. Wording the repo's own evidence supports: *"Executions are logged and accessible; history can be deleted (`DELETE /dags/{dagName}/history/{requestId}`) — see the rest-api skill."* Alternatively cite dagu upstream's actual retention setting and record it in `sources.toml`. Also, `:99`'s label is still "**History preservation**", which now argues with its own sentence.

### F4 [minor] Third occurrence of this defect class, still no regression guard

`claude-skills-169`, `-172`, and now `-178` are the same bug: a braced `CLAUDE_*` token in loaded body content, fixed by hand at the discovered sites each time. Grepping `test/*.nu` finds no check that would fail on it (the `brace` hits in `validate-skills-quality.nu` are the unrelated `{{}}`/bash-injection vocab self-test).

Per `/core:restraint` "Bug fix = root cause, not symptom", the durable fix is a lint: braced `CLAUDE_(PLUGIN_ROOT|SKILL_DIR|PROJECT_DIR|SESSION_ID|EFFORT)` is an error in `SKILL.md` and `commands/*.md` **bodies**; allowed in frontmatter, `references/`, `templates/`, `hooks.json`, `plugin.json`, `marketplace.json`. The hand classification I ran for this review is exactly that rule, and it produced zero false positives over 614 tracked files. Follow-up bee, not a blocker.

---

## Could not verify

- **No live load of an agent-sandboxing *command* file.** The plugin is absent from this machine's `~/.claude/plugins/cache/vinnie357/`, so I could not run the "verify by loading, never by reading source" loop against a `commands/*.md`. The expansion mechanism is confirmed for SKILL.md bodies and the angle form's safety is confirmed by direct observation, but the command-file leg is inherited from `claude-skills-178`'s reasoning, not re-derived. Cheap closure: install 0.2.9 and invoke `/agent-sandboxing:status`.
- **Whether dagu upstream exposes a retention setting.** I stayed inside the repo deliberately; the repo contains no such evidence, which is the whole basis of F3.

Unrelated, so it is not lost when `claude-skills-178` closes: that issue also asks for a side filing on upstream `plugin.json` fields (`outputStyles`, `lspServers`, `experimental`) that the `claude-plugins` skill does not mention.

PR #191, opened 2026-08-01
fix(pm,beads,rust): correct bees prime data claim, beads package identity, rust gate naming

vinnie357:
Closes claude-skills-160, claude-skills-168, claude-skills-147 — three one-liners in three plugins, batched into one PR and one CI cycle.

- beads installation.md / SKILL.md: correct `npm install -g @anthropic/beads` → `npm install -g @beads/bd` and `brew install anthropic/tap/beads` → `brew install beads` (homebrew-core formula, no tap). `@anthropic/beads` is a 404 — it never existed — and `anthropic/homebrew-tap` does not exist either. Verified via `npm view @beads/bd` (real package v1.1.2, maintainer steveyegge, repo gastownhall/beads) and `brew info beads` (homebrew-core, v1.1.2). The repo moved: `steveyegge/beads` 301-redirects to `gastownhall/beads`.

- The Go install line's **module path** is correct and was left unchanged: released Go modules still declare `github.com/steveyegge/beads`, `proxy.golang.org` resolves `steveyegge@latest` to v1.1.2, and `go.mod` on gastownhall main declares the steveyegge module path. That compatibility rationale is stated verbatim in upstream's installation doc.

  **Correction, per Gate 3:** an earlier draft of this body claimed the Go install line "was already correct." The module path is; the **command shape is not**. Upstream documents exactly two supported go-install modes, both requiring an env prefix — `CGO_ENABLED=0 go install ...` (server-mode only, no embedded Dolt) and `CGO_ENABLED=1 GOFLAGS=-tags=gms_pure_go go install ...` (embedded-capable). `installation.md:21` ships a bare `go install github.com/steveyegge/beads/cmd/bd@latest`, which is neither. Upstream also advises using Homebrew, npm, or the install script unless you specifically need `go install`. That defect is pre-existing and outside bee 168's scope, so it is **not fixed here** — it is recorded so the claim entering main is accurate. Filed as a follow-up.

- pm agents: drop `bees prime` from the "as returned by these commands" enumeration in `pm-discovery.md` and `pm-lead.md`. `prime` returns a static workflow cheatsheet with zero issue data; the sibling commands in that list do supply it.

- rust: rename `rust-implementer.md`'s "Gate 1"/"Gate 2" step names to "Check 1"/"Check 2" so they stop colliding with `/core:git`'s Three-Gate Merge Policy, where those numbers mean local and remote CI. Repo-wide grep confirms nothing cites rust's old numbering.

## Gates

Gate 1 `mise test` exit 0; `mise run test:version-bumps origin/main` exit 0 (rust 0.1.10→0.1.11, pm 0.1.2→0.1.3, beads 0.1.2→0.1.3); gitleaks clean; `test/quality-baseline.json` untouched.

Gate 2 both remote checks pass.

Gate 3 APPROVE — https://github.com/vinnie357/claude-skills/pull/191#issuecomment-5152790971. Reviewer independently reproduced every registry claim, confirmed zero surviving `@anthropic/` references repo-wide, and ran `mise test` on both main and the branch to confirm the debt summary is byte-identical (49 debt entries, 4 dupe groups) rather than trusting the reported number.

## Follow-ups filed rather than widened into this PR

- The bare `go install` command shape above.
- The mise github backend `github:steveyegge/beads` names the repo holding release assets, which is now `gastownhall/beads`; it resolves only via the 301. Three sites, one of them outside the beads plugin (`plugins/core/skills/mise/templates/multi-arch.md`).
- `beads/skills/sources.toml` and `sources.md` cite `steveyegge.github.io/beads/`, which 404s; the live successor is `gastownhall.github.io/beads/`.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.91",
+      "version": "0.4.92",
       "category": "meta",
       "keywords": [
         "skills",
@@ -104,7 +104,7 @@
       "source": "./plugins/tools/beads",
       "strict": true,
       "description": "Beads (bd) distributed git-backed graph issue tracker: task management, dependency tracking, AI agent workflows",
-      "version": "0.1.2",
+      "version": "0.1.3",
       "category": "tools",
       "keywords": [
         "beads",
@@ -311,7 +311,7 @@
       "source": "./plugins/pm",
       "strict": true,
       "description": "Product management toolkit: harvest implementation-agnostic feature specs from prototypes with SDLC guardrails and author PRDs as implementation contracts",
-      "version": "0.1.2",
+      "version": "0.1.3",
       "category": "tools",
       "keywords": [
         "pm",
@@ -401,7 +401,7 @@
       "source": "./plugins/languages/rust",
       "strict": true,
       "description": "Rust programming skills: ownership, borrowing, lifetimes, async, best practices, anti-patterns, CLI apps, and adversarial-TDD worker agents",
-      "version": "0.1.10",
+      "version": "0.1.11",
       "category": "language",
       "keywords": [
         "rust",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.91",
+  "version": "0.4.92",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/languages/rust/.claude-plugin/plugin.json
+++ b/plugins/languages/rust/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "rust",
-  "version": "0.1.10",
+  "version": "0.1.11",
   "description": "Rust programming skills: ownership, borrowing, lifetimes, async, best practices, anti-patterns, CLI apps, and adversarial-TDD worker agents",
   "author": {
     "name": "vinnie357"
--- a/plugins/languages/rust/agents/rust-implementer.md
+++ b/plugins/languages/rust/agents/rust-implementer.md
@@ -34,8 +34,8 @@ Invoke the Skill tool for `/rust:ownership`, `/rust:async`, or `/rust:troublesho
 
 1. **Sync.** `git fetch`, checkout the feature branch, pull. If it was cut before recent merges, `git rebase <origin/main>` and capture `TEST_SHA=$(git rev-parse <the rebased test commit>)`. Read the frozen test file(s) — they are your binding spec.
 2. **Implement the smallest change.** Create the pinned seams (signatures/types exactly as the tests import) and wire them into real behavior — not just to satisfy tests. Idiomatic error handling: propagate with `Result`/`?`, no `unwrap()`/`expect()`/`panic!` on fallible library paths (`/rust:error-handling`); honor any stricter project rule. Match surrounding style.
-3. **Gate 1 — frozen tests untouched.** `git diff <TEST_SHA>..HEAD -- <test-files>` MUST be empty — paste it. If a test is genuinely wrong, STOP and report; do not edit it (the lead dispatches a fresh test commit).
-4. **Gate 2 — full CI green.** Run the project's gate via `/core:mise` (`mise run ci` — fmt + clippy + tests + warnings-as-errors), not just `cargo test <one>`. Paste the verbatim summary. Iterate until clean.
+3. **Check 1 — frozen tests untouched.** `git diff <TEST_SHA>..HEAD -- <test-files>` MUST be empty — paste it. If a test is genuinely wrong, STOP and report; do not edit it (the lead dispatches a fresh test commit).
+4. **Check 2 — full CI green.** Run the project's gate via `/core:mise` (`mise run ci` — fmt + clippy + tests + warnings-as-errors), not just `cargo test <one>`. Paste the verbatim summary. Iterate until clean.
 5. **Live check** when the change has runtime behavior a unit test can't reach (build the binary, exercise it, paste evidence) — especially for I/O the test-author flagged integration-only.
 6. **Commit & PR.** Conventional commit, NO attribution. Push. Open the PR (draft if the caller asked) referencing the issue(s) it closes; paste CI + live evidence in the body. Watch remote CI to green. Never merge — report the PR URL to the caller.
 
--- a/plugins/pm/.claude-plugin/plugin.json
+++ b/plugins/pm/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "pm",
-  "version": "0.1.2",
+  "version": "0.1.3",
   "description": "Product management toolkit: harvest implementation-agnostic feature specs from prototypes with SDLC guardrails and author PRDs as implementation contracts",
   "author": {
     "name": "vinnie357"
--- a/plugins/pm/agents/pm-discovery.md
+++ b/plugins/pm/agents/pm-discovery.md
@@ -37,7 +37,7 @@ The lead passes:
 
 **`integrations-deps-stack`**: Read the manifest file(s) present (`package.json`, `mix.exs`, `Cargo.toml`, `requirements.txt`, `go.mod`, or equivalent). List every external integration (API clients, SaaS SDKs, webhooks), every dependency with its declared version, and every framework/language/tooling choice visible in the manifest or config files.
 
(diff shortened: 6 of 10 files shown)


vinnie357:

**Verdict: APPROVE**, conditional on one PR-body correction (MAJOR-1). The three fixes are correct and complete, both gates reproduce green, and no regression. Every registry claim in the PR body reproduced exactly. The findings below are one false claim in the description, one consistency judgment, and two follow-ups.

Brief was to defeat the change. What follows is what I attacked, what broke, and what survived.

---

### Findings

#### MAJOR-1 — the PR body's Go-install claim is contradicted by upstream

The body states:

> The Go install line was already correct and left unchanged — confirmed against the upstream README

The module **path** is correct — I confirmed that three independent ways (see "survived" below). The install **command** is not what upstream documents. From `gastownhall/beads` `docs/getting-started/installation.md` (fetched via the GitHub contents API):

> L112: Use Homebrew, npm, or the install script if you do not specifically need `go install`.
> L114: `go install` has two supported modes that give different capabilities:
> L116: **Server-mode only (nocgo, simplest):** `CGO_ENABLED=0 go install github.com/steveyegge/beads/cmd/bd@latest` … Produces a server-mode-only binary — you must run an external `dolt sql-server` and use `bd init --server`.
> L117: **Embedded-capable (cgo):** `CGO_ENABLED=1 GOFLAGS=-tags=gms_pure_go go install github.com/steveyegge/beads/cmd/bd@latest` … Produces a binary with the default embedded-Dolt backend — `bd init` Just Works.
> L119: ICU headers are not required. The embedded-capable command uses `gms_pure_go` so go-mysql-server uses Go's stdlib regexp instead of ICU.

`references/installation.md:21` ships a bare:

```bash
go install github.com/steveyegge/beads/cmd/bd@latest
```

That is **neither** documented mode — both carry a required env prefix. Upstream's README also emits the `CGO_ENABLED=0` form (README L162), never the bare one. Consequences a reader hits: with cgo off they get a server-mode-only binary whose plain `bd init` does not work as the rest of this skill assumes; with cgo on (the default when a C compiler is present) they build without `-tags=gms_pure_go`, which upstream L119 identifies as the flag that avoids needing ICU. I did **not** run either build — this is read from upstream's documentation, not measured.

The line itself is pre-existing and outside claude-skills-168's stated scope. The **claim** is not: it will be squash-merged into main as the durable record of this change. Cheapest disposition is a one-line body edit ("the module path was verified correct and left unchanged; the command form is not upstream's documented invocation — tracked separately"). Fixing the line properly (env prefix + server-mode caveat) is the better fix if you want it in this PR.

#### MODERATE-2 — the file now carries three identities; only one has an upstream reason

After this PR `references/installation.md` contains:

| Method | Identity | Justified? |
|---|---|---|
| npm | `@beads/bd` | yes — new identity, correct |
| Homebrew | `beads` (homebrew-core) | yes — new identity, correct |
| Go install | `github.com/steveyegge/beads` | **yes** — upstream documents this as deliberate |
| mise | `github:steveyegge/beads` | **no compatibility reason** |

The go-module compatibility argument is real and verified, but it does **not** extend to the mise github backend. That backend names the GitHub repo holding the release assets, and that repo is `gastownhall/beads`. It resolves today only because GitHub 301-redirects — I confirmed `gh api repos/steveyegge/beads/releases/latest` returns `v1.1.2`. Redirect-dependent config is fragile, and a reader has no way to distinguish the deliberate steveyegge reference from the incidental one.

Same `github:steveyegge/beads` also appears in `plugins/tools/beads/skills/beads/templates/mise.toml` and `plugins/core/skills/mise/templates/multi-arch.md` — three sites, one outside the beads plugin.

Suggested: point the mise backend at `github:gastownhall/beads`, and add one line on the Go section stating the old module path is intentional. That turns an ambiguity into a documented decision. Filing it as a follow-up is also acceptable.

#### MINOR-3 — `pm-discovery.md:40` says "use only" and then permits a sixth command

The line reads "use only: `bees list`, `bees show`, `bees ready --json`, `bees dep list`, `bees comment list`." and the next sentence adds "`bees prime` may also be invoked for workflow context". Strictly read, "use only" and "may also" contradict. `pm-lead.md:94` solves the identical problem better by keeping the caveat inside the sentence as a parenthetical. Cosmetic — no agent is misled about what `prime` returns, which was the bee's actual point. Suggest matching pm-lead's shape.

#### MINOR-4 (pre-existing, follow-up) — dead documentation URL

`plugins/tools/beads/skills/sources.toml:15` and `sources.md:10` both cite `https://steveyegge.github.io/beads/`, which returns **HTTP 404**. `https://gastownhall.github.io/beads/` returns 200; the repo's declared `homepage` is `https://beads.gascity.com`. Not introduced here and `sources.toml` `notes` already flags `sources.md` as predating the move — but this is the beads-identity-migration PR, so it is the natural home for the fix.

#### INFO-5 — `current_version` in beads `sources.toml`

Asked whether this is now inconsistent with the registries. It is not inconsistent — it is `"unknown"`, and `notes` already records "latest tag v1.1.2". Four independent sources now agree on 1.1.2 (npm `@beads/bd`, homebrew-core, GitHub latest release, `proxy.golang.org`). Leaving it unpinned matches this repo's own graphify precedent, which explicitly declines to bump `current_version` without re-verifying skill content against the new version. No action needed.

---

### What I attacked that survived

1. **Every registry claim reproduced.** `npm view @anthropic/beads` → E404. `npm view @beads/bd` → `1.1.2`, repo `git+https://github.com/gastownhall/beads.git`, maintainer `steveyegge`. `formulae.brew.sh/api/formula/beads.json` → real, `tap: homebrew/core`, stable `1.1.2`, homepage `github.com/gastownhall/beads`. `github.com/anthropic/homebrew-tap` → 404 both via HTTP and `gh api`. `github.com/steveyegge/beads` → **301** (body says 302; immaterial) to `gastownhall/beads`; `gh api repos/steveyegge/beads` resolves to `gastownhall/beads`, not archived.

2. **The go module path non-change is right.** I tried three ways to break it and all three confirmed the author: `proxy.golang.org/github.com/steveyegge/beads/@latest` → `v1.1.2`; `go.mod` on `gastownhall/beads` main declares `module github.com/steveyegge/beads`; upstream doc L121 says verbatim "Use the `github.com/steveyegge/beads` path for `go install`. The repository now lives under `gastownhall/beads`, but released Go modules still declare `github.com/steveyegge/beads` for compatibility."

3. **Completeness.** `git grep` on the PR ref finds **zero** remaining `@anthropic/`, `anthropic/tap`, or `anthropic.com/beads` anywhere in the repo. All three sites are fixed — `installation.md` ×2 and `SKILL.md:37`, the one the bee did not flag. No site the author missed.

4. **160 elsewhere in pm.** Three other pm files still mention `bees prime` — `skills/sources.md:10`, `spec-harvest/SKILL.md:56`, `spec-harvest/references/bees-to-spec.md:12`. All three frame it as read-only *surface* (write-vs-read), not as an issue-data source, and bees-to-spec.md:12 states "contains no issue data" explicitly. The false implication does not survive.

5. **147's rename.** `git grep "Gate"` across the entire rust plugin returns **zero** hits — the rename is complete, not partial. `**Check 1`/`**Check 2` collide with nothing repo-wide. `rust-test-author.md` never cites the implementer by step number; it emits "TEST COMMIT SHA: `<sha for the implementer's frozen-tests check>`" — which now reads *more* consistently under "Check 1" than it did under "Gate 1". The pair remains internally consistent on freezing: test-author freezes on commit and must not write implementation; implementer must not modify tests; both state the lead inspects the post-phase diff. Only other cross-reference is `plugins/tools/rig/skills/rig/SKILL.md:111`, which names the pair, not its steps.

6. **All eight version edits verified programmatically**, plugin.json against marketplace.json: all-skills 0.4.92/0.4.92, beads 0.1.3/0.1.3, pm 0.1.3/0.1.3, rust 0.1.11/0.1.11 — all MATCH, and descriptions agree between the two files for all four.

7. **`test/quality-baseline.json` untouched** — confirmed via `git diff --stat origin/main...branch -- test/quality-baseline.json` (empty).

8. **No silent count growth.** Ran `mise test` in an isolated clone on the PR branch (`ae5691a`) and on main (`e4149da`). Both exit 0. Summary lines are identical: `Total skills: 102`, `Perfect score: 57/102`, `49 debt/bug entries`, `4 cross-file duplicate group(s)`, `agents/commands/hooks surfaces: all clean`.

9. **mise asset globs.** Checked the actual v1.1.2 release asset names against the patterns in the mise block: `beads_*_linux_amd64.tar.gz` matches `beads_1.1.2_linux_amd64.tar.gz`, `beads_*_darwin_arm64.tar.gz` matches `beads_1.1.2_darwin_arm64.tar.gz`. Survives.

**Process note on my own work:** my first completeness sweep (a shell loop over `git ls-tree` piping into `grep /dev/stdin`) returned empty for *both* the pm and rust checks — a silent under-match, not a result. I re-ran with `git grep` against the ref and got real hits. Every completeness conclusion above rests on the `git grep` run, not the loop.

### Could not verify

- Whether mise's github backend follows GitHub's repo-rename redirect end to end. I verified the **API** redirect resolves (`gh api repos/steveyegge/beads/releases/latest` → v1.1.2), not mise's client behavior. Untested either way — this is why MODERATE-2 recommends removing the dependency rather than relying on it.
- Whether the bare `go install` command builds. Not run; MAJOR-1 rests on upstream's documentation, not measurement.

PR #192, opened 2026-08-01
fix(claude-code): fix dead describe==record guard in validate-marketplace.nu

vinnie357:
- Fix `validate-marketplace.nu`'s object-source guard: `(describe) == "record"` only matches an empty `{}` — `describe` renders a real record as `record<field: type, ...>`, so every source object silently skipped validation (bogus source types, a missing `source` field, and non-string/non-record values like a bare integer all passed with 0 errors)
- Add an `is-record` helper next to the existing `is-list` helper (same `describe --detailed | get type` discipline) and use it for the source-object branch plus the `author`, `hooks`, and `mcpServers` checks, which had the identical dead-comparison bug — `author`/`hooks`/`mcpServers` previously *always* errored when a valid record was supplied (verified against the pre-fix script)
- Add a catch-all branch so a `source` that is neither string nor record (e.g. a bare integer) is now reported instead of silently passing
- Expand the object-source allowlist from `[github, url]` to `[github, url, git-subdir, npm]`, confirmed against live docs (code.claude.com/docs/en/plugin-marketplaces)
- Add a `..`-path-traversal check and a directory-vs-file check on local string sources (upstream: "Don't use `../` to reference paths outside the marketplace root"); previously a `./`-prefixed path with `..` further in, or a path resolving to a file, passed silently
- Document `git-subdir` and `npm` source formats in `references/schema-specification.md`, and note the `..`/directory rules in its validation checklist
- Add an embedded `--self-test` for `is-list`/`is-record` (12 cases)
- Bump `claude-code` plugin version 0.2.44 → 0.2.45 in both `plugin.json` and `marketplace.json`

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -119,7 +119,7 @@
       "source": "./plugins/tools/claude-code",
       "strict": true,
       "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
-      "version": "0.2.44",
+      "version": "0.2.45",
       "category": "tools",
       "keywords": [
         "claude-code",
--- a/plugins/tools/claude-code/.claude-plugin/plugin.json
+++ b/plugins/tools/claude-code/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-code",
-  "version": "0.2.44",
+  "version": "0.2.45",
   "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/claude-code/skills/plugin-marketplace/references/schema-specification.md
+++ b/plugins/tools/claude-code/skills/plugin-marketplace/references/schema-specification.md
@@ -76,6 +76,28 @@ Complete JSON schema for Claude Code plugin marketplaces.
 }
 ```
 
+### Git Subdirectory (Object)
+```json
+"source": {
+  "source": "git-subdir",
+  "url": "https://github.com/acme-corp/monorepo.git",
+  "path": "plugins/formatter"
+}
+```
+Sparse-clones just the subdirectory — use for a plugin that lives inside a monorepo.
+
+### npm Package (Object)
+```json
+"source": {
+  "source": "npm",
+  "package": "@acme/claude-plugin",
+  "version": "2.1.0"
+}
+```
+Installed via `npm install`. Not supported for marketplaces distributed through Organization settings (Team/Enterprise plans) — those accept `github`, `url`, and `git-subdir` only.
+
+Valid `source` values for the object form: `github`, `url`, `git-subdir`, `npm`.
+
 ## Field Validation Rules
 
 ### name
@@ -198,7 +220,8 @@ Complete JSON schema for Claude Code plugin marketplaces.
 - [ ] `plugins` is an array (can be empty)
 - [ ] Each plugin entry has `name` and `source`
 - [ ] Versions use semantic versioning (if specified)
-- [ ] Source paths exist (for relative paths)
+- [ ] Source paths exist and resolve to a directory (for relative paths)
+- [ ] Relative source paths don't contain `..` (must stay inside the marketplace root)
 - [ ] No circular dependencies
 - [ ] All dependency references exist in marketplace
 - [ ] `${CLAUDE_PLUGIN_ROOT}` used correctly in paths
--- a/plugins/tools/claude-code/skills/plugin-marketplace/scripts/validate-marketplace.nu
+++ b/plugins/tools/claude-code/skills/plugin-marketplace/scripts/validate-marketplace.nu
@@ -3,11 +3,22 @@
 # Validate Claude Code marketplace.json file
 #
 # Usage: nu validate-marketplace.nu <path-to-marketplace.json> [--verbose]
+#   nu validate-marketplace.nu --self-test  # verify the is-list / is-record helpers
 
 def main [
-  marketplace_path: string  # Path to marketplace.json file
-  --verbose                 # Show detailed validation output
+  marketplace_path?: string  # Path to marketplace.json file (required unless --self-test)
+  --verbose                  # Show detailed validation output
+  --self-test                # Run embedded self-test for is-list / is-record and exit
 ] {
+  if $self_test {
+    exit (run-self-test)
+  }
+
+  if $marketplace_path == null {
+    print $"(ansi red_bold)Error:(ansi reset) marketplace_path is required unless --self-test is passed"
+    exit 1
+  }
+
   print $"(ansi green_bold)Validating marketplace:(ansi reset) ($marketplace_path)"
   print ""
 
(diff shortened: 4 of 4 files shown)


vinnie357:

**Verdict: APPROVE** (2 non-blocking findings, 1 nit, 1 question)

Brief was to defeat this change. I built my own fixtures rather than trusting the author's, re-derived every count and claim, and independently re-ran both gates. Nothing I attacked broke it. Details below, including what failed to land.

Reviewed read-only from a scratchpad clone of `fix/validate-marketplace-record-guard` @ `3963913`. No checkout in the shared tree.

---

### What I attacked, and what survived

**1. The allowlist as a factual claim about upstream docs — my highest-probability blocker. It did not land.**

I fetched `code.claude.com/docs/en/plugin-marketplaces` myself rather than accepting the author's reading. The Plugin sources table lists exactly five forms:

| Source | Type |
|---|---|
| Relative path | `string` (e.g. `"./my-plugin"`) |
| `github` | object |
| `url` | object |
| `git-subdir` | object |
| `npm` | object |

The "npm not supported" text sits inside a `<Note>` that opens *"If you distribute this marketplace through Organization settings > Plugins on a Team or Enterprise plan, **different source rules apply**"* and then reads *"Plugin sources of type `github`, `url`, and `git-subdir` are supported. `npm` sources are not."* The scoping in the PR body is correct: that caveat governs org-sync distribution, not general marketplaces, which have a whole `### npm packages` section documenting `"source": "npm"` with `package` / `version?` / `registry?`.

So `[github, url, git-subdir, npm]` is **exactly** the complete documented object-source set — not over-broad (nothing invalid admitted) and not under-broad (nothing valid still rejected). The `schema-specification.md` field lists match the table for both new types, and the `..` sentence quoted in the code comment is verbatim upstream: *"Don't use `../` to reference paths outside the marketplace root."*

**2. The scope expansion. It is justified — credit it, don't dock it.**

I reproduced the claimed false positives against the **original** script with my own fixtures. Each of these is valid input:

| Fixture | ORIG (main) | NEW (this PR) |
|---|---|---|
| `author: {name: "vinnie357"}` | ✗ ERROR `'author' must be an object` | clean |
| `hooks: {PreToolUse: "x"}` | ✗ ERROR `'hooks' must be a string or object` | clean |
| `mcpServers: {srv: {command: "c"}}` | ✗ ERROR `'mcpServers' must be a string or object` | clean |
| all three at once | ✗ 3 errors | clean |

These were real false positives rejecting valid records, not speculative cleanup. One dead comparison at four sites is one bug; fixing it once at the shared helper is `/core:restraint`'s "grep every caller ... fix the shared function once", and patching only the `source` site would have left three siblings broken. This is correct restraint, not scope creep.

I also checked the blast radius was fully covered: grepping `describe) == "record"` / `!= "record"` / `not-in ["string", "record"]` across `test/*.nu`, `mise.toml`, and `.claude-plugin/scripts/*.nu` returns **no hits**. Four sites in one file was the whole of it, and all four are fixed. No CI-facing validator carries the same bug.

**3. Is the self-test decorative? (the claude-skills-195 defect class) — no. I mutated the fix.**

Reverted `is-record` to the broken `($value | describe) == "record"`:
```
❌ is-record self-test: non-empty record (renders as record<...>) — want true, got false
❌ is-record self-test: single-field record — want true, got false
EXIT=2
```
Reverted `is-list` to `(describe) starts-with "list"`:
```
❌ is-list self-test: non-empty list of records (renders as table<...>) — want true, got false
EXIT=1
```
Both guards go red under their own target change. 12 cases confirmed, exit 0 unmutated.

**4. Malformed input the author did not try — including the empty-record edge.**

| Fixture | ORIG | NEW |
|---|---|---|
| `source: {source: "svn"}` | silently clean | ✗ `invalid source type: svn` |
| `source: {repo: "x/y"}` (no `source` field) | silently clean | ✗ `source object missing 'source' field` |
| `source: 42` | silently clean | ✗ `must be a string or object, got: int` |
| `source: ["a","b"]` | silently clean | ✗ `... got: list<string>` |
| `source: true` | silently clean | ✗ `... got: bool` |
| `source: {}` (empty record) | ✗ missing 'source' field | ✗ same |
| `source: {source:"github", deep:{a:{b:{c:1}}}}` | clean | clean |

The empty-record case is the tell: `{}` is the **one** value that renders as bare `"record"`, and it is the one case that behaved identically before and after — which is exactly the diagnosis in the PR body, confirmed from the other direction.

**5. Masking.** Nothing aborts; everything accumulates into `$errors` and prints at the end. A 2-plugin fixture with a kebab-case violation *and* a bogus source type reports **both**:
```
✗ Invalid plugin name: 'Bad_Name' (must be kebab-case)
✗ Plugin 'Bad_Name' invalid source type: svn
```

**6. `..` check for false positives — none found.** `path split` matches exact segments, so none of these fire: `./plugins/foo..bar`, `plugins/..foo`, `./...`, `./plugins/core/` (trailing slash), `./a/./b`. Genuine traversal does: `./plugins/../../etc` → error (ORIG merely warned about a path resolved *outside* the repo root). Also fires correctly when `metadata.pluginRoot` is set and the source is bare `../../etc`.

**7. `path type` portability — no regression.** A symlinked directory passes clean, because `path expand` resolves the symlink before `path type` runs. A nonexistent path never reaches `path type` at all — `path exists` is checked first and short-circuits to the pre-existing warning. Trailing slash is normalised by `path split`/`path join`.

**8. Claims in the PR body, re-derived not inherited.**
- Real `.claude-plugin/marketplace.json`: **0 errors, 0 warnings** on both ORIG and NEW. No regression on the repo's own file.
- Version bump: `plugin.json` = `0.2.45`, `marketplace.json` entry = `0.2.45`. Descriptions match; `jq -S` diff of `keywords` is clean.
- `test/quality-baseline.json`: `git diff origin/main` is **0 lines**. No new key, no count grown.
- Gate 1 re-run independently in my own clone: `mise run test` → **exit 0**, zero failure markers.
- Gate 2: `gh pr checks 192` → both checks pass.

---

### Findings

**[suggestion] The `--self-test` is not wired into CI — it will never run on its own.**

The house pattern is that every `--self-test` has a mise task invoking it: `mise.toml:248` (`test:skills-quality`), `:258` (`test:core-list`), `:268` (`test:disclosure`), `:278` (`test:sources`). Nothing in `mise.toml`, `.github/workflows/validate.yml`, or `test/` invokes `plugins/tools/claude-code/skills/plugin-marketplace/scripts/validate-marketplace.nu` at all — a grep for that path across `*.toml`/`*.yml`/`*.nu`/`*.json` outside the skill directory returns no hits. `test:marketplace` is a **separate inline reimplementation** in `mise.toml`, not this script.

Mitigating, and why this is not blocking: skill-*bundled* scripts are uniformly unwired in this repo (`container-*.nu`, `gitleaks.nu`, `restraint-audit.nu`, `sources-*.nu` are all in the same position). This PR follows the norm rather than breaking it, and I proved the guard genuinely fails under mutation — it is unwired, not decorative. Suggest a follow-up bee for a `test:skill-scripts` task that runs the self-tests of skill-bundled scripts, so the next refactor of this file is caught by CI rather than by a human remembering.

**[suggestion] `sources.toml` `last_checked` not refreshed.**

`plugins/tools/claude-code/skills/sources.toml:173-181` — `claude-code-plugin-marketplaces-docs` still reads `last_checked = "2025-11-15"`. This PR consulted that exact URL live and materially extended `schema-specification.md` from it (two new source formats, a scoped-caveat sentence, two checklist items). The freshness ledger now understates when the page was actually read. `test:sources` passes either way, so this is convention hygiene rather than a rule violation — but it is the ledger's whole purpose. One-line fix; adjacent to the in-flight claude-skills-176 work.

**[note, pre-existing, out of scope] A non-record plugin *entry* crashes instead of being reported.**

`"plugins": ["a-string"]`, `"plugins": [42]`, and `name` as an int abort with a raw Nushell `nu::shell::only_supports_this_input_type` / `cant_convert` instead of a validator error line, masking any other finding in the file. **Both ORIG and NEW exit 1**, so it fails loudly and never silently passes, and this PR does not change the behaviour — its new catch-all covers a bad `source` *value*, not a bad *plugin entry*. Flagging for a follow-up bee only; fixing it here would be genuine scope creep.

**[question] File-as-source is a warning, not an error.** A source resolving to a file is not a valid plugin directory under the docs ("Local directory within the marketplace repo"), so `exit 0` is preserved for a marketplace that can't install. It is consistent with the neighbouring "path not found" warning, so I read this as deliberate — confirming only that it was a choice.

**[nit] Mixed idiom at `hooks`/`mcpServers`** (lines ~296, ~305): `let X_type = $plugin.X | describe` for the string half, `is-record` for the record half. Correct, and it keeps `$X_type` for the verbose print — reads slightly inconsistently next to the pure-`is-record` `author` check. Leave it.

---

### What I could not verify

- Whether `claude plugin validate` (the `test:claude` step) exercises object source types at all. It passed locally and in CI; I did not probe what it actually checks.
- Runtime behaviour of Claude Code installing an `npm` or `git-subdir` source. I verified the documentation, not the installer.
- The allowlist is pinned to docs fetched **2026-08-01**. If upstream adds a sixth source type, the allowlist silently becomes over-strict; nothing version-pins that page (see the `sources.toml` finding).

PR #193, opened 2026-08-01
feat(core): default /core:setup to bees, document bees enum-validation gaps

vinnie357:
- claude-skills-154: `/core:setup` and `plugins/core/agents/setup.md` default new-project bootstrap to bees (`bees init`, `bees create`, `bees dep add`, `bees label add` — bees has no `--labels` flag on create); beads stays available behind `--beads`, preserving `--stealth`/`--contributor` as beads-only sub-flags since bees has no equivalent init modes
- claude-skills-159: documents bees 0.4.0's missing enum validation on `create -t`, `update -s`/`-t`, and `dep add -t` — a typo is stored verbatim and silently excludes the issue from `bees ready`; also resolves the `parent` vs `parent-child` dep-type naming inconsistency and the `story` vs `chore` issue-type inconsistency between `--help` and `bees prime`
- bumps core plugin 0.2.45 → 0.2.46 and all-skills 0.4.91 → 0.4.92 in both plugin.json and marketplace.json

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.94",
+      "version": "0.4.96",
       "category": "meta",
       "keywords": [
         "skills",
@@ -143,7 +143,7 @@
       "source": "./plugins/core",
       "strict": true,
       "description": "Essential development skills: Git, TDD, code review, gitleaks secret scanning, documentation, restraint, anti-fabrication, twelve-factor apps, the agent loop, bees issue tracking, mise, nushell, and Apple Container",
-      "version": "0.2.45",
+      "version": "0.2.47",
       "category": "development",
       "keywords": [
         "git",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.94",
+  "version": "0.4.96",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/core/.claude-plugin/plugin.json
+++ b/plugins/core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "core",
-  "version": "0.2.45",
+  "version": "0.2.47",
   "description": "Essential development skills: Git, TDD, code review, gitleaks secret scanning, documentation, restraint, anti-fabrication, twelve-factor apps, the agent loop, bees issue tracking, mise, nushell, and Apple Container",
   "author": {
     "name": "vinnie357"
@@ -39,6 +39,8 @@
   ],
   "agents": [
     "./skills/bees/agents/bees-worker.md",
-    "./skills/bees/agents/bees-manager.md"
+    "./skills/bees/agents/bees-manager.md",
+    "./agents/setup.md",
+    "./agents/gcms.md"
   ]
 }
--- a/plugins/core/agents/setup.md
+++ b/plugins/core/agents/setup.md
@@ -1,30 +1,46 @@
 ---
 name: setup
-description: Bootstraps a new project with beads task tracking, discovers project characteristics, and creates dependency-linked tasks for mise, CI, and gitleaks configuration.
+description: Bootstraps a new project with bees task tracking (default) or beads (--beads), discovers project characteristics, and creates dependency-linked tasks for mise, CI, and gitleaks configuration.
 tools: Bash, Read, Glob, Grep
 model: sonnet
 ---
 
-You are a project bootstrapping agent. Your role is to initialize beads task tracking, discover project characteristics, and create a structured set of beads tasks with dependencies.
+You are a project bootstrapping agent. Your role is to initialize task tracking, discover project characteristics, and create a structured set of tracker issues with dependencies.
 
 ## Input
 
 The user may pass flags:
-- `--stealth` — Use `bd init --local` for local-only beads
-- `--contributor` — Use `bd init --pull` for read-only beads
+- `--beads` — Use beads (`bd`) instead of bees. Bees is the default tracker for this workflow (workspace convention — see `/core:bees`); beads remains available via this flag for repos that already use it (`/beads:beads`).
+- `--stealth` (beads only, requires `--beads`) — Use `bd init --local` for local-only beads
+- `--contributor` (beads only, requires `--beads`) — Use `bd init --pull` for read-only beads
 
-If no flags are provided, use standard `bd init`.
+If `--stealth` or `--contributor` is passed without `--beads`, note in the Phase 3 summary that the flag is beads-only and was ignored, then proceed with the default bees flow. Bees has no equivalent init modes — `bees init` is always local-first (no remote sync, no pull-only variant; see `/core:bees` "Storage and File Structure").
 
-## Phase 1: Configure Beads
+## Phase 1: Configure Tracker
 
-Check if beads is already initialized:
+### Default: bees
+
+Check for a **local** `.bees/` directory — not `bees list` or `bees ready`. Both walk up to an ancestor `.bees/` the same way `git` walks up to `.git/` (verified against bees 0.4.0: run from a subdirectory of an already-initialized parent, `bees create` silently writes into the *parent's* tracker with no local `.bees/` created and no error — the exact failure class that put stray test issues in the wrong tracker during this skill's own review). A directory-existence check is the only signal scoped to the current directory:
 
 ```bash
-bd status
+test -d .bees
 ```
 
-- If already initialized: report existing status and skip to Phase 2
-- If not initialized: run the appropriate init command based on flags:
+- If false: run `bees init` (creates `.bees/` in the current directory)
+- If true: report existing status and skip to Phase 2
+
+`bees init` also rebuilds the database from `.bees/issues.jsonl` if one is present, and always preserves an existing `.bees/config.json` — safe to run whenever `test -d .bees` is false, never needed when it's true.
+
+### `--beads`: beads
+
+Check for a **local** `.beads/` directory — not `bd status`, which has the same upward-walk behavior as bees (verified: `bd status` from a subdirectory of an already-initialized parent reports the parent's stats, with no local `.beads/` created and no error):
+
+```bash
+test -d .beads
+```
+
+- If true: report existing status and skip to Phase 2
+- If false: run the appropriate init command based on flags:
   - Default: `bd init`
   - `--stealth`: `bd init --local`
   - `--contributor`: `bd init --pull`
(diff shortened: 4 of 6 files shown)


vinnie357:

Reviewed as a separate agent briefed to defeat the change. Everything below was re-derived, not inherited from the PR body. Work done in a throwaway tracker and a scratchpad clone at `feat/setup-bees-default-and-enum-trap` (b330337); the shared working tree was never checked out.

### What I attacked and what survived

**Every bees invocation in `setup.md`, against `bees <cmd> --help` at 0.4.0** — all correct. `bees create` really does have `-t`, `-d`, `--json` and really has **no** `--labels` flag. `bees label add <issue-id> <label>` and `bees dep add <issue-id> <depends-on-id> [-t blocks|related|parent-child]` match help verbatim. `bees init` takes no flags, and "preserves an existing `.bees/config.json`" is quoted from its help text. I did not probe `bees upgrade`.

**The recipe end-to-end, run verbatim** — this was the highest-risk item, since `bees create --json | jq -r '.id'` breaks every bootstrapped project if the JSON shape is wrong. It is right: `bees create --json` emits a flat object `{"id":"...","title":...}`, so `jq -r '.id'` resolves. Ran the mise block as written: id captured, both `bees label add` calls landed (`labels: ["phase:tooling","skill:mise"]`), `bees dep add` landed as `blocks`. The mitigation text's `bees dep list <id> --json` → `.depends_on[].type` shape is real. The Phase 1 detection string is also exact — `bees list` outside a project prints `Error: not in a bees project (no .bees/ directory found)`.

**The enum claims — I reproduced four of four, not the two required, and found a fifth vector:**

| probe | result |
|---|---|
| `bees update <id> --status bogus` | exit 0, `Updated <id>`, stored `"bogus"`, then absent from `bees ready` |
| `bees create <t> -t garbage` | exit 0, `"issue_type":"garbage"` |
| `bees dep add <a> <b> -t nonsense` | exit 0, stored `"type":"nonsense"` |
| `bees update <id> -t alsogarbage` | exit 0, stored — **not currently named in the new section's opening sentence, though the value-set list does cover `update -t`** |

The documented value sets are quoted correctly: `-s/--status (open, in_progress, closed, deferred)` and `-t (task, bug, feature, epic, story)` are verbatim from `--help`, and `bees prime` line 20 really does write `-t task|bug|feature|epic|chore` — `chore` where help says `story`, both accepted. `parent` and `parent-child` are both accepted and stored as **distinct** literals, exactly as the table note now says. The section documents both detection (`bees show --json` / `dep list --json`) and the recovery tell (missing from `ready` with no close and no blocker), so it is not a warning-only half-fix. 378/500 lines, placement is fine.

**The cross-reference correction is right.** beads lives at `plugins/tools/beads`, marketplace name `beads`, skill dir `skills/beads` → `/beads:beads`. `/core:beads` would not resolve. Repo-wide `git grep '/core:beads'` returns hits **only** inside `.bees/issues.jsonl` — historical issue descriptions and close reasons, not live cross-references. No stale tokens in any skill, reference, agent, or command.

**The `--beads` path is intact.** `bd status`, `bd init`, `bd init --local` (`--stealth`), `bd init --pull` (`--contributor`) are all preserved unchanged, and the beads-only sub-flag framing is accurate — `bees init` genuinely has no mode flags.

**Repo hygiene.** All four version edits present and consistent (core 0.2.46, all-skills 0.4.92 in both plugin.json and marketplace.json); core and all-skills descriptions agree between the two files. `git diff origin/main HEAD -- test/quality-baseline.json` is **0 lines**. `mise test` green: 102 skills, `agents/commands/hooks surfaces: all clean`, duplicate blocks = 4 pre-existing groups with **no new group**. Diff scope is exactly the 6 claimed files — no unrequested edits.

---

### Findings

#### [blocking] 1. `plugins/core/agents/setup.md` is not registered — 120 of the 161 added lines never load

`plugins/core/.claude-plugin/plugin.json` carries an explicit `agents` array with exactly two entries:

```
./skills/bees/agents/bees-worker.md
./skills/bees/agents/bees-manager.md
```

`./agents/setup.md` is not among them, and there is no `commands` key at all. That asymmetry is decisive, and I confirmed it against my own session's loaded surfaces rather than by reasoning about it:

- The cached `core/0.2.41` ships `agents/setup.md` and `agents/gcms.md` on disk.
- My available **agent types** include `core:bees-manager` and `core:bees-worker` — the two that are listed — and include neither `core:setup` nor `core:gcms`.
- My available **skills** include all five of `/core:setup`, `/core:gcms`, `/core:research`, `/core:restraint-audit`, `/core:work`, matching the five files in `commands/`.

So commands auto-discover when the manifest omits the key; agents do **not** when an explicit array is present. Every other plugin in this marketplace with agents at the conventional `./agents/` path lists them explicitly — pm (6), qa (7), rust (2), slidev (3), elixir, wasm. `core` is the only plugin with conventional-path agent files left out of its own array.

Net effect after merge: `/core:setup` updates correctly and will advertise bees-by-default, then close with "Use the `setup` subagent to perform project bootstrapping" — and that subagent does not exist. The entire bees bootstrap recipe this PR was chartered to write is inert.

To be fair to the author: this is a **pre-existing** registration gap, not a regression — `/core:setup` was already dispatching to an unregistered agent before this PR, so nothing that worked stops working. I am still calling it blocking because claude-skills-154's requested outcome ("bootstrap bees as the default tracker for a new project") does not occur, and the fix is one line in a manifest this PR already edits for the version bump:

```json
"agents": [
  "./agents/setup.md",
  "./skills/bees/agents/bees-worker.md",
  "./skills/bees/agents/bees-manager.md"
]
```

`./agents/gcms.md` is orphaned the same way and is worth the same line, though it is outside this PR's scope. If the operator prefers to ship the content now and register in a follow-up, that is defensible — but the PR body should not read as though `/core:setup` bootstraps bees until it does.

#### [suggestion] 2. `jq` is a new, undeclared dependency on the default path

Four occurrences, all `--json | jq -r '.id'`. The beads path needs no `jq`. On a machine without it the pipeline yields an empty `ID*`, and the next line runs `bees label add "" "phase:discovery"` against an empty id rather than failing loudly. A one-line prerequisite note, or capturing the id and guarding it before the `label add` calls, closes this.

#### [suggestion] 3. bees resolves `.bees/` by walking **up** the tree, so Phase 1 mis-detects in a subdirectory

Verified: from `throwaway/subproject/nested`, `bees list` listed the *parent's* issues, and `bees create` wrote its issue into the **parent** tracker, leaving no `.bees/` in the nested directory at all.

Phase 1 says "if it errors with `not in a bees project`: run `bees init`; if already initialized: skip to Phase 2." Bootstrapping a new subdirectory project inside an existing bees repo (a monorepo package, or any child of a bees root) therefore takes the "already initialized" branch, skips init, and files all four bootstrap issues into the **parent** repo's tracker. The new project ends up with no tracker.

Not a regression — `bd status` also exits 0 from a nested directory, so the beads path has the same detection shape — but this PR promotes it to the default path. A `test -d .bees` check before falling back to `bees list`, or a follow-up bee, would cover it. Worth noting this is the same failure class that put `claude-skills-200`/`201` into the real tracker during this PR's own authoring.

#### [nit] 4. The beads sub-section lost its task headers

The diff removes `**Task 1 — …**` through `**Task 4 — …**` from the beads blocks, leaving four unlabeled consecutive code fences. Unrequested, and it makes the retained beads path harder to scan than before. Restoring the four bold lines is free.

#### [nit] 5. `bees init` re-run is slightly more than "idempotent"

Its help also says "Rebuilds the database from `issues.jsonl` if one is present." The doc's "safe to run even when already initialized" is true for `config.json`, but a re-init rebuilds `bees.db` from a possibly-stale `issues.jsonl`. Low impact given the agent only inits when `bees list` errors.

---

### Could not verify

Whether Claude Code would load `./agents/setup.md` under some manifest handling I cannot observe from inside a session. My evidence is behavioural — the two listed agents surface, the two unlisted files on disk do not — which I consider conclusive, but I did not read the loader.

### Known conflict (reported, not fixed)

PR #194 is open simultaneously. Colliding files are exactly three: `.claude-plugin/marketplace.json`, `.claude-plugin/plugin.json`, `plugins/core/.claude-plugin/plugin.json` — and in all three the overlap is version lines only. #193's other three files are untouched by #194, and #194's `plugins/core/skills/sources.md` is untouched by #193. Both PRs claim core **0.2.46**; #190, #191, #194 and #193 all claim all-skills **0.4.92**. A rebase needs version-number changes only — no content merge. Merge order is the operator's call.

vinnie357:
Re-triggering CI: the push for the review-fix commit (a81474d) never dispatched a workflow run (0 check-runs, status=pending) — likely a webhook drop under heavy concurrent CI load in this repo. Closing/reopening to re-fire the pull_request event; no code changes.

vinnie357:
## Gate 3 re-verification — APPROVE

Scoped to the blocker and the four follow-ups at `4722224` (`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`). Did not re-review the rest of the PR — the first pass covers it. Scratchpad clone only; the shared tree was never checked out, and every bees/bd probe ran in a throwaway directory.

### Blocker — CLOSED

`plugins/core/.claude-plugin/plugin.json` now lists four agents; `./agents/setup.md` and `./agents/gcms.md` are both present. I verified each of the four paths resolves to a real file at the PR head (`git cat-file -e` on all four: OK), and that the array is still well-formed — `mise test`'s marketplace and plugin validators both pass.

I also checked the condition that registration alone doesn't guarantee: the frontmatter `name` has to match the string the command dispatches on. Both do.

| file | frontmatter | command dispatches on |
|---|---|---|
| `agents/setup.md` | `name: setup` | "Use the `setup` subagent to perform project bootstrapping." |
| `agents/gcms.md` | `name: gcms` | "Use the gcms (git-commit-message-short) subagent…" |

Catching `gcms.md` independently was the right call — it was orphaned identically, and it is a real second fix rather than a courtesy addition.

**On confirming resolution rather than presence — I could not do this, and I want to be exact about it.** My session's plugin cache is pinned at `core/0.2.41`, and the only ways to load 0.2.47 would mutate the user's plugin install, which is out of bounds for a read-only reviewer. So what I have is: every static precondition met (registered, path resolves, name matches the dispatch string, manifest validates), plus the negative control from my first pass — in cache 0.2.41 both `setup.md` and `gcms.md` were on disk and absent from the agent-type list, while the two *listed* bees agents were present. That isolates listing as the operative variable and the fix flips exactly that variable. It is strong inferential evidence, not a live positive observation of `core:setup` resolving. If you want a hard confirmation, the cheap check is `/plugin` reinstall of core after merge, then look for `core:setup` in the agent-type list.

### Follow-up 2 — jq guard: reasoning holds, one wording fix

I tested whether jq can be present and still yield an empty id. **It can** — the guard does not cover the whole space:

```
BAD=$( (bees create 2>/dev/null || true) | jq -r '.id' )   # -> '' ; jq exits 0 on empty stdin
```

So a `bees create` that fails for any reason (locked db, constraint) leaves an empty id with jq installed. But the author's justification is that the residual is loud, and that checks out:

```
$ bees label add "" "phase:discovery"   ->  Error: SQLITE_CONSTRAINT   exit=1
$ bees dep add   "" "<real-id>"         ->  Error: SQLITE_CONSTRAINT   exit=1
$ bees label add "<real-id>" "ok:label" ->  Added label ...            exit=0   (control)
```

One upfront guard is genuinely sufficient. It converts the likely, systemic failure (jq absent on every machine that lacks it) into an upfront error, and the rare residual already fails loudly at the next command. Four repeated per-capture checks would buy nothing. Correct call, and the restrained one.

**[nit] The preamble mis-describes the failure.** It says an empty id means "every following `bees label add ""` / `bees dep add ""` then silently mistargets or errors". It never silently mistargets — it always errors, exit 1, as above. The guard's actual value is different and better: without it the four `bees create` calls still *succeed*, so you are left with four real but unlabeled issues and no dependency graph — loud, but partial. Suggest replacing "silently mistargets or errors" with something like "errors out, after the four issues have already been created unlabeled and unlinked".

This also corrects my own first-pass finding #2, where I wrote that it "runs against an empty id rather than failing loudly." That was wrong — it does fail loudly. The guard is still worth keeping, for the partial-state reason rather than the silence reason.

### Follow-up 3 — `test -d` is the right detector, and better than it claims

This is the strongest of the fixes. I probed the edge cases:

- **Empty `.bees/` stops the upward walk.** From a child directory of a tracker-owning parent, with only an empty `.bees/` present locally, `bees list` reports `No issues found.` rather than listing the parent's issues. So `test -d .bees` is not merely a reasonable proxy — it matches the exact signal bees itself uses to terminate the walk. The detector and the resolver agree by construction.
- **Empty `.bees/` is not a trap.** `test -d` returns true, init is skipped, and bees self-heals by creating the db on demand (`bees create` succeeded). The only consequence is a cosmetic one: the project prefix defaults to `bee-` rather than being derived from the directory name, because `bees init` never ran. Pathological input, benign outcome, not worth code.
- **The beads claim is true.** With one issue in the parent, `bd status` from the child reports `Total Issues: 1` — the parent's stats — and creates no local `.beads/`. The author's independent finding is confirmed.
- **Stealth/contributor still work.** The change only gates *whether* init runs; the flag→command mapping (`bd init` / `--local` / `--pull`) is untouched, and it is now reached on `test -d .beads` false, which is the correct precondition for all three modes.

### Follow-up 4 — confirmed

Eight `**Task N — …**` headers restored, four in each tracker section. The `bees init` phrasing now reads "also rebuilds the database from `.bees/issues.jsonl` if one is present, and always preserves an existing `.bees/config.json`" — which is `bees init --help` verbatim.

### New `/core:bees` directory-walk note — accurate

393 lines, under the 500 cap. Placed at line 105, between "No Enum Validation" (81) and "Dependency Management" (120), exactly as specified. I re-tested every factual claim in it and all hold: the upward walk, the parent-prefixed id, no local `.bees/` created, no error, and the beads parity via `bd status`.

**On the conflation risk you flagged — the shipped wording is clean.** The sentence reads "A `cd` that lands anywhere under a bees-tracked tree — **not just the tracker root** — followed by `bees create`, silently files into that tree's tracker." That clause distinguishes the two mechanisms instead of merging them: it generalizes over both the tracker-root case (the cwd-reset incident) and the subdirectory case (the walk-up), without asserting the walk-up caused the former. That is the correct relationship between the two, and the correction landed.

**[nit]** The parenthetical in `agents/setup.md` is looser: "the exact failure class that put stray test issues in the wrong tracker during this skill's own review." "Exact" adds precision the claim doesn't need, and the strays were created during *authoring*, not review. It doesn't conflate mechanisms, so it isn't wrong — just self-referential color a future reader can neither verify nor act on. Cutting it or trimming to "a failure class seen in practice" would lose nothing.

### Gates

- `mise test` — **exit 0**. Claude native validation, marketplace, all plugins, sources (28 files), disclosure lint (616 tracked files, clean) all pass.
- `git diff origin/main HEAD -- test/quality-baseline.json` — **0 lines**.
- Duplicate blocks — same 4 pre-existing groups, **no new group**.
- `gitleaks detect` — 26 commits, **no leaks found**.

### Versions — all four confirmed, no collision

Main is core 0.2.45 / all-skills 0.4.94. PR head: `plugins/core/.claude-plugin/plugin.json` 0.2.47, `.claude-plugin/plugin.json` 0.4.96, and both marketplace entries match (core 0.2.47, all-skills 0.4.96). #194 sits at core 0.2.46 / all-skills 0.4.95, one lane below on both. The two can land in either order without touching each other's numbers.

### Verdict

**APPROVE.** The blocker is closed to the limit of what is statically verifiable, both follow-up fixes I could attack behaviourally survived the attack, and the `test -d` change turned out to be better-founded than the PR describes. The three nits are wording only — none blocks merge.

PR #194, opened 2026-08-01
fix(skills): correct host.docker.internal claim and add elixir test-isolation section

vinnie357:
- Fix the `host.docker.internal` claim in the container skill (claude-skills-165). Verified via `strings` probes of the installed 1.0.0 `container`, `container-apiserver`, `container-network-vmnet`, and `container-runtime-linux` binaries: zero occurrences of `docker.internal`/`host-gateway`/`gateway.` (read-only inspection of files on disk). Verified the actual mechanism via `gh api repos/apple/container/contents/docs/how-to.md` (read-only GitHub read, upstream official docs): the real command is `container system dns create <domain-name> --localhost <ip>` (apple/container#346, PR #1078) — the upstream how-to guide's own example domain is `host.container.internal`, not `host.docker.internal`; the 0.9.0 release notes' phrasing was illustrative, not a hostname Apple Container resolves automatically. Also confirmed via `docs/technical-overview.md` and `docs/how-to.md` (both official, read-only) that the default network's gateway is `192.168.64.1/24`.
- Add a "Reaching host services from a container" section to `references/command-reference.md`, correct the three historical claim sites (`references/version-history.md`, `templates/0.9.0/commands.md`, `sources.md`) to cite the real mechanism, and add a pointer from `SKILL.md`'s Network Management section. The section qualifies the gateway-IP mechanism as reaching only non-loopback-bound host services (loopback-bound services, the common local-dev-server case, need the DNS-domain approach instead), clarifies that `--localhost <ipv4-address>` sets the FROM side of the PF redirect rather than the host destination (the TO side is always the host's `127.0.0.1`), and documents upstream's two stated limitations: creating a localhost domain disables Private Relay, and the PF rule does not survive a restart.
- Add a real "Test Isolation" section to the elixir `testing` skill (claude-skills-166) and repoint the broken citation in `references/os-subprocess-adapter.md:87`, which referenced a section that did not exist (verified: no isolation-focused heading anywhere in the 415-line `SKILL.md`; the only substantive related mention was an inline test-isolation reference inside the Setup paragraph at line 107, not a dedicated section). Also aligns the reference's repro guidance with "vary concurrency shape, not repetition count" so it doesn't contradict the new section, and reconciles the pre-existing Async Tests bullet ("don't use async when tests modify global state") with the new section's "async: false is a last resort" guidance.
- Bump `core` 0.2.45 → 0.2.46, `elixir` 0.1.20 → 0.1.21, `all-skills` 0.4.91 → 0.4.92 in both `plugin.json` and `marketplace.json`.

Not verified / explicitly out of scope: no claim in this PR rests on a live `container run`/`container network inspect` probe — everything cited above traces to either static binary inspection (`strings`) or read-only GitHub reads (`gh api`, `gh pr view`) of apple/container's own docs and PR history.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.96",
+      "version": "0.4.97",
       "category": "meta",
       "keywords": [
         "skills",
@@ -143,7 +143,7 @@
       "source": "./plugins/core",
       "strict": true,
       "description": "Essential development skills: Git, TDD, code review, gitleaks secret scanning, documentation, restraint, anti-fabrication, twelve-factor apps, the agent loop, bees issue tracking, mise, nushell, and Apple Container",
-      "version": "0.2.47",
+      "version": "0.2.48",
       "category": "development",
       "keywords": [
         "git",
@@ -197,7 +197,7 @@
       "source": "./plugins/languages/elixir",
       "strict": true,
       "description": "Elixir development skills: Phoenix, OTP, Ports, Ecto, Tidewave MCP, testing, configuration, style, and anti-patterns",
-      "version": "0.1.20",
+      "version": "0.1.21",
       "category": "language",
       "keywords": [
         "elixir",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.96",
+  "version": "0.4.97",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/core/.claude-plugin/plugin.json
+++ b/plugins/core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "core",
-  "version": "0.2.47",
+  "version": "0.2.48",
   "description": "Essential development skills: Git, TDD, code review, gitleaks secret scanning, documentation, restraint, anti-fabrication, twelve-factor apps, the agent loop, bees issue tracking, mise, nushell, and Apple Container",
   "author": {
     "name": "vinnie357"
--- a/plugins/core/skills/container/SKILL.md
+++ b/plugins/core/skills/container/SKILL.md
@@ -113,6 +113,8 @@ Create, inspect, and prune container networks (`container network ...`). Full fl
 
 Full IPv6 support (0.8.0+); host-only/isolated modes (0.9.0+, verify syntax with `--help`); `mtu` network attachment (0.11.0+).
 
+To reach host services from a container: the default network's gateway (`192.168.64.1`, no setup required) only reaches services bound to a non-loopback address. Services bound to `127.0.0.1` (the common local-dev-server case) are NOT reachable via the gateway — use `container system dns create <domain> --localhost <ip>` instead, which is a reachability feature, not just a naming convenience. See [references/command-reference.md](references/command-reference.md) "Reaching host services from a container".
+
 ## Volume Management
 
 Create, inspect, and prune persistent volumes (`container volume ...`). Full flag tables: [references/command-reference.md](references/command-reference.md). Mount with `-v name:/path[:ro]` on `container run`; see Common Workflows below for a persistent-data example.
--- a/plugins/core/skills/container/references/command-reference.md
+++ b/plugins/core/skills/container/references/command-reference.md
@@ -433,6 +433,25 @@ Remove unused networks.
 container network prune
 ```
 
+### Reaching host services from a container
+
+For a host service bound to `127.0.0.1` (loopback) — e.g. a local dev server started with `--bind 127.0.0.1` — the container's network gateway does NOT reach it; loopback-bound services are only visible on the host itself. Per the upstream ["Access a host service from a container"](https://github.com/apple/container/blob/main/docs/how-to.md) guide, create a local DNS domain with `--localhost <ipv4-address>` (admin privileges required) to reach it:
+
+```bash
+sudo container system dns create host.container.internal --localhost 203.0.113.113
+container run -it --rm alpine/curl curl http://host.container.internal:8000
+```
+
+`--localhost <ipv4-address>` is the address ASSIGNED TO the domain name inside the container (the redirect's FROM side), not the host address the traffic is forwarded to — the PF rule always redirects to the host's `127.0.0.1` regardless of the address chosen. Pick an address unlikely to collide with real traffic; upstream recommends the RFC5737 documentation ranges (`192.0.2.0/24`, `198.51.100.0/24`, `203.0.113.0/24`).
+
+Two upstream-documented limitations (`docs/how-to.md`, "Access a host service from a container"):
+- Creating a localhost domain disables Private Relay.
+- The PF rule is removed on restart — the domain must be re-created after every reboot.
+
+The upstream docs' own example domain is `host.container.internal`, not `host.docker.internal` — the 0.9.0 release notes' "`host.docker.internal`" phrasing (apple/container#346, PR #1078) was illustrative, not a hostname Apple Container resolves automatically or ships pre-configured. `strings` probes of the installed 1.0.0 `container`, `container-apiserver`, `container-network-vmnet`, and `container-runtime-linux` binaries contain zero literal occurrences of `docker.internal`, consistent with it never being hardcoded.
+
+For a host service bound to a non-loopback address (`0.0.0.0` or a specific LAN IP), the default network's gateway reaches it with no setup: apple/container's [technical overview](https://github.com/apple/container/blob/main/docs/technical-overview.md) and the `how-to.md` guide both document the default network as `192.168.64.1/24`, with `192.168.64.1` as the gateway address. This does NOT help for loopback-bound services — use the DNS-domain approach above for those.
+
 ## Volume Management
 
 ### `container volume create`
--- a/plugins/core/skills/container/references/version-history.md
+++ b/plugins/core/skills/container/references/version-history.md
@@ -10,7 +10,7 @@ Per-release feature history, the full migration checklist, and the dependency ma
 
 **0.8.0**: `--read-only` for run/create, architecture aliases (amd64/arm64/x86_64/aarch64), `network prune`, full IPv6, volume relative paths, env vars from named pipes, CVE-2026-20613 fix
 
-**0.9.0**: Resource limits (`--cpus`/`--memory`), `host.docker.internal`, host-only/isolated networks, `--dns` on build, `--force` on image delete, zstd compression, container prune improvements, enhanced image inspection, Kata 3.26.0 kernel
+**0.9.0**: Resource limits (`--cpus`/`--memory`), `container system dns create --localhost` for named local DNS domains that redirect to the host (apple/container#346 — per upstream `docs/how-to.md`, `host.docker.internal` was an illustrative example name, not a hostname resolved automatically; see `templates/0.9.0/commands.md`), host-only/isolated networks, `--dns` on build, `--force` on image delete, zstd compression, container prune improvements, enhanced image inspection, Kata 3.26.0 kernel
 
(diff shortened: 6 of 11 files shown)


vinnie357:

Reviewed at `1751fc9` in an isolated clone (no checkout in the shared tree). Ran **no** `container` subcommand of any kind; every container claim was re-derived from the apple/container repo, upstream docs, and read-only `strings` on the installed binaries.

The central thesis of claude-skills-165 is **correct and well-evidenced** — I attacked it hard and could not break it. But the new "Reaching host services from a container" subsection substitutes one under-evidenced usage example for another, which is the exact failure this bee exists to prevent.

---

### BLOCKER 1 — the `--localhost 127.0.0.1` example is wrong

`references/command-reference.md`:

```
container system dns create host.docker.internal --localhost 127.0.0.1   # requires sudo
```

`--localhost` is **not** the host address to forward to. It is the address assigned to the domain *inside the container* — the FROM side of a packet-filter redirect whose TO side is hardcoded:

- Installed 1.0.0 binary help string (`strings /usr/local/bin/container`): `Set the ip address to be redirected to localhost`
- `Sources/ContainerCommands/System/DNS/DNSCreate.swift`: `let to = try! IPAddress("127.0.0.1")` then `pf.createRedirectRule(from: from, to: to, domain: domainName)` — `from` is the `--localhost` value.

Upstream `docs/how-to.md:208-213` is explicit:

> Any IPv4 address can be used as `<ipv4-address>`, which will be assigned to the domain name in container. **Choose an IP address that is least likely to conflict with any networks or reserved IP addresses in your environment.** Reasonably safe address ranges include: the documentation ranges 192.0.2.0/24, 198.51.100.0/24, and 203.0.113.0/24. The 172.16.0.0/12 private range.

Upstream's worked example (`docs/how-to.md:225`):

```bash
sudo container system dns create host.container.internal --localhost 203.0.113.113
```

Passing `127.0.0.1` resolves the domain to the container's **own loopback**; traffic never leaves the VM, so a host PF rule cannot intercept it, and the rule installed is a degenerate `127.0.0.1 → 127.0.0.1`. It is also the single most conflict-prone value, against explicit upstream guidance.

Use `203.0.113.113` — RFC5737, which also matches this repo's own disclosure policy — and adopt upstream's `sudo <cmd>` prefix rather than a `# requires sudo` trailing comment.

### BLOCKER 2 — the gateway claim is stated unconditionally and is false for the common case

> **Automatic, no setup**: every container reaches the host through the network's gateway IP.

True only for host services bound to a **non-loopback** address. The entire `container system dns create --localhost` feature exists precisely because loopback-bound host services are unreachable — upstream's worked example binds `python3 -m http.server 8000 --bind 127.0.0.1` and then needs the DNS/PF mechanism to reach it. Presenting the gateway as the general-purpose answer sends readers down a dead end for the most common case (a dev server on 127.0.0.1). Add the binding caveat.

### BLOCKER 3 — upstream's two documented limitations are dropped

`docs/how-to.md:203-206` flags these as IMPORTANT:

> - Creating a localhost domain disables Private Relay.
> - The local domain packet filter rule is removed on a restart.

The new section instructs a `sudo` command that has a **privacy side effect** and **does not survive reboot**, and mentions neither. Both belong in the skill.

---

### MAJOR 4 — version collision with PR #193 (reported, not fixed)

Three files collide: `.claude-plugin/marketplace.json`, `.claude-plugin/plugin.json`, `plugins/core/.claude-plugin/plugin.json`. #193 makes **byte-identical** bumps — all-skills `0.4.91→0.4.92` and core `0.2.45→0.2.46`. There is no content overlap otherwise (#193 touches `core/skills/bees/SKILL.md`, `core/agents/setup.md`, `core/commands/setup.md`; #194 touches `core/skills/container/**`, `core/skills/sources.md`, `elixir/**`). **A rebase needs version-number changes only** — no semantic conflict. Whichever lands second must re-bump to all-skills `0.4.93` / core `0.2.47`, or its content ships under a version marketplace clients already consumed. `elixir 0.1.20→0.1.21` is unique to #194.

### MINOR 5 — the elixir skill now contradicts itself

`### Async Tests` (~line 325) says "Don't use async when tests: Modify global state / Access shared resources". The new `## Test Isolation` (line 351) says `async: false` is a LAST RESORT and the fix is to repair the isolation. Opposing guidance, ~20 lines apart, in one 426-line body. Reconcile.

### MINOR 6 — cite the reproducible source for the gateway IP

The skill sources `192.168.64.1` to a local `container network inspect` / `ip route` probe. Upstream documents it: `docs/how-to.md:123` (`"gateway": "192.168.64.1"`) and `docs/technical-overview.md:77` ("vmnet creates the container network using the CIDR address 192.168.64.1/24"). Cite those — a reader can check them; a probe on the author's machine they cannot.

### NIT 7
"SQLite/Postgres 'database busy'" attributes an SQLite-specific error string to Postgres.

### NIT 8
PR body: "only a single Key Principles bullet". The pre-existing mention is the Setup/teardown paragraph at line 107; the Key Principles bullet (line 406) is about Ecto sandbox "isolated database tests" — a different sense. Conclusion unaffected.

---

### What I attacked and FAILED to break

- **The `strings` claim.** Reproduced independently on all four binaries — two are not in `/usr/local/bin`, they live at `/usr/local/libexec/container/plugins/{container-network-vmnet,container-runtime-linux}/bin/`. `docker.internal` = 0, `host-gateway` = 0, `gateway.` = 0, and `192.168.64` = 0, on every one.
- **The 165 thesis.** apple/container PR #1078 is real, MERGED 2026-01-24, titled "Host connection", and is listed in the 0.9.0 release notes' What's Changed. It references issue #346, a real user request for `host.docker.internal`. The 0.9.0 highlights read "Access host services through hostnames like `host.docker.internal`" linking #346 — confirming it is a release-note *example*, and upstream's own docs use `host.container.internal` instead. **The version attribution to 0.9.0 is correct** and the rewrite of all three historical sites is accurate.
- **Command existence.** `container system dns create` / `delete (rm)` / `list (ls)` all exist — upstream `docs/command-reference.md:1488-1537`, and confirmed in the installed binary's own strings ("Create a local DNS domain for containers (must run as an administrator)", "Delete a local DNS domain", "List local DNS domains", "Manage local DNS domains"). The "requires admin privileges" claim matches the binary's `CommandConfiguration` abstract verbatim.
- **Disclosure lint.** `192.168.64.1` **is** on the reviewed allowlist (`test/validate-disclosure.nu:43`). Not a blocker. I did not take this on trust — I mutated the line to a non-allowlisted `10.42.7.9` and `mise run test:disclosure` failed as it should, so the lint genuinely guards this file.
- **Baseline.** `test/quality-baseline.json` is byte-identical between `main` and the branch. No new key, no count grew.
- **Gates.** `mise test` exits 0 on the branch in my clone. Skill scores unchanged: container 17/17 (391→393 lines), elixir testing 16/17 (415→426) — the `anti_fab` miss on elixir testing is **pre-existing on main**, not introduced here. Both under the 500-line cap; reference depth stays one level (zero `.md` links inside either touched reference).
- **Version bumps.** All six edits present and consistent across `plugin.json` and `marketplace.json`; descriptions and keywords unchanged.
- **166 correctness.** 415-line file and absence of any isolation heading both confirmed on `main`. `System.unique_integer([:positive])` and `mix test --max-cases 4` are valid; the 2-vCPU → `--max-cases 4` reasoning is right (ExUnit defaults `max_cases` to 2× schedulers online). No bang functions introduced. The `os-subprocess-adapter.md` repoint and the "vary the shape, not the repetition count" correction are both improvements.

**Verdict: REQUEST CHANGES.** Blockers 1-3 are all confined to the one new subsection in `references/command-reference.md` (plus a caveat sentence in `SKILL.md`); everything else in the PR is sound. The 166 half is clean and could land as-is.

vinnie357:
## Gate 3 re-verification at `2e05484` — REQUEST CHANGES (one line)

Scoped to the three blockers and the minors from [my review at `1751fc9`](https://github.com/vinnie357/claude-skills/pull/194#issuecomment-5152813626). Confirmed head `2e05484` (CLEAN/MERGEABLE); my independent fetch resolved to the same sha, so nothing is stranded on an orphaned commit. Again ran **no** `container` subcommand, no shared-tree checkout.

**Blockers 1 and 3 are fully closed. Blocker 2 is closed in the reference but still open in `SKILL.md`.**

---

### BLOCKER 1 — CLOSED, and the new statement is correct

Example is now `sudo container system dns create host.container.internal --localhost 203.0.113.113` — byte-identical to upstream's. I re-checked the new explanatory sentence against `DNSCreate.swift` rather than accepting that it merely differs from before:

```swift
try resolver.createDomain(name: domainName, localhost: localhostIP)
if let from = localhostIP {
    let to = try! IPAddress("127.0.0.1")
    try pf.createRedirectRule(from: from, to: to, domain: domainName)
}
```

- "the address ASSIGNED TO the domain name inside the container" — matches `createDomain(name:localhost:)` and upstream `how-to.md:208` ("will be assigned to the domain name in container"). ✓
- "(the redirect's FROM side)" — `localhostIP` is passed as `from`. ✓
- "the PF rule always redirects to the host's `127.0.0.1` regardless of the address chosen" — `to` is hardcoded. ✓

Accurate on all three counts.

### BLOCKER 3 — CLOSED

Both limitations present and accurate against `docs/how-to.md:203-206`. "Creating a localhost domain disables Private Relay." is verbatim; "The PF rule is removed on restart — the domain must be re-created after every reboot" correctly renders upstream's "The local domain packet filter rule is removed on a restart" plus its operational consequence.

### BLOCKER 2 — closed in `command-reference.md`, REOPENED in `SKILL.md:116`

The reference section is now right: it leads with the loopback case, states the gateway "does NOT reach it", and the gateway paragraph is properly qualified and routes back to the DNS approach. A reader who opens the reference is well served.

But `SKILL.md:116` is unchanged from the version I flagged, and now **contradicts the reference it points at**:

> To reach host services from a container, use the default network's gateway (`192.168.64.1`, no setup required) or `container system dns create` for a named hostname

This is the same unconditional framing Blocker 2 was about, sitting in the file that matters more. `SKILL.md` is Level 2 — it loads on **every** activation of the skill; `references/command-reference.md` is Level 3, loaded only when something sends the reader there. So the default reader gets "use the gateway, no setup required", with the DNS command mis-framed as being about *naming* rather than *reachability* — and hits exactly the dead end this blocker identified. The gotcha belongs in the body; that is what the body tier is for.

One line fixes it, e.g.:

> To reach a host service from a container: a service bound to `0.0.0.0` or a LAN IP is reachable at the default network's gateway (`192.168.64.1`, no setup); a loopback-bound (`127.0.0.1`) service is **not** and needs `container system dns create --localhost` — see [references/command-reference.md](references/command-reference.md) "Reaching host services from a container".

---

### Minors — all closed

- **Async/isolation contradiction**: resolved properly. The old "Don't use async when tests: Modify global state / Access shared resources" list is gone, replaced with "not automatically an `async: false` candidate — isolate the state first ... reach for `async: false` only when isolation genuinely is not possible, e.g. real unsandboxed database access." No longer contradicts `## Test Isolation`. ✓
- **`/Postgres` dropped**: now "SQLite 'database busy'". ✓
- **Gateway sourcing**: now cites upstream `technical-overview.md` and `how-to.md` instead of the local probe. Verified both document `192.168.64.1/24` and the `192.168.64.1` gateway. ✓
- **PR-body misattribution**: corrected. ✓

### The `172.16.0.0/12` removal — clean, and the right call

Fully removed; no dangling reference anywhere under `plugins/core/skills/container/`. The remaining text ("upstream recommends the RFC5737 documentation ranges") is still true — upstream does recommend those — it simply doesn't enumerate every option upstream lists, which costs the reader nothing. Agreed that widening a public-repo lint for one doc example would have been the wrong trade.

### Gates re-run at `2e05484`

- `mise test` → **exit 0**; all plugins valid.
- `mise run test:disclosure` → clean, 616 tracked files, self-test 9 bad + 11 good passing.
- `test/quality-baseline.json` → **byte-identical to main**. No new key, no count grew.
- **Versions**: main is all-skills `0.4.94` / core `0.2.45` / elixir `0.1.20`; branch is `0.4.95` / `0.2.46` / `0.1.21`, consistent across `plugin.json` and `marketplace.json`. The #193 collision is resolved.
- **Rebase lost nothing**: same 11 files, +48/−15. All three historical sites still corrected; the `os-subprocess-adapter.md` repoint intact.

---

**Verdict: REQUEST CHANGES**, narrowly — `SKILL.md:116` only. Everything else on this head is verified and merge-ready; fix that one sentence and this is an APPROVE with no further re-review needed on my side.

PR #195, opened 2026-08-01
refactor(claude-code): split claude-agents into references, sharpen benchmark-skills numbers

vinnie357:
- claude-skills-129: split claude-agents/SKILL.md (441 -> 156 lines) into references/patterns.md (writing style, best-practice and security worked examples) and references/plugin-config.md (plugin.json wiring, invocation, troubleshooting) — deleted content that only restated the frontmatter field table (empty tool-allowlist example, duplicated model-selection block, turn-limits YAML comment, comment-form frontmatter example)
- Fixed a fabricated metric in the Clear Instructions example ("minimum 80% code coverage" -> match the project's own coverage conventions)
- Sharpened claude-agents' description for activation: names frontmatter fields, tool-allowlist restriction, skill preloading, and invocation troubleshooting
- claude-skills-199: removed the benchmark-skills.md example scorecard's hardcoded `120/500`/`380/500` line-count denominators (same drift risk PR #189 fixed for the score denominator) — both line cap and check_count are now copied verbatim from `mise run test:skills-quality` output
- Rewrote benchmark-skills.md step 3 to drop numeric literals (1024/64/500) and two claims the validator doesn't check (imperative language, "Has references/"), correctly attributes source coverage to `test/validate-sources.nu`, and states explicitly that the narration is a reading aid, not the authoritative check list
- Bumped claude-code plugin version 0.2.44 -> 0.2.45 (plugin.json + marketplace.json) and all-skills 0.4.91 -> 0.4.92 (no skill paths changed, confirmed via `update-all-skills --dry-run`)

PR #196, opened 2026-08-01
feat(typescript): add typescript language plugin with four skills

vinnie357:
- Add `typescript` language plugin with four skills: `language` (strict mode, type system, utility types, declaration files, discriminated unions), `testing` (Vitest, Jest, Playwright, type-level testing), `linting` (ESLint + typescript-eslint, Biome, Prettier integration), `node` (async patterns, streams, ESM/CJS, package managers)
- Add `sources.toml`/`sources.md` with 14 tracked upstream sources, versions verified live against each tool's own docs and the npm registry on 2026-08-01
- Register the plugin in `.claude-plugin/marketplace.json` (category: language) and sync `all-skills` meta-plugin via `mise update-all-skills`

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -6,14 +6,14 @@
   },
   "metadata": {
     "description": "A curated collection of Claude skills for development workflows",
-    "version": "1.0.30"
+    "version": "1.0.31"
   },
   "plugins": [
     {
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.97",
+      "version": "0.4.98",
       "category": "meta",
       "keywords": [
         "skills",
@@ -463,6 +463,23 @@
         "tree-sitter"
       ],
       "license": "MIT"
+    },
+    {
+      "name": "typescript",
+      "source": "./plugins/languages/typescript",
+      "strict": true,
+      "description": "TypeScript development skills: language and type system, Vitest/Jest/Playwright testing, ESLint/typescript-eslint/Biome linting, and the Node.js runtime",
+      "version": "0.1.0",
+      "category": "language",
+      "keywords": [
+        "typescript",
+        "javascript",
+        "node",
+        "vitest",
+        "eslint",
+        "testing",
+        "linting"
+      ]
     },
     {
       "name": "ui",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.97",
+  "version": "0.4.98",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
@@ -110,6 +110,10 @@
     "./plugins/tools/slidev/skills/interactive",
     "./plugins/tools/tweag/skills/topiary",
     "./plugins/tools/tweag/skills/nickel",
+    "./plugins/languages/typescript/skills/language",
+    "./plugins/languages/typescript/skills/testing",
+    "./plugins/languages/typescript/skills/linting",
+    "./plugins/languages/typescript/skills/node",
     "./plugins/ui/skills/daisyui",
     "./plugins/wasm/skills/wasmtime",
     "./plugins/wasm/skills/wit",
--- a/plugins/languages/typescript/.claude-plugin/plugin.json
+++ b/plugins/languages/typescript/.claude-plugin/plugin.json
@@ -1,0 +1,17 @@
+{
+  "name": "typescript",
+  "version": "0.1.0",
+  "description": "TypeScript development skills: language and type system, Vitest/Jest/Playwright testing, ESLint/typescript-eslint/Biome linting, and the Node.js runtime",
+  "author": {
+    "name": "vinnie357"
+  },
+  "repository": "https://github.com/vinnie357/claude-skills",
+  "license": "MIT",
+  "keywords": ["typescript", "javascript", "node", "vitest", "eslint", "testing", "linting"],
+  "skills": [
+    "./skills/language",
+    "./skills/testing",
+    "./skills/linting",
+    "./skills/node"
+  ]
+}
--- a/plugins/languages/typescript/skills/language/SKILL.md
+++ b/plugins/languages/typescript/skills/language/SKILL.md
@@ -1,0 +1,237 @@
+---
+name: language
+description: Guide for core TypeScript language features and the type system. Use when writing or reviewing TypeScript types, enabling strict mode, choosing a utility type, designing generics or conditional types, writing declaration files, or modeling data with discriminated unions.
+license: MIT
+---
+
+# TypeScript Language
+
+Core TypeScript: strict mode, the type system, utility types, declaration files, and discriminated unions.
+
+## When to Use This Skill
+
+Activate when:
+- Configuring `tsconfig.json`, especially `strict` and its component flags
+- Designing generic functions, classes, or constraints
+- Writing conditional types, mapped types, or template literal types
+- Choosing between the built-in utility types (`Partial`, `Pick`, `Omit`, `Record`, ...)
+- Writing or consuming `.d.ts` declaration files
+- Modeling variant data with discriminated unions and exhaustiveness checks
+- Deciding between `interface` and `type`, or between `enum` and a union of literals
+
+## Strict Mode
+
+`strict: true` in `tsconfig.json` is the floor for new TypeScript code in this workspace, not an opt-in extra. It is a single flag that enables a family of component flags — turn it on first, then reach for a component flag only when you need something `strict` doesn't cover.
+
+```json
+{
+  "compilerOptions": {
+    "strict": true
+  }
+}
+```
+
+`strict` enables, among others:
+
+| Flag | What it catches |
+|---|---|
+| `noImplicitAny` | A variable or parameter would be inferred as `any` because no annotation is present |
+| `strictNullChecks` | `null`/`undefined` stop being implicitly assignable to every type |
+| `strictFunctionTypes` | Function parameter types are checked contravariantly instead of bivariantly |
+| `strictBindCallApply` | `.call`/`.bind`/`.apply` are checked against the underlying function's signature |
+| `strictPropertyInitialization` | A class field must be assigned in the constructor or have a definite-assignment marker |
+| `noImplicitThis` | `this` inside a function with unclear context is flagged instead of defaulting to `any` |
+| `alwaysStrict` | Every emitted file is parsed as, and emits, ECMAScript strict mode |
+| `useUnknownInCatchVariables` | `catch (e)` types `e` as `unknown`, not `any` — forces a type check before use |
+
+Two flags `strict` does **not** enable — turn them on explicitly when the codebase can bear the churn:
+
+- `exactOptionalPropertyTypes` — an optional property (`foo?: string`) rejects an explicit `undefined` assignment unless the type says `foo?: string | undefined`. Catches call sites that "clear" a field by writing `undefined` when the schema meant "omit it."
+- `noUncheckedIndexedAccess` — indexing into a `Record<string, T>` or array by a computed key returns `T | undefined` instead of `T`, since JavaScript doesn't guarantee the key exists.
+
+Both close real holes; both also touch a large fraction of existing index/optional-property call sites when retrofitted onto an established codebase. Turn them on for new projects by default; treat enabling them on an existing one as its own reviewed change, not a drive-by.
+
+## The Type System
+
+### Generics
+
+A generic function or class defers a type decision to its caller instead of committing to one type or widening to `any`:
+
+```typescript
+function firstOf<T>(items: T[]): T | undefined {
+  return items[0];
+}
+
+const n = firstOf([1, 2, 3]);        // T inferred as number
+const s = firstOf<string>(["a"]);    // T explicit
+```
+
+Constrain a type parameter with `extends` when the function body needs to rely on a property existing:
+
+```typescript
+interface HasLength {
+  length: number;
+}
+
+function longest<T extends HasLength>(a: T, b: T): T {
+  return a.length >= b.length ? a : b;
+}
+```
+
+`keyof` combined with a second type parameter constrained to it is the standard shape for a type-safe property getter:
+
+```typescript
+function get<T, K extends keyof T>(obj: T, key: K): T[K] {
+  return obj[key];
+}
+```
+
+Generic classes are generic only over their instance side — a static member cannot reference the class's own type parameter.
+
+### Conditional types and `infer`
+
+A conditional type picks between two branches based on an `extends` check evaluated at the type level:
+
+```typescript
+type IsString<T> = T extends string ? true : false;
+```
+
+`infer` introduces a type variable inside the `extends` clause and captures whatever the checked type matched there — this is how library-authored utility types extract a piece of a larger type:
+
+```typescript
+type Flatten<T> = T extends Array<infer Item> ? Item : T;
+type ElementOf<T extends readonly unknown[]> = T extends readonly (infer E)[] ? E : never;
+```
+
+Conditional types **distribute** over a union type parameter by default — `Cond<A | B>` evaluates as `Cond<A> | Cond<B>`. Wrap both sides in a tuple (`[T] extends [U] ? ... : ...`) to compare the union as one type instead of distributing across it. Distribution is usually what you want for a "filter this union" utility; non-distribution is what you want when the conditional is deciding between two shapes of the same aggregate type.
+
+### Mapped types and key remapping
+
+A mapped type transforms every property of an existing type using the same rule:
+
+```typescript
+type Readonly<T> = { readonly [K in keyof T]: T[K] };
+type Partial<T> = { [K in keyof T]?: T[K] };
+```
+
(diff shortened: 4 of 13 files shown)


vinnie357:
## Gate 3 — Adversarial Review: REQUEST CHANGES

Reviewed on an isolated clone (never checked out in the shared tree). Every number below was re-derived, not inherited from the PR body.

**One blocking defect: the linting skill's flagship ESLint config does not run.** Everything else survived a deliberate attempt to break it — including all 14 version numbers, which I resolved independently and found correct in all 14 cases.

---

### BLOCKING

**B1 — `linting/SKILL.md`: the ESLint setup example crashes ESLint.**

The config uses `js.recommended`. That property does not exist — `@eslint/js` exports only `{ meta, configs }`, so the value is `undefined`.

```
typeof js.recommended         = undefined
typeof js.configs.recommended = object
keys of js                    = [ 'meta', 'configs' ]
```

Running the skill's example verbatim against ESLint 10.8.0 (the exact version `sources.toml` records):

```
Oops! Something went wrong! :(
ESLint: 10.8.0
TypeError: Config (unnamed): Unexpected undefined config at user-defined index 0.
    at wrapConfigErrorWithDetails (.../lib/config/flat-config-array.js:63:9)
```

Changing the single token to `js.configs.recommended` makes it work and produce real diagnostics:

```
sample.ts
  1:7  error  Type number trivially inferred from a number literal, remove type annotation  @typescript-eslint/no-inferrable-types
✖ 1 problem (1 error, 0 warnings)
```

This is the first code block a reader copies out of the linting skill, and it fails closed with a stack trace rather than a hint. Fix: `js.configs.recommended`.

---

### MAJOR

**M2 — `linting/references/rule-catalog.md`, "Applying These Deliberately": both claims are wrong for half the rules.**

Read straight from the installed plugin's own rule metadata and preset definitions:

```
RULE                       requiresTypeChecking
no-floating-promises        true
no-misused-promises         true
no-explicit-any             false
consistent-type-imports     false

preset              consistent-type-imports
recommended                  no
strict                       no
stylistic                    no
strictTypeChecked            no
recommendedTypeChecked       no
stylisticTypeChecked         no
```

- *"These four rules require the type-aware parser configuration ... they inspect actual inferred types, not just syntax"* — false for `no-explicit-any` and `consistent-type-imports`. Both are purely syntactic. A reader is told they need `projectService` wiring (the expensive, slow option) to get two rules that work without it.
- *"Enable them as part of `strictTypeChecked` ... the preset already bundles them"* — `consistent-type-imports` is in **no** preset at all. A reader who follows this instruction believes the rule is on when it is silently off. That is the worse half of the error: it fails silently rather than loudly.

**M3 — `node/references/module-resolution.md`: dual-package-hazard mitigation #2 is inverted.**

The author flagged this section as its own synthesis rather than a quote. That flag is honest — current Node docs have reduced "Dual CommonJS/ES module packages" to a single line ("See the package examples repository for details"), so there is genuinely nothing to quote. The *diagnosis* half of the synthesis is correct and matches the historical Node text: two instances, two copies of module state, `instanceof` failing across them.

The *mitigation* half is not. The skill says:

> Make the CJS entry point a thin wrapper that re-exports the ESM implementation

Node's documented Approach #1 is the opposite direction:

> **Approach #1: Use an ES module wrapper** — Write the package in CommonJS or transpile ES module sources into CommonJS, and create an ES module wrapper file that defines the named exports. Using Conditional exports, the ES module wrapper is used for `import` and the CommonJS entry point for `require`.

The skill's direction is not impossible on current Node, but it is narrow. Tested on Node v24.18.0:

```
--- require(esm), no top-level await ---
OK bump= 1
--- require(esm) WITH top-level await ---
Error [ERR_REQUIRE_ASYNC_MODULE]: require() cannot be used on an ESM graph with top-level await.
  code: 'ERR_REQUIRE_ASYNC_MODULE'
```

So it works only on Node ≥ 22.12 (`require(esm)`), and breaks outright on any top-level await in the graph. The recommendation is also self-defeating on its own terms: the same sentence justifies the CJS entry as existing "purely for backward compatibility with older Node consumers" — precisely the Node versions where `require(esm)` does not exist. Reverse the direction to match Node's Approach #1, or drop the mitigation.

---

### MINOR

**m4 — `sources.toml`: the TypeScript entry's `check_method` contradicts its own `current_version`.**

`current_version = "7.0.2"` is **correct** — I confirmed it independently (npm `latest` dist-tag, published 2026-07-08, not deprecated). But the entry sets `check_method = "github-releases"` against `microsoft/TypeScript`, whose release feed and tag list both top out at `v6.0.3` (2026-04-16). The repo's own tooling therefore reports the latest as *older* than current, permanently:

```
│  # │   plugin   │  skill   │  source    │ current │ latest │
│  0 │ typescript │ language │ typescript │ 7.0.2   │ 6.0.3  │
```

The `notes` field documents the discrepancy but the machine-readable field was left pointing at the source that disagrees. Point `check_method` at the npm registry, or set it to `manual` like the other entries whose recorded value came from npm. Cosmetic sibling: Biome resolves to the raw tag `@biomejs/biome@2.5.6` rather than `2.5.6`, so it will never string-match either.

**m5 — `node/SKILL.md`: conditional-export ordering conflates authoring order with evaluation order.**

> Conditions are evaluated most-specific-first (`node-addons` → `node` → `import`/`require` → `default`)

Node's rule is object order, not specificity: *"Within the `exports` object, key order is significant. During condition matching, earlier entries have higher priority and take precedence over later entries. The general rule is that conditions **should be** from most specific to least specific in object order."* Specificity is the recommended way to *write* them; precedence is purely positional. Write `default` first and it wins regardless. The actionable advice ("always include a `default` fallback last") is correct.

**m6 — `language/SKILL.md`: the `tsc --init` claim is right but incomplete.**

All three named defaults check out. Running `tsc --init` on TypeScript 7.0.2:

```
"module": "nodenext",  "strict": true,  "moduleDetection": "force",   ← all three as claimed
"noUncheckedIndexedAccess": true,
"exactOptionalPropertyTypes": true,
```

Current `tsc --init` also turns on `noUncheckedIndexedAccess` and `exactOptionalPropertyTypes` — the exact two flags the Strict Mode section frames as manual opt-ins ("turn them on explicitly"). Nothing stated is false (neither is part of `strict`), but a reader who follows the skill's own advice to start from `tsc --init` already has them and will be confused by the framing. Worth one clause.

---

### What I attacked and could not break

- **All 14 tracked versions — every one correct.** Resolved independently against npm dist-tags, the GitHub releases API (not the rendered page), `nodejs.org/dist/index.json`, and the nodejs/Release `schedule.json`. Zero wrong, zero unverifiable.
  - The flagged TypeScript three-way conflict resolves in the author's favour: `7.0.2` is the real npm `latest`, published 2026-07-08, four months newer than the GitHub feed's `v6.0.3`. Recording it was the right call.
  - `node = "24"` as Active LTS is right: v24 "Krypton", LTS from 2025-10-28, maintenance not until 2026-10-20.
  - "Vitest requires Vite ≥6 and Node ≥20" is right: `engines.node = ^20.0.0 || ^22.0.0 || >=24.0.0`, `peerDependencies.vite = ^6.0.0 || ^7.0.0 || ^8.0.0`.
- **`duplicate_block` — zero new groups.** Ran the full corpus scan on both branches. Identical group-for-group: `1182fd06` (5 windows), `22d40d6a` (1), `28ccd32d` (7), `639413bb` (1) on main and on the branch. No typescript file appears in any group, and nothing landed against elixir/rust/zig.
- **`test/quality-baseline.json` — genuinely zero diff**, and the working tree is still clean after a full run, so the suite is not silently rewriting it.
- **Line counts exact as claimed**: 237 / 229 / 173 / 199 and 87 / 119 / 79 / 66. All well under 500.
- **17/17 × 4**, re-derived by running the validator, not read from the PR body. `mise test` exits 0. Total skills 102 → 106.
- **Frontmatter**: no `allowed-tools` anywhere; `name` matches directory in all four; descriptions 258–308 chars (cap 1024), all third-person with a `Use when` trigger.
- **`ref_depth` clean** — zero local links inside any of the four reference files, so no reference→reference chain. All four SKILL.md → reference links resolve to real blobs.
- **Registration correct**: marketplace `description` and `keywords` byte-identical to `plugin.json`; `typescript` placed correctly between `tweag` and `ui`; all four skill paths present in all-skills; metadata 1.0.30→1.0.31; all-skills 0.4.91→0.4.92. (The plugin array is not globally sorted, but that disorder — `agent-sandboxing`, `awman`, `claudio` — is pre-existing on main and this PR adds none.)
- **gitleaks clean** (native binary, 50 commits, no leaks). Disclosure lint passes as part of `mise test`.
- **Restraint**: four skills map to four genuinely distinct activation contexts, mirroring the elixir/rust precedent, and the references carry look-it-up material rather than padding. The skills cross-reference `/core:restraint`, `/core:tdd`, `/core:git`, `/core:mise` instead of restating them. The zero-new-duplicate-groups result is objective evidence there is no copied boilerplate. No filler finding.

---

### Merge-order collision (reported, not fixed)

Six PRs are open and this is real but **invisible to CI**.

| PR | all-skills version | skills array |
|---|---|---|
| #190 | 0.4.93 | 102 (unchanged) |
| #191 | 0.4.92 | 102 (unchanged) |
| #193 | 0.4.92 | 102 (unchanged) |
| #194 | 0.4.92 | 102 (unchanged) |
| #195 | 0.4.92 | 102 (unchanged) |
| **#196** | **0.4.92** | **106 (+4 typescript paths)** |

Simulated merging #195 then #196 onto main:

```
start:      0.4.91  skills=102
after #195: 0.4.92  skills=102
after #196: 0.4.92  skills=106     ← no conflict raised
```

Both sides rewrite `0.4.91` → `0.4.92`, so the 3-way merge sees base≠ours but ours==theirs and resolves silently. Two content changes, one version bump. Post-merge, `plugin.json` and the marketplace entry still agree (both 0.4.92) and `mise test` exits 0 — so **nothing in CI catches the under-bump**.

The skills array does *not* collide: #196 is the only PR touching it, and it merged cleanly to 106.

Practical consequence: only the version field needs manual attention, and only for the 0.4.92 cohort (#191/#193/#194/#195/#196). #190's 0.4.93 is correct only if exactly one 0.4.92 PR lands before it. Simplest handling is for whichever of these merges last to carry the final correct number — merge order is yours.

vinnie357:
## Gate 3 — Re-verification at `befbdae`: APPROVE

All six findings closed. Scoped to the fixes plus regression checks; the version table from the first pass was fully verified and nothing in it changed.

Rebase confirmed clean: merge-base of `befbdae` and `main` is `d1a79d9` (current main head). The typescript-side diff from the old head is exactly six files — `language/SKILL.md`, `linting/SKILL.md`, `rule-catalog.md`, `node/SKILL.md`, `module-resolution.md`, `sources.toml` — one per finding, nothing else.

---

### BLOCKING — closed by execution, not inspection

**B1 — `js.recommended`.** Extracted the corrected config verbatim from `SKILL.md` and ran it against ESLint 10.8.0 (`@eslint/js` 10.0.1 — that package versions independently of ESLint):

```
/…/sample.ts
  1:7  error  Type number trivially inferred from a number literal, remove type annotation  @typescript-eslint/no-inferrable-types
  2:7  error  'unused' is assigned a value but never used                                   @typescript-eslint/no-unused-vars

✖ 2 problems (2 errors, 0 warnings)
EXIT=1
```

It loads and lints. Exit 1 is lint findings, not the config-load `TypeError` from the first pass. Both layers are live — `no-inferrable-types` proves `stylistic` is applied, `no-unused-vars` proves the explicit `rules` block with `argsIgnorePattern` takes effect. The added note (`@eslint/js` exports `{ meta, configs }`; the bare form resolves to `undefined`) matches what I measured.

---

### MAJOR — both closed

**M1 — `rule-catalog.md`.** Re-read every classification from the installed `typescript-eslint` 8.65.0 (the exact version `sources.toml` records):

```
RULE                       requiresTypeChecking
no-floating-promises        true
no-misused-promises         true
no-explicit-any             false
consistent-type-imports     false

PRESET                  no-floating-p  no-misused-pr  no-explicit-a  consistent-ty
recommended             no             no             YES            no
strict                  no             no             YES            no
stylistic               no             no             no             no
recommendedTypeChecked  YES            YES            YES            no
strictTypeChecked       YES            YES            YES            no
stylisticTypeChecked    no             no             no             no
```

Every claim in the rewrite holds: both Promise rules type-aware and in `strictTypeChecked`; `no-explicit-any` syntax-only and live under plain `recommended`/`strict`; `consistent-type-imports` syntax-only and in **none** of the six presets. The new explicit-enable snippet also runs:

```
uses.ts
  1:1  error  All imports in the declaration are only used as types. Use `import type`  @typescript-eslint/consistent-type-imports
```

So the rule the old text said was already bundled is now correctly shown as something you must add by hand — and the shown snippet actually turns it on.

**M2 — dual-package mitigation.** Direction reversed to Node's Approach #1 (ESM entry wraps the CJS implementation). Verified the corrected direction works on Node v24.18.0 — one shared module instance, state intact across calls (`bump= 1 2`).

**Checked the closing sentence specifically, per the half-fix concern — it was fixed, not left behind.** The old conclusion ended at "not the default for an application"; the new one appends *"When that choice is made, keep the CJS file as the canonical implementation and the ESM file as the wrapper, not the other way around."* Body and conclusion now point the same way.

The supporting claims check out: ESM-imports-CJS is old and universal; `require(esm)` is version-gated and throws `ERR_REQUIRE_ASYNC_MODULE` on top-level await (reproduced in the first pass).

*Informational, not a finding:* Node's `modules.md` YAML history shows `require(esm)` came out from behind `--experimental-require-module` in **v23.0.0, v22.12.0, and v20.19.0** — so there is also a Node 20 backport the "≥22.12" phrasing doesn't mention. Node 20 hit end-of-life 2026-04-30, so "≥22.12" is correct for every supported release line and the argument is unaffected. Nothing to change.

---

### MINORS — all three closed

**m4 — `sources.toml`.** `check_method` is now `manual` with `github_repo` retained as informational. The author's claim that this is legal is backed by the validator's own test case:

```
test/validate-sources.nu:1344:  label: "github_repo present with check_method=manual (A-F1: legal, not an error)"
```

The `github_repo` requirement is conditional on `check_method == "github-releases"` (line 434) and is not enforced in reverse — there is even a comment noting zig legitimately carries the same combination. `nu test/validate-sources.nu` runs clean on this head: **29 files validated, 0 pending**. The rewritten `notes` states the reason plainly (GitHub's feed is four months stale for this repo, so an automated check would permanently misfire).

**m5 — conditional exports.** Now accurate. "Node evaluates conditions in the **order they're written in the object** — first matching key wins, not some engine-enforced specificity ranking" matches Node's text: *"Within the `exports` object, key order is significant. During condition matching, earlier entries have higher priority and take precedence over later entries."* The retained most-specific-first advice is correctly reframed as authoring convention, with the right reason (a broader condition written earlier shadows every narrower one after it), and the `"default"` rule follows from the same mechanism rather than being asserted separately.

**m6 — `tsc --init`.** Verified against TypeScript 7.0.2's actual generated `tsconfig.json`:

```
    // Stricter Typechecking Options
    "noUncheckedIndexedAccess": true,
    "exactOptionalPropertyTypes": true,
    …
    // Recommended Options
    "strict": true,
```

The claim is right to the letter — both flags are scaffolded in by default, under a heading named exactly "Stricter Typechecking Options", and `strict: true` genuinely sits in a separate "Recommended Options" block. The new/existing split (accept the scaffolded default for new projects; treat retrofitting as its own reviewed migration) is sound.

---

### No regression

- **Four skills still 17/17**: language 237, testing 229, linting 175, node 199 — all under 500. References 87 / 119 / **96** / 66. The two grown files stayed clean.
- **Duplicate-block corpus scan re-run on both sides of the rebase.** Against the *new* main `d1a79d9` (which has taken four merges since my first pass), groups are identical: `1182fd06`(5), `22d40d6a`(1), `28ccd32d`(7), `639413bb`(1). Same four on the branch. No typescript file in any group. Zero new groups still holds post-rebase and post-edit.
- `test/quality-baseline.json` zero diff vs main; working tree clean after a full run.
- `mise test` exits 0. Total skills 102 → 106.
- gitleaks clean (61 commits, no leaks).
- Registration intact through the rebase: marketplace `description`/`keywords` still byte-identical to `plugin.json`; `typescript` still between `tweag` and `ui`; all four skill paths in all-skills; typescript 0.1.0 agrees across both manifests; all four SKILL→reference links resolve.

### Version bumps checked by hand

Given `test:version-bumps` does not police the all-skills number, I compared against main directly:

| | main | PR | |
|---|---|---|---|
| all-skills `plugin.json` | 0.4.94 | **0.4.98** | upward |
| all-skills marketplace entry | 0.4.94 | **0.4.98** | agrees |
| marketplace `metadata` | 1.0.30 | **1.0.31** | upward |
| typescript | — | 0.1.0 / 0.1.0 | agrees |

Strictly upward, and #196 remains the only PR touching the all-skills `skills` array.

**One merge-order hazard worth naming**, since the same tooling gap applies to the other open PRs. Current all-skills numbers against main at 0.4.94:

```
#199 → 0.4.91  (below main)
#198 → 0.4.93  (below main)
#197 → 0.4.94  (equal to main — no bump)
#194 → 0.4.95
#196 → 0.4.98
#200 → 0.4.99
```

#196 is safe merged now. But if **#200 (0.4.99) lands first**, #196 would move all-skills *down* to 0.4.98 — cleanly and silently, exactly as before. Merging this cohort in ascending order avoids it; otherwise whichever lands after a higher number needs a re-bump. #197/#198/#199 have the equal-or-downward problem independently of this PR.

---

**APPROVE.** All six findings closed with executed evidence rather than inspection, no regressions, and remote CI green.

PR #197, opened 2026-08-01
fix(claude-code): stop sources:check/stale/report false-flagging unpinned versions as stale

vinnie357:
- Fix `classify-staleness` in `sources-check.nu`/`sources-stale.nu`/`sources-report.nu`: `current_version = "unknown"` (a documented schema placeholder, 152 entries in the corpus) was compared against `latest` and reported as false-positive drift on every hit — now classifies as `no-pin`, distinct from `unset` (field missing) and `yes` (real drift)
- Fix a second false-positive class: a `current_version` carrying a `v` prefix (e.g. `"v0.0.0"`) compared unequal to a v-stripped `latest` and reported drift for entries that are actually current — the comparison now strips a leading `v` from both sides (display values are untouched)
- Fix a markdown-table regression: the Stale Sources section printed each entry's `notes` as a sibling bullet line between table rows, which closes a GFM table after the first note and drops every later row to raw pipe text — `notes` now renders as a sanitized, truncated table cell (pipes escaped, newlines collapsed) instead
- Distinguish a dead/404 upstream releases feed (`no-releases`) and a rate-limited/unauthenticated GitHub response (`rate-limited`) from a generic `error`, via a pure `classify-fetch-error` function that matches the structural `(404)`/`(403)` pattern nushell's HTTP errors actually use — not a bare digit search, which could collide with byte-offset digits embedded in the same error text
- Surface each entry's `notes` field in all three reports' output, so a report can mark an upstream-documented default as intentionally pinned (e.g. act's `node:16-buster-slim`) without a new schema field
- Add `--self-test` to all three scripts (24-26 case-table cases each) and wire them into `mise run test:sources` (part of `mise test`/`ci`) — pure logic, no network, ~0.1s total; the live network calls themselves stay unwired, on-demand only
- Document the staleness state machine and a live-verified example table in `skill-update/SKILL.md`, with an explicit "Known gap" section on fenced-literal coverage (see below)

## Scope — what this delivers and what it doesn't (claude-skills-176)

This PR makes `sources:check`/`sources:stale`/`sources:report` trustworthy: a tracked `sources.toml` pin now gets an honest freshness signal, and the `notes` field surfaces per-entry rationale as the bee itself specified (no new schema field). It does **not** close the other half of claude-skills-176 — version literals inside markdown code fences (Docker image tags, CLI examples) only get a signal when they happen to name-match a tracked entry, and the bee's own named exemplar, act's `node:16-buster-slim`, has no entry and **cannot get one under the current schema**: `test/validate-sources.nu`'s `CHECK_METHODS` is `["github-releases" "hex-pm" "crates-io" "manual" "none"]`, with no Docker Hub method, while the bee explicitly named npm, GitHub releases, endoflife.date, and Docker Hub as the upstreams to query. Leaving 176 open against the schema-extension follow-up filed below.

## Verified live against the real corpus (2026-08-01, GITHUB_TOKEN-authenticated)

206 sources across 28 plugins: 25 up to date, **21 real drift** (e.g. `core/mise: v2026.3.15 → 2026.7.18`, `core/nushell: 0.113.1 → 0.114.1`), 135 manual, 25 correctly reported `no-pin` (previously false-positive drift), 0 `no-releases`/`rate-limited`/`error` live today. `agent-sandboxing/agent-substrate` (`v0.0.0` vs `0.0.0`) now correctly reports current instead of the v-prefix false positive.

## BEES REQUESTS

- File a follow-up against claude-skills-176: extend `sources.toml`'s `check_method` enum (`test/validate-sources.nu`'s `CHECK_METHODS`) to cover Docker Hub image tags and, per the bee's original proposal, npm registry and endoflife.date — needed before act's `node:16-buster-slim` (or any other fenced Docker/CLI literal) can be tracked at all.
- `sources-check.nu`/`sources-report.nu`/`sources-stale.nu` still triplicate ~150 lines of fetch/classify logic (kept the per-script self-contained house convention rather than introduce a shared-module import pattern this codebase doesn't otherwise use) — a future bug in one needs fixing three times unless someone dedupes.
- hex-pm/crates-io fetch failures still fall through to a bare `catch { "error" }` with no structural classification (unlike GitHub's 404/403 handling), and GitHub's secondary/undocumented-shape 429 rate limit is not distinguished from a generic error either.


vinnie357:
## Gate 3 — adversarial review: **REQUEST CHANGES**

Brief was to defeat the change, not approve it. The core fix is **sound and genuinely load-bearing** — I could not break `classify-staleness`, and every headline number in the PR body reproduced exactly on an independent live run. But the `notes` surfacing introduced a verified output regression in `sources-report.nu`, and one documentation claim does not hold.

All work done in a throwaway clone; the shared working tree was never checked out.

---

### What I attacked and what survived

| Attack | Result |
|---|---|
| Premise: do the 3 tasks really predate the bee and sit outside CI? | **Survived.** `sources-check.nu` introduced in #35 (2026-03-25); bee filed 2026-07-29. `mise test` depends on `test:claude, test:marketplace, test:plugins, test:skills-quality, test:core-list, test:disclosure, test:sources` — `test:sources` is the *schema* validator (`test/validate-sources.nu`), not the network check. Pivot justified. |
| Do the 152 and 25 figures contradict each other? | **Survived — reconciled exactly.** Independently re-derived: 206 entries across 28 plugin `sources.toml` (my first count of 207/153 was wrong — it swept in `skill-update/templates/sources.toml`, which the script's discovery correctly ignores). 152 unknown-pin, of which **127 sit on `check_method = "manual"`** and are never queried, and **25 on API methods** (22 github-releases + 1 crates-io + 2 hex-pm) = exactly the reported `no-pin` count. |
| Mutate `classify-staleness` back to naive `current != latest` | **Guard held.** 5 cases fail in each of the 3 scripts, exit 1. |
| Mutate `classify-fetch-error` to always-`"error"` | **Guard held.** 3 cases fail in each, exit 1. |
| Subtle mutant: drop only the `"rate limit"` phrase branch | **Guard held.** 1 case fails, exit 1. Not decorative. |
| Are the 3 duplicated copies actually identical? | **Yes.** `diff` on the extracted `classify-staleness` and `classify-fetch-error` bodies: byte-identical across all three files. Triplication is a real cost but not a divergence bug today. |
| Is `$err.debug` real, or is `classify-fetch-error` fed `""` and dead? | **Real.** nu 0.113.1 error record columns: `msg, debug, raw, rendered, json`. `msg` is only `"Network failure"` — the status digits live **only** in `debug`, so reading `debug` first is correct and necessary. Live 404 against the GitHub API → `no-releases`. Confirmed. |
| "0 dead feeds" — verify independently | **Conclusion holds.** All **36** distinct `github-releases` repos return HTTP 200 on `/releases/latest` (the PR's count of 32 is an undercount; the finding is unaffected). |
| Does the fix actually change behaviour? | **Yes, precisely.** Isolating the version axis (`--days 99999`, full corpus): **main flags 80, PR flags 55 — delta exactly 25**, matching the `no-pin` count to the unit. |
| Live run vs PR body summary | **Exact match**: 206 total / 135 manual / 25 no-pin / 24 current / 22 stale / 0 no-releases / 0 rate-limited / 0 error / 0 unset. Both cited drifts real (`mise v2026.3.15→2026.7.18`, `nushell 0.113.1→0.114.1`). |
| Gate 1 | `mise run test` **exit 0**. |
| gitleaks | `no leaks found`, 319 commits scanned. |
| Baseline | `test/quality-baseline.json` **unchanged**, 2 keys both sides. No new key. |
| New `TODO`/`FIXME`/`HACK` markers | None. |
| Terminal table shape at default width | **No regression** — main and PR truncate identically at 80 cols. |

---

### Findings

**F1 — [blocking] `sources-report.nu` now emits structurally broken markdown.**
Line 378, `print $"  - Notes: ($r.notes)"`, injects a non-table line *between* rows of a markdown table. In GFM a table ends at the first non-row line. Verified against GitHub's own renderer (`gh api markdown -f mode=gfm`): the table closes after **row 1**, and every subsequent row renders as raw pipe text swallowed into `<li>` elements. On main the Stale Sources table is contiguous and valid; this is a regression, and it hits the section an operator most needs. It looks fine in a terminal, which is why it slipped — but the script's stated job is "Generate markdown update report".
Fix options: make `notes` a table column (as `sources-check` does), move notes to a list *after* the table, or inline with `<br>` inside the cell.

**F2 — [medium] SKILL.md labels a table "verbatim" whose `notes` column is authored prose.**
SKILL.md:82 reads "Live run against this repo's own corpus, 2026-08-01 (verbatim, columns trimmed)". The seven non-notes fields in all three rows are verbatim-correct — I checked each against live output. The `notes` values shown (`pinned version is months behind upstream`, `pinned version is one minor behind upstream`, `current_version was never pinned — not drift`) appear nowhere in the tool's output. The real notes are 700–1100-char paragraphs starting `"mise uses CalVer, not semver…"`, `"Verified against 0.113.1 per sources.md…"`, `"sources.md Accessed 2025-11-15…"`. Across all 206 rows: median notes length 245 chars, 119 over 200, 38 over 500. "Trimmed" covers truncation, not rewriting — and the difference matters, because a reader calibrates expected table shape from this example.

**F3 — [medium] Latent span-digit collision in `classify-fetch-error`.**
The matched text is `$err.debug`, which embeds `Span[<byte-offset>..<byte-offset>]`. Those offsets are ~6-digit numbers that can contain `404`/`403`. Demonstrated: padding a script by 306 bytes makes a pure **DNS failure** classify as `no-releases`; 305 bytes makes it `rate-limited`. At the current byte offsets `sources-check.nu` correctly yields `error` for a DNS failure — I forced this with a byte-length-preserving invalid-host patch at the real call site, so it is **not firing today**. But the offset shifts on any edit above the call site or any nushell stdlib size change, so the misclassification is silently non-deterministic across edits and nu versions, and the wrong answer (`no-releases` = "this repo has no releases feed") reads as actionable. Suggest stripping `span: Span[...]` before matching, or matching `\b40[34]\b` against only the extracted `msg:` substring.
Related, lower priority: `check-hex-pm` and `check-crates-io` still use bare `catch { "error" }`, so a throttled crates.io/hex.pm never reports `rate-limited`. And GitHub's *secondary* limit is HTTP 429, which yields `Error is "429 Too Many Requests"` → `error`, since nushell's error text carries the status line, not the response body (so the `"rate limit"` phrase branch is unlikely to ever fire in practice — harmless, but it is the one branch the self-test exercises against a string nushell won't produce).

**F4 — [medium] A residual false positive of the same family, visible in the PR's own evidence.**
`check-github-releases` strips a leading `v` from the upstream tag but never normalises `current_version`. Live run: `agent-sandboxing/agent-substrate: v0.0.0 -> 0.0.0` classified **`stale: yes`** — and that entry's own `notes` says "…v0.0.0 published 2026-05-19 is still the latest release — **no drift**". So the `notes` column this PR added prints "no drift" beside a `yes`. 4 v-prefixed `current_version`s in the corpus; 1 is a pure false positive. The honest headline is **21 real drift + 1 residual false positive**, not 22. `classify-staleness` is precisely where the normalisation belongs and the new case table is precisely where a case would have caught it — per `/core:restraint`, "fix the shared function once".

**F5 — [medium] The three self-tests are unwired, and the stated reason doesn't hold.**
The new header comment justifies non-wiring with "CI has no credentials and must pass with zero external dependencies". But `--self-test` is by construction pure and network-free — that is the whole point of the extraction — and all three run in **0.116 s total**. The repo's actual established norm is the opposite: `mise test` already invokes `--self-test` on four `test/*.nu` validators (`validate-skills-quality`, `validate-core-list`, `validate-disclosure`, `validate-sources`). These three are the **only** `.nu` files under `plugins/` with a self-test at all (of 27 bundled skill scripts), so there is no "skill-bundled self-tests are uniformly unwired" norm to appeal to — this PR creates the first ones and leaves them unguarded. Nothing in CI stops `classify-staleness` regressing to the naive form. ~3 lines in the `test:sources` task closes it.

**F6 — [blocking, mechanical] Version bump is stale; branch conflicts with main.**
Main is now at claude-code **0.2.46** / all-skills **0.4.94** (#190, #192, #195 all merged after this branch point). PR #197 sets **0.2.45 / 0.4.92** — *behind* main. Test-merge yields exactly **3 conflicts, all version manifests**: `.claude-plugin/marketplace.json`, `.claude-plugin/plugin.json`, `plugins/tools/claude-code/.claude-plugin/plugin.json`. **The collision with #195 is version-fields-only** — #195 touched `skills/claude-agents/**` and `commands/benchmark-skills.md`, #197 touches `skills/skill-update/**`; no overlap outside the manifests. Rebase target: **0.2.47 / 0.4.95**.

**F7 — [claims] Bee 176 is partially resolved; the PR should say so.**
Bee 176's title is "No freshness signal for version literals inside **code fences**", and its body asks for a task querying "npm registry, GitHub releases, endoflife.date, Docker Hub".
*Delivered:* the `notes` mechanism — and the bee's comment explicitly said "that rationale belongs in the entry's `notes` field. Do not reserve a new schema field for it", which this PR follows exactly; plus a drift table that is now trustworthy enough to act on.
*Not delivered:* any linkage from a fenced version literal to a `sources.toml` entry. The bee's own named exemplar, `node:16-buster-slim`, is still at `plugins/tools/github/skills/act/references/setup-and-config.md:98`, is **not tracked in any `sources.toml`**, and **cannot be** — `test/validate-sources.nu:31` restricts `check_method` to `["github-releases" "hex-pm" "crates-io" "manual" "none"]`, with no npm, Docker Hub, or endoflife.date method. Fenced literals still get no freshness signal.
Recommend not closing 176 as fully resolved: close with an explicit scope note plus a follow-up for the fenced-literal↔`sources.toml` linkage and the missing check methods, or leave it open.

---

### Could not verify
- The `no-releases` and `rate-limited` states end-to-end against a *live* upstream — the author's own caveat is accurate, and I confirmed why: all 36 github-releases repos return 200, so neither state is reachable from the current corpus. Both are proven by pure self-test and, for `no-releases`, by my direct 404 probe against the GitHub API.
- Remote CI on the rebased result (the branch does not currently merge cleanly).

### What did NOT break
The central claim of this PR is correct and well-evidenced: the `unknown` false positive was real, the fix removes exactly the 25 affected entries and nothing else, the pure functions are genuinely tested, and the restraint pivot (harden three existing scripts rather than build a fourth) was the right call. F1 and F6 are what block; F2–F5 are fixable in place.

vinnie357:
## Gate 3 re-verification at `e1930e1` — **APPROVE**, pending one mechanical version bump

Scoped to the six findings; did not re-review `classify-staleness`'s core logic, already mutation-tested last round. Throwaway clone throughout; shared tree never checked out.

**Five of six fixed and independently verified. F6 is not done** — `all-skills` is now *below* main, not equal.

---

### F1 — broken markdown table: **FIXED** ✓

Re-ran my own method rather than trusting the report. Regenerated the core-plugin report and pushed it through `gh api markdown -f mode=gfm`:

```
tables: 4
  table0:  9 rows, cell-widths=[2]   (Summary)
  table1:  4 rows, cell-widths=[8]   (Stale Sources)
  table2: 22 rows, cell-widths=[8]   (All Tracked Sources)
  table3: 22 rows, cell-widths=[5]
li fallback present: False
raw pipe text leaked into body: False
non-table lines containing a pipe: 0
```

Four real `<table>` elements, **uniform cell width within every table** (so no row is silently split by an unescaped pipe), zero `<li>` fallback. The sibling-bullet is gone; `notes` is a genuine 8th column.

`sanitize-notes-cell` probed directly on 10 inputs — literal pipe, embedded newline, tab+CR, already-escaped `\|`, double-escaped `\\|`, 300-char note, truncation landing mid-escape-pair, empty, whitespace-only, trailing backslash. **All 10 render as intact 3-cell rows through GFM**, including the two I expected to break it: `\\|` collapses to a literal pipe *inside* the cell rather than splitting it, and a truncation that severs `\|` leaves `\…`, where the backslash escapes a non-delimiter. Corpus exposure measured (via `tomllib`, not grep): 1 note with a literal pipe, **0** with escaped pipes, **0** with newlines, **0** ending in a backslash, max length 1122 vs the 80-char cell budget.

### F3 — span-digit collision: **FIXED, and the stronger claim holds** ✓

This was the one to attack hardest, since requiring parens is a bet on nushell's message shape. I probed real nushell 0.113.1 against a live server for five statuses rather than trusting the self-test fixture:

| status | `$err.debug` message | parens? |
|---|---|---|
| 403 | `Access forbidden (403) to "…"` | **yes** |
| 404 | `Requested file not found (404): "…"` | **yes** |
| 401 | `Cannot make request to "…". Error is "401 Unauthorized"` | no |
| 429 | `… Error is "429 Too Many Requests"` | no |
| 500 | `… Error is "500 Internal Server Error"` | no |

nushell special-cases **exactly 403 and 404** with parenthesised messages; everything else takes the generic `Error is "NNN Text"` form. So the assumption is not a lucky fixture — it holds for precisely the two statuses this function classifies.

I then re-ran my original collision attack against the new code, targeting the exact colliding offsets:

```
PAD=301  Span[160403..160411]  -> error   [span literally contains 403]
PAD=302  Span[160404..160412]  -> error   [span literally contains 404]
```

Plus a 51-offset brute sweep across pads 0–600: **no misclassification at any offset**. Under the old bare-digit matching these same shapes produced `rate-limited` / `no-releases`.

**Failure direction is safe.** If nushell ever drops the parens, `classify-fetch-error` degrades to `error` — "fetch failed for another reason", which is honest — rather than fabricating a confident `no-releases`. The old behaviour failed in the opposite, unsafe direction. This is a better fix than the one I asked for.

### F5 — self-tests wired: **FIXED** ✓

All three suites now run inside `test:sources`, and the case counts grew (20 / 26 / 25, up from 16 / 16 / 21). Verified propagation by breaking a case rather than reading the wiring:

```
mise run test:sources           -> exit 1, "❌ classify-staleness: … want DELIBERATELY-WRONG, got no-pin"
mise run test                   -> exit 1, same message surfaced under [test:sources]
(restored)  mise run test:sources -> exit 0
```

### F4 — v-prefix false positive: **FIXED, no false negatives** ✓

Live full-corpus run reproduces the claimed figures exactly: **206 total / 25 up to date / 21 stale / 135 manual / 25 no-pin / 0 no-releases / 0 rate-limited / 0 errors**. `agent-substrate` (`v0.0.0` vs `0.0.0`) now reports current.

Probed the v-strip for the false-negative risk across 10 cases — `v1.0.0→1.0.1`, `v1.0.0→v1.0.1`, `1.0.0→v1.0.1` all still flag `yes`; the three real corpus v-prefixed drifts (mise CalVer, agent-sandbox, whamm) still flag. **0 failures.** Display values are untouched, so the report still shows the real `v`-prefixed string.

*One observation, not a defect:* an intermediate run showed `24 no-pin / 1 check error`; a clean re-run gave `25 / 0`. That was a transient crates.io/hex throttle, and it confirms live that non-GitHub fetch failures land in generic `error` — supporting the bees request already filed below.

### F2 — the "verbatim" claim: **FIXED** ✓

Now reads "`plugin`/`skill`/`source`/`current`/`latest`/`stale`/`priority` are copied verbatim from the run; `notes` is a short paraphrase, not a literal quote (real `notes` values run 700–1100+ characters)", and the column header is `notes (paraphrased)`. Accurate — I measured max notes length at 1122.

### 176 disposition: **accurate, and I re-derived every new claim** ✓

The PR no longer implies 176 closes. New factual claims in SKILL.md all check out:

- "**six** entries across **five** plugins document a live 404, all on `check_method = "manual"`" — **correct**. My first grep returned 8 across 6 plugins, but reading the note text shows the 2 extras are legitimately excluded: `allium-tools` explicitly cross-references *another* entry's 404, and `proxmox/internal-infra-repo`'s 404s are press-release URLs, not a releases feed. The remaining 6 (`example-skills-repository`, `skills-cookbook`, `juxt-allium`, `json-schema-to-nickel`, `actions-toolkit`, `runex`) are all genuine releases-feed 404s, all on `manual`, across 5 distinct plugins.
- gitleaks `rev: v8.30.1` at `core/skills/security/SKILL.md:287` with a `github-releases` entry whose `current_version` is `"unknown"` — **confirmed**.
- `CHECK_METHODS = ["github-releases" "hex-pm" "crates-io" "manual" "none"]`, no Docker Hub method, so `node:16-buster-slim` is **unexpressible** rather than merely untracked — **confirmed**, and the wording says exactly that.

`## BEES REQUESTS` files the schema extension plus both residual gaps I raised (triplication; hex/crates bare `catch` and the undistinguished 429). Triplicated `classify-staleness` / `classify-fetch-error` / are byte-identical across all three scripts — no divergence today.

### F6 — versions: **NOT DONE** ✗

Main moved again during this pass (now `cfd02dc`, #198 landed). Current state:

| component | main (`cfd02dc`) | PR (`e1930e1`) | |
|---|---|---|---|
| claude-code plugin + marketplace | 0.2.46 | **0.2.47** | ok, above |
| all-skills plugin + marketplace | **0.4.99** | **0.4.98** | **below main** |

`all-skills` needs **0.4.100** (`sort -V` confirms `0.4.100 > 0.4.99`; no 3-digit-patch precedent in this repo yet, so `0.5.0` is the alternative if you'd rather roll the minor — your call). Local test-merge against `cfd02dc` conflicts on `.claude-plugin/marketplace.json` and `.claude-plugin/plugin.json` only. Note the `gh` API still reports `MERGEABLE / CLEAN` — that field is stale relative to the newest main; trust the local merge.

### Gates
`mise run test` → **exit 0**, with all three new suites visible in the run. `gitleaks` → **no leaks found**, 328 commits scanned.

---

**Verdict: APPROVE.** F1 and F3 are genuinely well-built — F3 in particular replaces a heuristic with a structural invariant I was able to verify against real nushell, and the failure direction is now conservative. The only outstanding item is the mechanical `all-skills` → `0.4.100` bump; no code objection remains.

vinnie357:
## Gate 3 verdict re-pinned to `2fae593` — **APPROVE, no reservations**

My detailed re-verification ([previous comment](https://github.com/vinnie357/claude-skills/pull/197#issuecomment-5153131314)) was performed against `e1930e1`. The head has since moved to `2fae593` (merge of main `cfd02dc`). Rather than assume a merge is content-neutral, I compared blob SHAs:

```
IDENTICAL  scripts/sources-check.nu
IDENTICAL  scripts/sources-stale.nu
IDENTICAL  scripts/sources-report.nu
IDENTICAL  skill-update/SKILL.md
IDENTICAL  mise.toml
```

All five substantive files are byte-identical between `e1930e1` and `2fae593`, so the F1–F5 evidence carries over unchanged — the merge touched only the version manifests.

### F6 — now resolved

| location | main `cfd02dc` | PR `2fae593` | |
|---|---|---|---|
| `claude-code/.claude-plugin/plugin.json` | 0.2.46 | **0.2.47** | above |
| marketplace → claude-code | 0.2.46 | **0.2.47** | above |
| `.claude-plugin/plugin.json` (all-skills) | 0.4.99 | **0.4.100** | above |
| marketplace → all-skills | 0.4.99 | **0.4.100** | above |

All four locations consistent and above main (`sort -V`). Local test-merge against `cfd02dc`: **clean, exit 0, zero conflicts** — and the `gh` API now agrees (`MERGEABLE / CLEAN`), unlike last round when it was reporting stale.

### Gates re-run at `2fae593`
- `mise run test` → **exit 0**, zero failures, all three new suites present (`sources-check 13+7`, `sources-report 26`, `sources-stale 25`).
- `gitleaks` → **no leaks found**, 328 commits scanned.

### Verdict
**APPROVE.** The five substantive findings were verified fixed at `e1930e1` and the code is byte-identical here; the version item is now correct. Nothing outstanding from Gate 3.

Carried forward as follow-ups, already captured in the PR's own `## BEES REQUESTS` (not merge blockers): the `check_method` schema extension needed before claude-skills-176's fenced-literal half can close; the ~150 lines triplicated across the three scripts; and hex-pm/crates-io fetch failures still falling through to a bare `catch { "error" }` with GitHub's 429 undistinguished — for which I have live evidence, having caught a transient crates.io throttle surface as `1 check error` mid-review.

PR #198, opened 2026-08-01
feat(core): make subset satellites drift-checkable in validate-core-list

vinnie357:
- Add an `expected: [...]` entry shape to `SATELLITES` in `test/validate-core-list.nu` (doc item 3c) so a satellite that deliberately subsets the mandatory core list is compared against its own declared subset instead of the full canonical set, in both directions (missing / extra)
- Register `references/validator.md` (2 names) and `references/fix-agent.md` (5 names) as subset satellites, closing the gap where a tier dropping — or unexpectedly gaining — a skill from its declared subset passed CI silently
- Reformat `validator.md`'s Phase 1 step 1 to a fenced load-list line (matching `fix-agent.md`'s existing shape) so the grammar-based run detection can parse it; no semantic change
- Validate `expected` names against the canonical set as a per-satellite finding, not a global abort — an earlier version of this check used an immediate `exit 1` that masked every other satellite's real drift in the same run; fixed and verified with a planted-double-failure test (see below)
- Add `NEVER_OMISSIBLE` enforcement (extracted into a standalone `expected-preflight-errors` function): a subset satellite's declared `expected` list is checked against `/core:anti-fabrication` independently of the file's actual load list, so deleting the name from BOTH the file and its `expected` declaration at once can no longer pass by agreeing with itself
- Update the now-false "what a tier omits is not checked" sentence in `agent-loop/SKILL.md`, plus the matching doc comment, satellite count, and tier-reference count in `test/validate-core-list.nu`
- Add 7 self-test cases for the subset shape and the never-omissible invariant (exact match, missing/lower-bound, extra/upper-bound, plus 4 direct cases for `expected-preflight-errors`) — `test:core-list` self-test now 39 cases, up from 32
- Bump `core` 0.2.47 → 0.2.48 and `all-skills` 0.4.96 → 0.4.97 (one lane past #193/#194 to avoid version collision)

## Reproduction (before this change)

Deleting `/core:anti-fabrication` from `validator.md`'s load list — the skill the tier's own text calls never-omissible — left both `test/validate-core-list.nu` and `test/validate-skills-quality.nu` at exit 0. Same result deleting a name from `fix-agent.md`'s subset. Restored before implementing.

## Verification (after this change)

- Deleting `/core:anti-fabrication` from `validator.md`'s subset → exit 1 (caught via the anchor/adjacency path, since the single remaining name falls under the run-detection's 2-name-minimum threshold)
- Deleting a name from `fix-agent.md`'s 5-name subset → exit 1, reported as `missing from load list — /core:anti-fabrication`
- Adding an unexpected `/core:security` to `fix-agent.md`'s subset → exit 1, reported as `extra in load list — /core:security` (the upper-bound case)
- Planted a real drift in `team-leader.md` (pointer instruction corrupted) simultaneously with a malformed `SATELLITES` entry (typo'd `expected` name) in the same run: an earlier version of the sanity check `exit 1`'d immediately on the malformed entry and never reported the `team-leader.md` drift at all. Fixed by folding the `expected`-name validation into the per-satellite loop as a regular finding; re-ran and both failures now surface together in one report.
- Deleting `/core:anti-fabrication` from BOTH `fix-agent.md`'s load list AND its `expected: [...]` declaration at once → was exit 0 before the `NEVER_OMISSIBLE` check (the two agreed with each other, so the ordinary missing/extra comparison never fired). Now exit 1, `expected subset drops never-omissible name(s) — /core:anti-fabrication`.
- All mutations restored; working tree clean before each commit.

## Gate 1

`mise test` — exit 0 (includes `test:core-list` self-test, 39 cases). `mise run test:version-bumps origin/main` — exit 0, `core: 0.2.47 → 0.2.48`. No diff to `test/quality-baseline.json`. Rebased onto origin/main (`bc72282`, includes #193) — `gh pr view --json mergeable` reports `MERGEABLE`.

## Interaction with claude-skills-151

`test/validate-skills-quality.nu`'s `core-list-satellites` helper parses `test/validate-core-list.nu` with a regex matching `path: "..."` anywhere in the file to build the dupe-exemption 3 file set. This PR adds exactly two new `path: "..."` entries — the `validator.md` and `fix-agent.md` `SATELLITES` registrations themselves — no stray strings elsewhere in the diff. This correctly *expands* the dupe-exemption to cover the two files now legitimately owned by `validate-core-list.nu`; verified `test/validate-skills-quality.nu`'s duplicate-block report is unchanged before/after (same 4 groups, same members) since neither file was previously part of any duplicate-block finding.

## Gate 2

`mise exec gitleaks@latest -- gitleaks detect --source="." -v` on the branch commits → "no leaks found" (used the mise-managed binary directly, not the `container run`-wrapping shell alias `gitleaks` resolves to on this host).

## BEES REQUESTS

- close claude-skills-152 on merge, referencing this PR
- file a follow-up: `mise run test:version-bumps` only checks that a modified plugin's version increased on the diff — it does not independently verify the new value against origin/main's CURRENT value for a plugin like `all-skills` that aggregates others. An earlier push on this PR set `all-skills` to 0.4.93 while main was already at 0.4.94 (a regression, not just staleness) and `test:version-bumps` reported green throughout.


vinnie357:

**Verdict: REQUEST CHANGES** — on merge mechanics (F1, F2), not on the core design. The subsetting check itself is sound, its self-tests are real, and the masking fix the author reports is genuine and reproducible. One design gap (F3) is worth closing while rebasing, because it is the literal reason the bee exists.

All work done in a scratchpad clone; the shared tree was never checked out. Every mutation restored (`git status` clean after each round).

---

### What I attacked, and what survived

**Re-ran the author's three mutations — all three genuinely go red.**

| Mutation | Exit | Reported as |
|---|---|---|
| Delete `/core:anti-fabrication` from `validator.md` (2→1) | 1 | `no load list directly after the anchor` |
| Delete `/core:tdd` from `fix-agent.md` (5→4) | 1 | `missing from load list — /core:tdd` |
| Add `/core:security` to `fix-agent.md` (5→6) | 1 | `extra in load list — /core:security` |

**Went beyond, on the entry shape:**

| Probe | Exit | Result |
|---|---|---|
| `expected` names `/elixir:testing` (real skill, wrong plugin) | 1 | caught, named, run continues |
| `expected` carries typo `/core:anti-fabricaton` | 1 | caught, named, run continues |
| `expected: []` on a registered satellite | 1 | falls back to full canonical — sensible default, but a silent footgun if an author means "expects nothing" |
| Duplicate `/core:git` inside a run | 0 | invisible (set-membership compare). Pre-existing for all satellites, not a regression; cosmetic. |
| Irregular whitespace `/core:a ,  /core:b,/core:c` | 0 | parses correctly — robust, not a hole |

**The 2-name-minimum weak spot — probed hard, it holds.** The author's disclosure is accurate but understates why it is safe. Constructing the 3-name case the brief asked for:

- 3-name subset → 2 names: **caught properly**, `missing from load list — /core:anti-fabrication`. The mechanism works for its own reasons whenever the run stays ≥2 names.
- 2-name subset → 1 name: caught by adjacency, with a different message. I tried to find a configuration where the 1-name residue escapes: a lone name is not a run, so no run begins at the adjacent line and adjacency fires; adding a second consecutive name line makes it a ≥2-name run, which is then compared against `expected` and caught (verified — reported both `missing` and `extra`). A blank line before a decoy run does not help either, since adjacency inspects the first non-blank line. **Every degradation path either trips adjacency or forms a detectable run.** I could not construct an escape.
- One genuinely uncaught shape, pre-existing and documented: a *correct* 2-name run followed by a stray 1-name line passes (exit 0). Not introduced here.

**Masking bug — reproduced, and the fix is real.** Reverting the check to a pre-loop `exit 1` and planting a real `team-leader.md` pointer drift alongside a malformed `SATELLITES` entry: the reverted shape reported **only** the malformed entry and never surfaced the `team-leader.md` drift. The shipped version reports **both** in one run. The author's account is accurate.

**Self-tests are not decorative.** 35 cases confirmed. Four mutations to the shipped comparison logic — `target` forced to `$canonical`, `extra` direction deleted, `missing` direction deleted, empty-check inverted — turn the suite **red every time**, and two name the exact new cases (`subset_extra_name_reported`, `subset_missing_name_reported`). This guard can fail under its own target change.

**Counts re-derived, not inherited.** `grep -c 'path: "'` on each tree: main = **8** satellites / **4** under `references/`; branch = **10** / **6**. Both the old and the new numbers are correct for their respective trees.

**Other checks:** `mise test` green on the branch. `gitleaks detect` — 275 commits, no leaks. `test/quality-baseline.json` diff empty. Duplicate-block report byte-identical to main (same 4 groups, same members) — the PR body's claim verified independently.

---

### Findings

**F1 — BLOCKING: the PR does not merge.** `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. The branch is based on `05b4174`; main is at `d1a79d9` (#190 and #195 landed since). Needs a rebase.

**F2 — BLOCKING: `all-skills` version is a regression.** Main is at **0.4.94**; this PR sets **0.4.93**, i.e. *below* main. Correct targets depend on merge order — current observed values, which differ from the brief (193 has moved):

| If 198 merges… | core | all-skills |
|---|---|---|
| next (before 193/194) | 0.2.46 ✓ as-is | **0.4.95** (currently 0.4.93) |
| after 194 (wants core 0.2.46, all-skills 0.4.95) | 0.2.47 | 0.4.96 |
| after 193 (wants core 0.2.47, all-skills 0.4.96) | 0.2.48 | 0.4.97 |

Worth noting for the tooling owner: `mise run test:version-bumps gh/main` reports **green** on this branch, listing only `core: 0.2.45 → 0.2.46`. It does not flag the downward `all-skills` move, so Gate 1 evidence structurally cannot catch this class. Pre-existing gap, not this PR's defect.

**F3 — HIGH (design, should close here): the never-omissible invariant on `/core:anti-fabrication` is now unenforced.** `SKILL.md` states "`/core:anti-fabrication` is never omissible." Nothing checks that an `expected` list contains it. Verified:

> Deleting `/core:anti-fabrication` from **both** `fix-agent.md` **and** its `expected: [...]` array → **exit 0, green.**

This is a strict improvement over the pre-PR state — the edit now requires touching a test file to stay green, rather than passing on a docs-only change — so the hole is *narrowed*, not merely relocated. But it is the exact invariant claude-skills-125 broke, and the guard is ~3 lines next to the existing `unknown_expected` check:

```nu
if ("/core:anti-fabrication" not-in $expected) { ...finding... }
```

Not a false claim anywhere in the text (see F5), so this is a gap to close, not a misstatement. Recommend closing it in the rebase rather than filing a follow-up, since the code is already open on exactly this line.

**F4 — LOW (informational): intra-file masking remains.** The `continue` after an `unknown_expected` finding skips `check-satellite` for that satellite entirely. Planting a bad `expected` name *and* breaking that same file's anchor reports only the `expected` error. Cross-file masking — the bug the PR fixes — is genuinely gone; this is the much smaller residue. Defensible (no valid comparison target), though the anchor check does not depend on `expected` and could still run.

**F5 — checked, no defect: the rewritten SKILL.md sentence does not over-claim.** "dropping a name from the subset, or adding one beyond it, both fail CI" — verified true in both directions. "A subsetting tier that has not registered its subset is still unchecked on omission" — correctly scoped and honest about the remaining gap. This is the same defect class the PR fixes, and the PR does not commit it.

**F6 — checked, no defect: the `validator.md` reformat is legitimate.** I judged this on whether the fenced form reads the same to an agent following it, and whether the file was bent to suit a parser quirk. The grammar genuinely cannot parse names inline after prose + colon (`find-load-list-runs` requires names-only lines), so the claim is true. The resulting shape is **not** a bespoke accommodation — it is byte-identical in structure to `fix-agent.md`'s pre-existing load list, so the change makes `validator.md` consistent with its sibling rather than idiosyncratic. Dropping the per-name backticks is correct inside a fence. Teaching the grammar prose-prefixed inline lists would loosen precision across all ten satellites to serve one file; reusing the house shape is the higher rung. "No semantic change" holds.

**F7 — the claude-skills-151 interaction is legitimate; merge order does not matter.** `core-list-satellites` parses `path: "..."` with `(?m)path: "(?P<p>[^"]+)"` to build the dupe-exemption set. I confirmed all 10 matches in the file are real `SATELLITES` registrations — no stray strings, and the two new entries use the identical `  { path: "…"` line shape as the existing eight, so any tightening that accepts the current entries accepts these. The expansion is also semantically right: these two files are now genuinely owned by `validate-core-list.nu`, which is what the exemption describes. Empirically a no-op today — the dupe report is unchanged, and both subset lists are far below the 8-line window. The two PRs touch different files (`validate-core-list.nu` vs `validate-skills-quality.nu`), so there is no textual conflict in either order; whoever lands second just re-runs `mise test`.

---

### What I could not verify

- Remote CI on the merge commit — the branch does not currently merge, so no post-rebase CI result exists to check.
- Whether `mise test` stays green after the rebase pulls in #190 and #195. It is green on the branch as-is and those PRs touch unrelated plugins, but that is inference, not a run.

### Restraint

The diff earns its size. `expected` is one optional field reusing the existing `get -o … | default` idiom and one `if` selecting the comparison target — no new abstraction, no config knob, no parser change. Three self-test cases cover exact/lower/upper bounds with no redundancy. The doc-comment expansion is proportionate to a rule readers must follow. Nothing here is speculative scaffolding.

vinnie357:
## Gate 3 — re-verification (round 2)

**Verdict: APPROVE.** F1 and F2 confirmed mechanically. F3 is real enforcement, not a token check: my exact round-1 repro now fails, the guard fails under its own target change, and it reports rather than aborts. Two minor notes below, neither blocking.

Scratchpad clone throughout; shared tree never checked out; every mutation restored (`git status` clean).

---

### F1 — confirmed

Rebased onto `a8112e9`; `git merge-base gh/main HEAD` = `a8112e9` exactly. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, both checks SUCCESS ("Validate marketplace and plugins", "Check version bumps").

### F2 — confirmed, all four reads by hand

| | main `a8112e9` | branch `38fabef` |
|---|---|---|
| `plugins/core/.claude-plugin/plugin.json` | 0.2.48 | **0.2.49** |
| `.claude-plugin/plugin.json` | 0.4.98 | **0.4.99** |
| `marketplace.json` → core | 0.2.48 | **0.2.49** |
| `marketplace.json` → all-skills | 0.4.98 | **0.4.99** |

All four exactly one above main, and the two `all-skills` reads agree with each other — checked by hand via `jq`, since `test:version-bumps` still reports only `core` and cannot see the all-skills pair.

---

### F3 — attacked properly, holds

**1. The original repro is closed.** Deleting `/core:anti-fabrication` from **both** `fix-agent.md`'s load list **and** its `expected: [...]` array — the mutation that returned exit 0 in round 1:

```
❌ plugins/core/skills/agent-loop/references/fix-agent.md:
   expected subset drops never-omissible name(s) — /core:anti-fabrication
```

Exit 1. Same mutation against `validator.md`'s 2-name subset also fails with the same message — so the fix covers the small-subset case that previously relied on adjacency.

**2. The file-only case is still caught independently**, as the author claims — dropping the name from the file while `expected` still declares it reports `missing from load list — /core:anti-fabrication` via the ordinary comparison, not the new check. The two paths are genuinely separate.

**3. It reports rather than aborts.** Planting a never-omissible violation in `validator.md` *and* a real pointer drift in `team-leader.md` in the same run surfaces **both** findings. The round-1 masking fix survived the refactor into `expected-preflight-errors`.

**4. The guard is not decorative.** Four mutations to the new code, each failing exactly the case that targets it — no over- or under-coverage:

| Mutation | Self-test cases that go red |
|---|---|
| `expected-preflight-errors` gutted to always `[]` | `unknown_name_reported` + `drops_never_omissible_reported` |
| `NEVER_OMISSIBLE` comparison deleted | `drops_never_omissible_reported` only |
| `NEVER_OMISSIBLE` const emptied | `drops_never_omissible_reported` only |
| unknown-name comparison deleted | `unknown_name_reported` only |

39 cases confirmed.

**5. Two further escape attempts, both closed or honestly documented:**

- Dropping `/core:anti-fabrication` from the **canonical block itself** in `SKILL.md` → caught: `Expected 10 canonical core skill names, found 9`.
- Deleting the `fix-agent.md` `SATELLITES` registration entirely *and* dropping the name from the file → **passes green.** Registration is opt-in, so deleting it removes the protection silently. This is not a regression and not a new hole — it is exactly the limit the doc comment and `SKILL.md` both state in as many words ("A subsetting tier that has NOT registered an `expected` entry is still invisible here"). Noting it only so the boundary is on the record.

---

### Findings

**N1 — LOW, no code defect: the stated reason for the local fixture is wrong, but the fixture is right.** The brief asked me to verify the reasoning. I substituted the shared `$canonical` (`[/core:aa /core:bb /core:cc]`) for `ef_canonical` and ran it. The shared fixture does **not** make the new tests pass vacuously — it makes two of them **fail loudly**:

```
• expected_preflight_clean_subset_passes:
    expected [], got ["...not in the canonical list — /core:anti-fabrication"]
• expected_preflight_unknown_name_reported:
    expected ["...— /core:zz"], got ["...— /core:anti-fabrication, /core:zz"]
```

The mechanism is that `expected` names are cross-checked against canonical, so `/core:anti-fabrication` must be a *member* of the fixture canonical or it pollutes the `unknown` list. Notably, `drops_never_omissible_reported` — the case that actually matters — would have passed either way, because `NEVER_OMISSIBLE` is a const independent of canonical.

So: **the local fixture is genuinely required and the decision is correct** (credit where due — it is not noise). The justification recorded for it is inaccurate. Worth correcting in the bee/commit record so the next reader does not inherit a wrong model of why it exists.

**N2 — LOW: `NEVER_OMISSIBLE` duplicates doctrine with nothing checking the copy.** Answering the brief's question directly: the const at `validate-core-list.nu:122` encodes a rule stated in prose at `SKILL.md:173` ("`/core:anti-fabrication` is never omissible"). Nothing ties them. If the doctrine ever grows a second never-omissible skill, the const stays at one entry and the drift is silent.

This is a weaker position than the file's own precedent — `EXPECTED_COUNT = 10` *is* validated against the real canonical block, which is what caught escape attempt 5a above. The asymmetry is defensible today: the canonical block is a machine-parseable fenced list, whereas the never-omissible rule is a clause inside a paragraph with no extractable shape. Making it checkable would mean giving the doctrine a parseable form — a larger change than this PR should carry. Reasonable as-is; worth a follow-up bee rather than a change here.

**N3 — informational: `SKILL.md` under-claims.** The "Two limits" paragraph says subset membership is checked in both directions but does not mention that `expected` itself is now guarded against dropping a never-omissible name. That is conservative rather than wrong, and a reader combining it with the doctrine sentence reaches the correct conclusion. No misleading text. Optional one-clause addition.

---

### Restraint

The new code is proportionate. `expected-preflight-errors` extracted out of `main`'s loop earns the extraction — the four new cases call it **directly**, which is what makes them precise enough that each mutation fails exactly one case. Left inline, they would have had to go through `check-satellite` or `main` and would have been coarser. `NEVER_OMISSIBLE` as a one-element const rather than a bare literal is a readability call, not speculative config: it is named, documented, and points at the doctrine it encodes. No new abstraction, no config knob, no parser change.

### Gates re-run on the rebased branch

`mise test` green · `gitleaks detect` — 279 commits, no leaks · `test/quality-baseline.json` diff empty · 39/39 self-test cases.

### Correction to my own round-1 process

My first attempt at the fixture-necessity probe used a `perl` replacement in which `$canonical` interpolated as an empty Perl variable, producing a nushell parse error that I nearly read as a test failure. Re-run with proper escaping; the result reported in N1 is from the corrected run.

PR #199, opened 2026-08-01
fix(test): wire Pass-2 detail counts and share the ratchet accumulator

vinnie357:
- Wire Pass-2 agents/commands `links`/`fm_schema` detail counts into the ratchet's `failing_counts` (claude-skills-175) — previously computed but never fed, so a waiver on either check absorbed unbounded later growth on the same file
- Extract `evaluate-agent-file` / `evaluate-command-file` (Gate 3 follow-up): all Pass-2 per-file checks, including the detail-count wiring, now live in these two functions instead of inline in main's loops, and are unit-tested directly against real mktemp fixtures (`run-pass2-eval-self-test`) — this is the regression test Gate 3 asked for; deleting the wiring now fails `--self-test`, not only a live corpus scan
- Extract shared accumulators `accumulate-detail-counts` (Pass 1 + Pass 2) and `accumulate-findings` (Pass 3 dupe + Pass 4 vocab), tested directly (`run-accumulator-self-test`, 6 cases including one chaining into `ratchet-baseline`)
- Restructure Pass 4 (vocab) from a per-iteration `mut` loop to a functional collect-then-batch shape matching Pass 3 — one `accumulate-findings` call regardless of how many formats fire, instead of a call repeated per firing format
- Harden `core-list-satellites` (Gate 3 finding): replace the exact-count self-test assertion with a lower bound + "every derived path resolves on disk" — the exact count collided with claude-skills-198 registering two more satellites, red on either merge order; and filter full-line comments before parsing, since a comment containing `path: "..."` inside the SATELLITES block was leaking into the result (case 14c)
- Correct two doc-comment claims that Gate 3 review disproved by measurement: the accumulator extraction narrows the accumulation VALUE's correctness (now unit-tested) but does NOT remove the two-statement half-delete shape from Pass 3/4's call sites — deleting only the `failing_counts` unpack line after `accumulate-findings` reproduces claude-skills-151 exactly, with both `--self-test` and the full run staying green. What DOES change: a WHOLE call-site deletion is caught by the full run via a baseline stale-key failure. Comments at each call site now state this precisely instead of claiming the surface narrowed.

## Gate 3 round 2 — what changed and why

Round 1 review (REQUEST CHANGES) found three defects, all now fixed:

1. **Exact-count assertion collided with #198 in either merge order.** Fixed per the reviewer's proposed design: lower bound + path-exists, not a bumped number. Also fixed the underlying parse gap the reviewer found while proposing it (a comment-embedded `path: "..."` leaking through `core-list-satellites`) rather than just working around it in the self-test.
2. **175's fix had no regression test.** Fixed by extracting `evaluate-agent-file`/`evaluate-command-file` — this is the "cheaper unexplored rung" the reviewer suggested (mktemp fixture pattern matching `run-pass2-links-self-test`), engaged with directly rather than restated.
3. **A PR-body claim was contradicted by the reviewer's own measurement** (single-line delete of the `failing_counts` unpack after `accumulate-findings` reproduces 151 exactly). Corrected in five places: the `accumulate-detail-counts` doc comment, the Pass 3 (`dupe_acc`) call-site comment, the Pass 4 (`vocab_acc`) call-site comment, the `evaluate-agent-file` doc comment, and the `apply-surface-finding` doc comment. Each now states precisely what's proven (accumulation VALUE correctness, hermetically; a whole call-site deletion caught via stale baseline keys) versus what isn't (a half delete at any unpack site, main-loop or Pass 3/4, remains invisible to both self-test and the full run).

On the two "medium" items: `accumulate-findings` now pays for itself for Pass 4 too, after the batch-collect restructuring (fixed 3-line footprint regardless of firing-format count, matching Pass 3, rather than a call repeated per format). The cheaper rung for concern 2 (extract + mktemp-test) was implemented, not just cited.

Every fix in this round was verified by reproducing the reviewer's exact mutation before the fix (red) and again after (still red where inherent, now caught where fixable) — see the PR thread for verbatim before/after output.

### DIFF
--- a/test/validate-skills-quality.nu
+++ b/test/validate-skills-quality.nu
@@ -568,10 +568,31 @@ def ratchet-baseline [baseline: list, failing_keys: list, failing_counts: list] 
 # every pass's detail-count accumulation through this one function turns
 # "a pass computed a count but never fed the ratchet" into a single,
 # reviewable call site instead of N independently-forgettable inline loops.
-# It narrows the mutation surface (no more two-statement loop body to
-# half-delete) but does not make a *whole-call* deletion self-test-visible —
-# see the claude-skills-151 PR body for why a full main-body integration
-# harness was weighed and not built this round.
+#
+# What this DOES and does NOT close (corrected after a Gate 3 review of the
+# first version of this comment, which overclaimed): the accumulation VALUE
+# is now correct and directly unit-tested (run-accumulator-self-test,
+# run-pass2-eval-self-test) wherever a caller routes through this function
+# instead of hand-rolling the loop. It does NOT remove the two-statement
+# shape from every call site — Pass 3 and Pass 4 still unpack this
+# function's `{failing_keys, failing_counts}` result into two separate
+# assignment lines, and the Gate 3 reviewer proved that pair is exactly as
+# half-deletable as the original inline loop was: deleting only the
+# failing_counts unpack line reproduces claude-skills-151 with both
+# --self-test and the full run staying green. Two things genuinely did
+# change: (1) the VALUE this function computes, given its inputs, is now
+# provably correct in isolation — a bug in the accumulation logic itself,
+# as opposed to a caller forgetting to use it, is caught directly; (2) a
+# *whole* call-site deletion (removing this function's call along with
+# BOTH unpack lines) IS caught by a full run against a real baseline,
+# because the finding's key also disappears from failing_keys, which
+# trips the baseline's stale-key check — not self-test-visible, but not
+# silent either. The genuinely invisible mutation, in both Pass 3 and
+# Pass 4, remains the half delete. Pass 2 (agents/commands) closes this
+# further via evaluate-agent-file / evaluate-command-file below, which
+# fold the wiring call and its result into one function that IS
+# hermetically self-tested directly — see that pair's doc comment for
+# what residual remains even there.
 def accumulate-detail-counts [failing_counts: list, key_base: string, failed: list, detail_entries: list] {
     mut fc = $failing_counts
     for e in $detail_entries {
@@ -1429,7 +1450,15 @@ def satellites-block-lines [all_lines: list] {
 
 def core-list-satellites [script_path: string] {
     let raw = (open --raw $script_path)
-    let block = (satellites-block-lines ($raw | lines) | str join "\n")
+    # Full-line comments are dropped before parsing (claude-skills-151 Gate 3
+    # finding): satellites-block-lines only bounds WHERE the parse looks
+    # (inside vs outside the SATELLITES block); a comment line INSIDE the
+    # block that happens to contain the literal substring `path: "..."` —
+    # documentation, a worked example, a commented-out entry — would
+    # otherwise leak into the result exactly as if it were a real record.
+    let block_lines = (satellites-block-lines ($raw | lines)
+        | where {|l| not ($l | str trim | str starts-with "#")})
+    let block = ($block_lines | str join "\n")
     let sat = ($block | parse --regex 'path: "(?P<p>[^"]+)"' | get p)
     let canonical = ($raw | parse --regex 'const CANONICAL_FILE = "(?P<p>[^"]+)"' | get p)
     $sat | append $canonical | uniq
@@ -1623,17 +1652,46 @@ def run-duplicate-self-test [] {
     # Case 14: satellite derivation — the list parsed out of
     # test/validate-core-list.nu carries the canonical file and the known
     # satellite shapes; if the parse silently broke, exemption 3 would too.
-    # Bounded to the real, measured count (8: the 8 SATELLITES entries,
-    # deduped against CANONICAL_FILE which repeats the first) rather than
-    # only a lower bound — a lower-bound-only assertion would not have
-    # caught the unscoped-regex bug this case's sibling (case 14b) exists
-    # to guard.
+    # Bounded to a LOWER bound plus "every derived path resolves on disk"
+    # (claude-skills-151 Gate 3 finding) rather than an exact count: an
+    # exact-count assertion collided with claude-skills-198 registering two
+    # more satellites — green on either branch alone, red once both merge,
+    # in EITHER merge order, since each branch only sees its own addition.
+    # A lower bound plus "every path is real" tolerates legitimate growth
+    # with zero maintenance while still catching a leaked bogus path — a
+    # comment-embedded `path: "..."` string won't resolve on disk.
     let repo_root = (git rev-parse --show-toplevel | str trim)
     let derived = (core-list-satellites ($repo_root | path join "test" "validate-core-list.nu"))
-    if ("plugins/core/skills/agent-loop/SKILL.md" not-in $derived) or ("plugins/core/hooks/session-start.sh" not-in $derived) or (($derived | length) != 8) {
-        print $"(ansi red_bold)❌ duplicate self-test: satellite derivation from validate-core-list.nu broke \(got ($derived | length) entries, want 8\)(ansi reset)"
+    let missing_on_disk = ($derived | where {|p| not (($repo_root | path join $p) | path exists)})
+    if ("plugins/core/skills/agent-loop/SKILL.md" not-in $derived) or ("plugins/core/hooks/session-start.sh" not-in $derived) or (($derived | length) < 5) or ($missing_on_disk | is-not-empty) {
+        print $"(ansi red_bold)❌ duplicate self-test: satellite derivation from validate-core-list.nu broke \(got ($derived | length) entries; missing on disk: ($missing_on_disk | str join ', ')\)(ansi reset)"
         $failed = true
     }
+
+    # Case 14c (claude-skills-151 Gate 3 finding): a comment LINE inside the
+    # SATELLITES block that happens to contain the literal substring
+    # `path: "..."` — documentation, a worked example, a commented-out
+    # entry — must not leak into the parsed satellite list. Exercises
+    # core-list-satellites end-to-end (not just satellites-block-lines) via
+    # a real fixture file, since the comment-filter lives in
+    # core-list-satellites itself.
+    let cl_root = (mktemp -d)
+    let cl_script = ($cl_root | path join "fixture.nu")
+    ('const SATELLITES = [' + (char newline)
+        + '  # example: path: "fake/example.md" -- not a real satellite' + (char newline)
+        + '  { path: "real/entry.md"' + (char newline)
+        + '    anchor: "x" }' + (char newline)
+        + ']' + (char newline)) | save $cl_script
+    let cl_derived = (core-list-satellites $cl_script)
+    if ("fake/example.md" in $cl_derived) {
+        print $"(ansi red_bold)❌ duplicate self-test: a path: \"...\" string inside a comment line leaked through core-list-satellites(ansi reset)"
+        $failed = true
+    }
+    if ("real/entry.md" not-in $cl_derived) {
+        print $"(ansi red_bold)❌ duplicate self-test: core-list-satellites dropped a genuine in-block path after comment filtering(ansi reset)"
+        $failed = true
+    }
+    rm -rf $cl_root
 
     # Case 14b (claude-skills-151 follow-up): a `path: "..."` line OUTSIDE
     # the SATELLITES block — e.g. a future test fixture or self-test case in
@@ -1698,7 +1756,7 @@ def run-duplicate-self-test [] {
     }
 
     if not $failed {
-        print $"(ansi green_bold)✅ Duplicate-block self-test passed \(16 cases\)(ansi reset)"
+        print $"(ansi green_bold)✅ Duplicate-block self-test passed \(17 cases\)(ansi reset)"
     }
     $failed
 }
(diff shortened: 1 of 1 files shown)


vinnie357:

Reviewed by a separate agent briefed to defeat the change. Every number below was re-derived in an isolated clone (`gh/main` = `d1a79d9`, branch = `c9d98b6`); nothing was inherited from the PR body. All mutation runs used `nu 0.113.1` (the pinned version).

Line numbers refer to `test/validate-skills-quality.nu` **on this branch**.

---

### BLOCKING — 1. This PR and open PR #198 break each other's CI, in either merge order

Counted `SATELLITES` entries in `test/validate-core-list.nu`:

| ref | entries | `core-list-satellites` result |
|---|---|---|
| `gh/main` | 8 | 8 |
| this branch | 8 | 8 |
| `gh/feat/subset-aware-satellites` (#198) | 10 (adds `validator.md`, `fix-agent.md`) | 10 |

Constructed the merged state (main corpus + this branch's script + #198's `validate-core-list.nu`) and ran the repo's own gate:

```
$ mise run ci
[test:skills-quality] ❌ duplicate self-test: satellite derivation from validate-core-list.nu broke (got 10 entries, want 8)
[test:skills-quality] ERROR task failed
MISE_CI_BOTH_PRS_EXIT=1
```

Each PR is green in isolation only because neither branch contains the other's change. Whichever lands second puts `main` red.

**2. The exact-count assertion is the wrong shape for a list that is expected to grow.**

Case 14's job is to stop exemption 3 *silently broadening*. `!= 8` achieves that, but it also fires on every legitimate satellite addition — #198 is the first instance and it arrived before this PR even merged. Fixing the collision by bumping `8` → `10` just re-arms the same trap for the next addition.

I attacked the alternatives to check an upper bound is even needed. It is — here is a leak the old `< 5` lower bound missed and that block-scoping alone does **not** close:

```
# a comment INSIDE the SATELLITES block:
  # shape reminder: path: "some/example/file.md" plus an anchor
```
→ `core-list-satellites` returns **9** entries including `some/example/file.md`. Scoping bounds the parse to the block but does not distinguish comments from records.

A growth-tolerant assertion that still catches that leak, and needs no maintenance on additions:

```nu
# every derived satellite must be a real file in this repo — a leaked
# comment/fixture string is not, a legitimate new satellite always is
let missing = ($derived | where {|p| not (($repo_root | path join $p) | path exists)})
if ($missing | is-not-empty) or (($derived | length) < 5) { ... }
```

Verified all 10 of #198's satellite paths resolve on `main`, so this assertion is green under both PRs and the collision disappears — merge order stops mattering.

**Recommended order:** land **#198 first**, then this PR. If the exact count is kept anyway, whichever goes second must edit `test/validate-skills-quality.nu` case 14 — a file this PR already owns and #198 does not touch at all, so putting #199 second costs zero extra surface. But the growth-tolerant assertion is the better fix; please prefer it.

---

### HIGH — 3. The claude-skills-175 fix ships with no regression test

Deleting the wiring block that **is** the fix leaves everything green:

| mutation | `--self-test` | full run |
|---|---|---|
| delete lines 2681–2684 (Pass-2 **agents** `accumulate-detail-counts` call) | exit 0 | exit 0 |
| delete lines 2741–2744 (Pass-2 **commands** call) | exit 0 | exit 0 |

The accumulator self-test proves the shared function's *value* semantics. Nothing proves the Pass-2 loops still *call* it. Per `/core:tdd`, a fix with no failing-test-first is not complete — and this one is reachable: my own reproduction (below) is the harness.

### HIGH — 4. The claimed mutation-surface narrowing does not hold

PR body: *"narrows the mutation surface a two-statement loop body previously exposed"*, and the `accumulate-detail-counts` doc comment: *"It narrows the mutation surface (no more two-statement loop body to half-delete)"*.

The two-statement half-delete was **relocated to the call site**, not removed. Delete line 2808 alone — `$failing_counts = $dupe_acc.failing_counts`, keeping `$failing_keys = $dupe_acc.failing_keys`:

```
M3b_SELFTEST_EXIT=0
M3b_FULL_EXIT=0
```

Then reproduced the real defect on top of it. Grew the waived `dupe/22d40d6a` zig group by appending an identical 11-line block to both member files:

```
# unmutated
UNMUT_EXIT=1
  dupe/22d40d6a:duplicate_block — 6 shared window(s)
  dupe/22d40d6a:duplicate_block: current 6 > baselined 1

# with line 2808 deleted, same corpus
MUT_EXIT=0        (no regression lines at all)
MUT_SELFTEST_EXIT=0
```

That is claude-skills-151 verbatim, in one line, invisible to both gates. Pass 4 behaves identically (delete line 2870: self-test 0, full run 0).

Refinement of the stated residual, which is worth correcting because it points at the wrong risk: a **whole**-call-site deletion (lines 2806–2808) is self-test-invisible **but the full run catches it** (exit 1 — the dupe keys go stale in the baseline). The genuinely invisible mutation is the **half**-delete, which the PR body does not name. Please correct both the body bullet and the doc comment; as written they overstate what the extraction bought.

---

### MEDIUM — 5. `accumulate-findings` does not earn its place (`/core:restraint`)

- Pass 4 went from 2 lines (two direct appends) to **3** lines (call + two unpacks).
- Pass 3 went from 4 lines to 3.
- Mutation surface: unchanged (finding 4).
- The behaviour it encapsulates is two `| append` calls.

Its only demonstrated benefit is self-test cases 4–6. Rung 1 of the ladder — does this need to exist? `accumulate-detail-counts` clearly does (it *is* the 175 fix and dedups a real conditional loop). `accumulate-findings` reads as symmetry for its own sake. Consider dropping it and keeping the two direct appends, or fold the unpack into the call so there is one statement per site.

### MEDIUM — 6. The `VOCAB_REAL_FLOORS` argument is accurate but not decisive

Read the floor logic. The claim checks out: `VOCAB_REAL_FLOORS` is a `const` read directly inside `vocab-real-floor-errors` (line 1833) with no injection seam, `main` has no fixture mode, and the floor check at line 2837 is an unconditional `exit 1`. A small synthetic repo does trip it regardless of the mutation.

But it is one of several unconditional exits (the `doc_empty` extractor canary at line 2876 would also fire on a fixture that lacks the claude-code doc skills), and `main` derives `repo_root` from `git rev-parse --show-toplevel`. The floors are not the obstacle; the whole-repo shape of `main` is. Presenting the floors as the deciding factor makes the restraint call look narrower-grounded than it is.

More substantively, a cheaper rung was not considered. `run-pass2-links-self-test` already builds a synthetic tree with `mktemp -d` and tests an extracted predicate (`resolve-pass2-path`) directly. Extracting the per-file Pass-2 evaluation into a function returning `{failed, counts}` and testing it with that same existing pattern would close findings 3 and 4 together, without any live-corpus perturbation. That is rung 2 (reuse a pattern already in this codebase), and it sits well below the full integration harness that was weighed and rejected.

On "live corpus perturbation with restore guarantees … a bigger, riskier undertaking": I performed exactly that perturbation five times during this review with plain `cp` restore and verified byte-identical output afterward each time. The interrupt-safety concern under CI is real; the difficulty claim is overstated.

---

### LOW / nits

7. `run-accumulator-self-test` case 1's message misreports the failure mode. Under mutation 1 (gut the count-append) it prints *"included a check absent from failed"* when the actual failure is that it included **nothing**. The assertion is correct; the diagnostic sends a reader the wrong way.
8. `satellites-block-lines` matches the opener with `str starts-with` on the **untrimmed** line. An indented `const SATELLITES = [` collapses the derived list to 1 entry. Loud (case 14 and the full run both fail), but `str trim` is free.
9. The doc comment asserts *"Entries in SATELLITES are flat `{ path: … anchor: … }` records with no nested `[`/`]`"*. That invariant is **already false on #198's branch**, which adds `expected: [/core:mise …]`. Single-line today so the parse survives; a wrapped `expected` list truncates it (verified: 2 entries returned, both gates fail loudly). Please restate the invariant as "no bare `]` line inside the block" — which is what the code actually depends on.

---

### What I attacked and could not break

Both reproductions were re-run from scratch, not inherited.

**claude-skills-175, independently reproduced.** 3 broken links added to `plugins/pm/agents/pm-lead.md`, waiver at `detail_count: 1`:
```
MAIN_EXIT=0                                                   ← fully absorbed, pre-fix
BRANCH_EXIT=1  pm/agents/pm-lead.md:links: current 3 > baselined 1
```
Same result on the commands surface (`core/commands/work.md:links`) and on `fm_schema` (`pm/agents/pm-lead.md:fm_schema`) — all four wiring points work.

**Accumulator mutations both go red.** Gutting the count-append inside `accumulate-detail-counts` (line 579) → self-test exit 1, cases 1 and 3. Gutting it inside `accumulate-findings` (line 596) → self-test exit 1, cases 4 and 6. The extraction achieves its stated purpose for the in-function shape.

**No behaviour change.** Ran `main`'s script and this branch's script against an identical corpus (main + branch script only). Full-run stdout **byte-identical**: 102 skills, 57/102 perfect, 4 dupe groups (`1182fd06`=5, `22d40d6a`=1, `28ccd32d`=7, `639413bb`=1), 49 debt entries. All match the PR's claims.

**Pass-1 semantics unchanged (finding 6 of the brief).** The byte-identical comparison above is weak evidence on its own — I dumped `failing_counts`/`failing_keys` and found only **4 count entries** (the dupe groups) and 49 keys (45 `anti_fab` + 4 dupe) on the clean corpus, meaning Pass 1's, Pass 2's and Pass 4's accumulator calls are all **dormant** today. So I forced Pass 1 live (3 broken links in `core/code-review` SKILL.md + a waiver at 1): both scripts produced identical `failing_counts`, identical `failing_keys`, and identical stdout, both `current 3 > baselined 1`. Pass 4 forced live too (`{{…}}` added to the claude-commands doc): call site executes cleanly and `syntax/commands:vocab_disjoint: current 4 > baselined 1` reaches the ratchet.

**Housekeeping, all verified:**
- `git diff gh/main -- test/quality-baseline.json` → empty. Baseline untouched.
- `mise run test:version-bumps main` → `✅ No plugin directories modified`. No bump required, none made.
- `mise exec gitleaks@latest -- gitleaks detect` over the PR patch → `no leaks found`.
- `mise run ci` on the true merge state (main + this branch's script) → exit 0, 24.3s.

### Could not verify

- Whether GitHub's merge of this branch (which is behind `main` by `#188`/`#189`/`#192`/`#195`) produces the same tree as my reconstruction. `main` does not touch `test/validate-skills-quality.nu` or `test/quality-baseline.json`, so no conflict is possible, but the merge itself was not performed.
- Nothing was run on a Linux CI runner; all results are macOS/arm64 with the pinned nu 0.113.1.

---

**Verdict: REQUEST CHANGES.** Two real bugs are genuinely fixed and the refactor is behaviour-preserving — that part is solid and independently confirmed. Blocking on finding 1 (CI-breaking collision), and asking for findings 3 and 4 to be addressed or answered: a fix with no regression test, and a PR body plus code comment that both claim a narrowing my mutations contradict.

vinnie357:
## Gate 3 round 2 — re-verification at `b0f6386`: APPROVE

All three round-1 findings are fixed. Every claim below was re-measured in an isolated clone against **current** `main` (`3aeaf7f`, which now carries #198's 10 satellites). Nothing inherited from the PR body.

Line numbers refer to `test/validate-skills-quality.nu` at `b0f6386`.

---

### 1. The collision — fixed, verified against the real merged state

The branch is behind `main`, but `main` never touches `test/validate-skills-quality.nu` or `test/quality-baseline.json`, so the merge is conflict-free. Built it (current `main` + this branch's script) and ran the gate that failed last round:

```
$ nu -c "source …/validate-skills-quality.nu; print (core-list-satellites …/validate-core-list.nu | length)"
10

$ mise run ci
MISE_CI_EXIT=0
```

Same construction returned `❌ … got 10 entries, want 8` / exit 1 last round. Case 14 is now a lower bound plus "every derived path resolves on disk", which is the design I proposed — and it tolerates the growth that broke the exact count.

**Re-ran the comment-leak attack against the new parser**, plus the three variants you named. Fixture files built by hand, `core-list-satellites` called directly:

| variant | result |
|---|---|
| commented-out whole entry inside the block | stripped — correct |
| real path containing `#` (`…/x#frag.md`) | survives — no false negative |
| full-line comment with `path: "fake/example.md"` | stripped (case 14c) — correct |
| bogus path leaked via **inline trailing** comment | leaks, but `❌ … missing on disk: legacy/old-work.md` → exit 1 |

One leak gets through both gates — see nit N1. It is contrived and I am not blocking on it.

Also re-tested the truncation hazard from round 1's nit 9, now that `main` really ships `expected: [...]` entries. Wrapping one of those lists across lines truncates the derived list 10 → 5; caught loudly by both gates (`I_SELFTEST_EXIT=1`, `I_FULL_EXIT=1`) via the `session-start.sh not-in $derived` clause. Narrowing stays loud, as the design claims.

---

### 2. The Pass-2 wiring test — the rung was implemented, and it holds

Reproduced the claimed proof rather than accepting it. Reverting `evaluate-agent-file`'s `failing_counts` to `[]` — the same wiring whose deletion was invisible at old lines 2681–2684:

```
MUT_A_SELFTEST_EXIT=1
❌ pass2-eval self-test: evaluate-agent-file computed a links finding but failing_counts is missing it (got [])
❌ pass2-eval self-test: evaluate-agent-file computed an fm_schema finding but failing_counts is missing it
```

Same mutation on `evaluate-command-file`: exit 1, two matching messages. Round-1 finding 3 is closed.

**Then attacked the new seam, per your question.** Dropped only the counts append *inside* `apply-surface-finding` (line 2436 → `failing_counts: $failing_counts`):

```
MUT_C2_SELFTEST_EXIT=1
❌ pass2-eval self-test: apply-surface-finding did not append all three accumulators together
```

Same for the `surface_results` append (line 2437): exit 1. **The function is atomic and hermetically tested — the invisibility did not move inside it.** It moved to main's three-line unpack, and only there: deleting `$failing_counts = $acc.failing_counts` at line 2963 gives self-test 0 / full run 0, exactly as the doc comment at 2422–2429 states, with the correct reason (nushell `mut` has no one-statement destructure-assign). Documented residual, accurately scoped.

So the pattern you flagged — "extraction moves the untestable wiring rather than removing it" — is now bounded to a single one-line unpack per call site, with the previously-untestable 4-line block pulled inside a tested function. That is a real reduction, not a relocation.

---

### 3. The contradicted claim — corrected, and the corrections are accurate

Read all five sites and checked each against my own measurements: `accumulate-detail-counts` (572–595), `dupe_acc` (3032–3042), `vocab_acc` (3094–3100), `evaluate-agent-file` (2296–2306), `apply-surface-finding` (2422–2429). Each now states the boundary correctly — value unit-tested; whole call-site deletion caught by the full run via a stale-key baseline failure; half delete invisible except inside the two new Pass-2 functions. The 572–595 comment explicitly records that its first version overclaimed.

I re-ran the mutation behind the correction on the new code (delete only `$failing_counts = $dupe_acc.failing_counts`, line 3045): self-test 0, full run 0. Still invisible, still stated as such. One sixth site was missed — see nit N2.

---

### 4. Restraint — `accumulate-findings` pays for itself now

Round 1 I said it didn't: Pass 4 went from two direct appends to three lines, with no mutation-surface reduction. The restructure changes that assessment. Pass 4 is now a functional collect (`vocab_evals` → `vocab_doc_empty` / `vocab_fired` / `vocab_findings`) plus one call and two unpacks, matching Pass 3 exactly. It removed two `mut` accumulators and the per-format `for` loop, and the call site is fixed-size regardless of how many formats fire. Both call sites now have one uniform shape. Withdrawn.

---

### Behaviour preservation — the check that mattered most, given the diff grew to 603/147

Byte-identical full-run output between `main`'s script and this branch's script on an identical corpus (102 skills, 57/102 perfect, 4 dupe groups, 49 debt entries).

That alone is weak, because Pass 1, Pass 2 and Pass 4 are all dormant on a clean corpus — I established last round that `failing_counts` holds only the 4 dupe entries. So I forced findings in **every** pass simultaneously (3 broken links in a SKILL.md; 2 unknown frontmatter keys + 2 broken links in `pm-lead.md`; 3 broken links in `commands/work.md`; `{{…}}` in the claude-commands doc) and diffed the internal accumulator state between the two scripts:

```
failing_keys     → IDENTICAL
surface_results  → IDENTICAL
stdout           → IDENTICAL
failing_counts   → differs by exactly 3 entries, all additions:
                     core/commands/work.md:links       count 3
                     pm/agents/pm-lead.md:links        count 2
                     pm/agents/pm-lead.md:fm_schema    count 2
```

That delta *is* the claude-skills-175 fix and nothing else. The Pass-2 extraction into `evaluate-agent-file`/`evaluate-command-file` preserved `surface_results` byte-for-byte, and the Pass-4 restructure changed no output.

### Housekeeping — all confirmed

- `mise test` → exit 0.
- All 10 skills-quality self-test suites green: frontmatter (5), baseline schema/ratchet (18), check-fix (23), duplicate-block (17), vocab (10), Pass-2 links (8), orphans (6), frontmatter-schema (12), accumulator (6), Pass-2 eval (9).
- `git diff gh/main -- test/quality-baseline.json` → empty.
- `mise run test:version-bumps main` → `✅ No plugin directories modified`. `test/` is not a plugin; no bump required, none made.
- `mise exec gitleaks@latest -- gitleaks detect` over the full PR patch → `no leaks found`.
- Disclosure lint green (runs inside `mise test`).

---

### Nits — none blocking, none needs to hold the merge

**N1 — inline trailing comments are not stripped.** The filter is `str trim | str starts-with "#"`, so only full-line comments go. A trailing comment on a real record line still contributes to the regex:

```nu
  { path: "plugins/core/commands/work.md"  # superseded path: "plugins/core/skills/git/SKILL.md"
```
→ derives 11 entries including `plugins/core/skills/git/SKILL.md`. That path **exists**, so the disk check passes and both gates stay green (`H_SELFTEST_EXIT=0`, `H_FULL_EXIT=0`) — exemption 3 silently widens by one real file. The realistic form (a bogus/example path) is caught, as shown above. If you want it closed, anchoring the regex to a record-opening line does it in one edit: `(?m)^\s*\{ path: "(?P<p>[^"]+)"`, which kills both this and the full-line case without needing the comment filter at all.

**N2 — a sixth comment site was missed by the five-place correction.** `run-accumulator-self-test`'s doc comment (line 909) still reads *"that whole-call-site wiring gap is the recorded limitation of this fix"* — the framing the other five sites replaced. It is defensible in context (it is describing the *self-test's* scope, and a whole-call deletion genuinely is self-test-invisible) and it defers to the corrected comment, but "the recorded limitation" now names the wrong one of the two.

**N3 — round-1 nit 9 is still open and is now live.** Lines 1431–1433 assert *"Entries in SATELLITES are flat `{ path: … anchor: … }` records with no nested `[`/`]`"*. Since #198 merged, `main` ships two entries with `expected: [/core:mise …]`. Single-line today so the parse is safe, and a wrapped list fails loudly (verified above), but the invariant as written is false. Suggest restating it as what the code actually depends on: no *bare* `]` line inside the block.

---

**Verdict: APPROVE.** All three round-1 findings are fixed, each verified by re-running the mutation that exposed it rather than by reading the fix. The suggested cheaper rung was implemented and works. Behaviour is preserved everywhere I could force a live finding, with the single intended exception. The three nits are cosmetic or contrived and can land as follow-ups or not at all.

PR #200, opened 2026-08-01
fix(agent-sandboxing,zig,claude-code): example, dedupe, manifest-check gaps

vinnie357:
Three judgment calls (claude-skills-142, 150, 170), each decided deliberately rather than defaulted.

## claude-skills-142 — agent-substrate-overview had zero examples anywhere

Re-verified before touching anything: zero fenced code blocks, no "## example" header, no `references/` directory at all — the bee's DEBT classification (not a check defect) still holds.

**Decision: add a worked example**, not reclassify as CHECK_DEFECT. The skill's own "When to use this skill" section names "Reading a Substrate `ActorTemplate` / `WorkerPool` YAML and wanting to know the field semantics" as a documented use case — an example isn't filler here, it's the thing the skill promises. Pulled the real `WorkerPool` + one `ActorTemplate` (trimmed from the upstream `demos/claude-code-multiplex/claude-code-multiplex.yaml.tmpl`, Apache-2.0, fetched via `gh api` against `agent-substrate/substrate` to avoid inventing CRD field names) and annotated five fields whose semantics aren't obvious from the names alone. `sources.md` updated to record the new extraction and access date.

## claude-skills-150 — zig build/testing shared 8-line block

Confirmed the bee's correction: only one pair (`build.md` ↔ `testing.md`) reaches the 8-line window under the shipped check; it's the "wire the test step into build.zig" snippet, byte-identical apart from indentation.

**Decision: dedupe.** This isn't incidental overlap — `testing.md`'s copy silently assumed the reader already knew about the 0.15+ `root_module` requirement that `build.md` explains inline; a reader landing on `testing.md` first would get a working-looking snippet with no context for its shape, an early symptom of drift. Replaced `testing.md`'s fenced duplicate with a prose pointer to `/zig:build` `references/build.md` "build.zig Structure" plus a one-line explanation of what each call does — shorter, no longer duplicated, and doesn't lose the reader who never opens `build.md`.

## claude-skills-170 — manifest-agreement check: deletion escape hatch + order-sensitivity

Located the check first: `plugins/tools/claude-code/skills/claude-plugins/scripts/validate-plugin.nu` (`test/validate-plugin.nu` is a thin wrapper that delegates to it — one implementation, no duplicate copy to fix).

**Gap 1 decision: marketplace entries must carry `description`/`keywords` when plugin.json defines them.** The comparison used to gate on the *marketplace* side being non-empty, so deleting the marketplace field skipped the whole block silently. Regated on the *plugin.json* side instead — plugin.json is already the documented authority — so a marketplace-side deletion is now flagged as its own error ("marketplace.json entry is missing 'description'/'keywords' that plugin.json defines"), distinct from a content mismatch.

**Gap 2 decision: compare keywords as sets, not ordered lists.** The arrays are meant to carry the same discovery keywords; reordering them changes no meaning, so an order-only difference isn't a genuine mismatch and the old "mismatch" message was misleading about what actually differed. Now `sort`ed before comparison.

### Verification (breaking the check, not just reading it)

Ran all four mutations against `test/validate-plugin.nu dagu`, restoring between each:
1. Deleted `dagu`'s marketplace `description` → new error fires, exact text confirmed.
2. Restored, confirmed clean.
3. Reversed `dagu`'s marketplace `keywords` array (same set, different order) → passes clean (no false "mismatch").
4. Restored, added a genuine extra keyword → real "keywords mismatch" still fires, proving the set comparison isn't a no-op.
5. Combined mutation (description deleted **and** keywords set-mismatched simultaneously) → both errors surface together, not just the first — confirms the check reports rather than aborting on first finding.

## Baseline

`mise test` reported 2 baseline entries now pass; regenerated with `nu test/validate-skills-quality.nu --update-baseline` (shrink-only) — `agent-sandboxing/agent-substrate-overview:examples` and `dupe/22d40d6a:duplicate_block` removed, 49 → 47 allowed failures.

## Version bumps

Main moved twice while this was in flight (#192, #195 landed). Rebased onto current main (`d1a79d9`) before bumping, then bumped one above main's tip for every touched plugin: zig 0.1.5→0.1.6, agent-sandboxing 0.2.9→0.2.10, claude-code 0.2.46→0.2.47, all-skills 0.4.94→0.4.95 (plugin.json + marketplace.json for all four).


vinnie357:

Reviewed at branch HEAD `7c8f545` in a private clone (`git clone` from GitHub + `git fetch origin fix/examples-zig-dupe-manifest-agreement`); the shared working tree was never checked out. All numbers below are re-derived, not inherited from the PR body.

The 142 example and the 150 dedupe both survive attack. The 170 change does not: flipping the gate to the plugin.json side breaks both of the validator's non-CI invocation modes, and CI cannot see it.

---

### BLOCKER 1 — the gate flip breaks direct-file mode and external-plugin mode

`validate-plugin-content` declares `--mkt-description: string = ""` and `--mkt-keywords: list<string> = []`. **Two call sites deliberately omit both flags**, relying on those defaults to mean "no marketplace context — skip the comparison":

- `validate-plugin-file` (direct-file mode)
- the `$is_ext` branch of `validate-from-marketplace` — whose own comment says the comparison "is skipped entirely" for GitHub-object sources

The old gate `if ($mkt_description | is-not-empty)` skipped in both cases. The new gate cannot distinguish *"not supplied"* from *"supplied, but the marketplace entry deleted the field"*, so it fires on the defaults and reports a missing marketplace entry that was never passed in.

**Direct-file mode** — the invocation `SKILL.md:89` documents as the primary usage, and the script's own usage header lists as "Direct: Validate a plugin.json file directly":

```
$ nu plugins/tools/claude-code/skills/claude-plugins/scripts/validate-plugin.nu \
      plugins/languages/zig/.claude-plugin/plugin.json
  Errors: 2
  ✗ marketplace.json entry is missing 'description' that plugin.json defines: 'Zig programming skills: ...'
  ✗ marketplace.json entry is missing 'keywords' that plugin.json defines: [zig, systems-programming, comptime, allocators]
✗ Validation failed with 2 errors        (exit 1)
```

Same command, same file, `origin/main`'s copy of the validator: **0 errors, exit 0, "✓ Plugin is valid!"**

**External-plugin mode** — `claudio`, the marketplace's one GitHub-object source:

```
$ nu test/validate-plugin.nu claudio
  Errors: 2
  ✗ marketplace.json entry is missing 'description' that plugin.json defines: 'Project discovery, planning, and quality analysis system with parallel agents'
  ✗ marketplace.json entry is missing 'keywords' that plugin.json defines: [project-analysis, planning, ...]
✗ Validation failed with 2 errors        (exit 1)
```

Same command against `origin/main`'s validator: **0 errors, "✓ Plugin 'claudio' is valid!"**

**Why CI is green anyway:** `test:plugins` calls `test/validate-plugin.nu <name>` only for local sources (which do pass both flags), and handles GitHub sources with its own inline `git clone` + name-equality check that never invokes the validator. Neither broken path is exercised by any task in `mise run test`. I confirmed `mise run test` exits 0 on this branch — that green is real and also uninformative here.

**Fix requirement:** the two params need to distinguish "not supplied" from "supplied and empty" — e.g. nullable params gated on `!= null`, or an explicit `--compare-marketplace` boolean set only by the local-source call site. I did not verify either shape compiles; the requirement is what matters. Please add a check that would have caught this: an assertion that direct-file mode and the `claudio` path both still exit 0.

### BLOCKER 2 — two documents now describe behavior the code no longer has

Neither file is touched by this PR, and both are wrong on two counts each:

- `plugins/tools/claude-code/skills/claude-plugins/SKILL.md:94` — "Entries whose `source` is a GitHub object are skipped — there is no local manifest to compare — and **a field absent from either side is not a mismatch**."
- `plugins/tools/claude-code/skills/claude-plugins/references/validation-and-troubleshooting.md:58–61` — the same two claims.

The GitHub-object half is false today (Blocker 1). The "absent from either side" half is false *by design* — it is the thing the PR deliberately changed — so it needs rewriting even after Blocker 1 is fixed. Neither file documents the two new error messages, though the troubleshooting reference exists precisely to explain each message.

### MAJOR — all-skills 0.4.95 is a downgrade against current main

Current `origin/main` is `bc72282` with **all-skills at 0.4.96**. This PR takes 0.4.95. The PR body's "bumped one above main's tip" was accurate at `d1a79d9` but main moved.

I reproduced the tooling gap independently: `mise run test:version-bumps origin/main` prints `✅ All modified plugins have version bumps` and never mentions all-skills at all, because all-skills' *content* did not diff — a downward move is invisible to it. Its green is not evidence here.

Trial rebase onto `bc72282` in my own clone: **exactly two conflict hunks, both the all-skills version line** (`.claude-plugin/plugin.json:3`, `.claude-plugin/marketplace.json:16`). Nothing beyond version fields — the baseline shrink, the validator change, and the two skill edits all replay cleanly. Against current main, claude-code 0.2.48, all-skills 0.4.99, zig 0.1.6, and agent-sandboxing 0.2.10 are all free.

### MINOR 1 — factual error in the new example's framing

`agent-substrate-overview/SKILL.md:65` says `mars` and `orion` differ from `luna` "only in `metadata.name` and the `TASK` env value". Upstream also gives each a distinct `ACTOR_NAME` env value — `"luna"` / `"mars"` / `"orion"` at template lines 73–74, 103–104, 133–134. Three fields differ, not two.

### MINOR 2 — the `workerSelector` annotation cites the wrong Kubernetes object

"the same label-selector idiom as a Kubernetes Service" — a Service's `spec.selector` is a bare `map[string]string` with no `matchLabels` key. `matchLabels` belongs to `metav1.LabelSelector`, the Deployment/ReplicaSet/NetworkPolicy idiom, and that is exactly how upstream types it (`WorkerSelector *metav1.LabelSelector`). A reader following the analogy writes the wrong shape.

Secondly, "how an `ActorTemplate` **claims** a `WorkerPool`: it **must** match" overstates an `+optional` field. Upstream's godoc: "restricts which worker pools actors from this template may use. The scheduler only considers pools whose labels match this selector. **If nil, all pools are eligible**... Acts as a gate" — and it is AND'd with `SandboxClass`, which the annotation does not mention.

### NIT — "compared as SETS" is not what the code does

`sort` does not dedupe, so this is a *sorted-list* comparison. I verified the difference is load-bearing: a keyword duplicated on one side only still fires a mismatch (mutation M11 below). The behavior is stricter than a set, which is the safe direction — but the code comment and the PR body both say "sets", and a future maintainer reading that could reasonably assume duplicate drift is tolerated. Reword to "order-insensitive comparison".

---

## What I attacked, and what survived

**The added YAML example — verified against upstream, not invented.** Fetched `demos/claude-code-multiplex/claude-code-multiplex.yaml.tmpl` from `agent-substrate/substrate` via `gh api` and diffed field by field. Every field name, nesting level, and value in the skill matches upstream verbatim — the `WorkerPool` is upstream lines 42–53 including its comment, `agent-luna` is lines 62–88. The only omission is `INTERVAL_SECONDS: "45"` from the env list, which is legitimate under the declared "Trimmed from", though a reader lifting the snippet gets a manifest without it. The path is correct and the file is where the PR says it is.

**Four of five annotations check out against upstream source, not just plausibility:**

- `replicas` — `pkg/api/v1alpha1/workerpool_types.go:55–58` ("number of worker pods to run"); the "2 pods, 3 actors, must suspend at least one" reading is upstream's own header comment.
- `pauseImage` — `actortemplate_types.go:302–311` ("the container to use as the root sandbox container"), and upstream's suggested value is the exact digest used.
- `snapshotsConfig.location` — `actortemplate_types.go:269–273` ("Location to store snapshots in"); "resume onto a different pod" is supported by `docs/glossary.md` on `atenet` ("resumes suspended Actors on demand and routes traffic to the right worker pod").
- `ateomImage` — `docs/glossary.md`: "**ateom**: the coordinator that runs inside each worker pod and drives the sandbox runtime on behalf of atelet." "Interior helper image each pod runs" is right; "to perform checkpoint/restore" narrows slightly (atelet is what "streams snapshots to and from snapshot storage"). Not worth blocking.

The fifth is MINOR 2 above.

**All five of the author's mutations reproduce exactly as described**, plus seven I invented — twelve total against `test/validate-plugin.nu dagu`, restoring `marketplace.json` between each and confirming a clean tree afterward:

| # | Mutation | Result |
|---|---|---|
| M1 | delete marketplace `description` | new error fires, exact text |
| M2 | pristine | clean, exit 0 |
| M3 | reverse keywords (same set) | clean — no false mismatch |
| M4 | extra keyword in marketplace | genuine mismatch fires |
| M5 | description deleted **and** keywords mismatched | **both** surface; reports, does not abort |
| M6 | marketplace missing one keyword (strict subset) | fires |
| M7 | keywords `[]` on marketplace side | fires as "missing" |
| M8 | keywords `null` | fires as "missing" |
| M9 | description `""` | fires as "missing" |
| M10 | description `null` | fires as "missing" |
| M11 | keyword duplicated on marketplace side only | **fires** — sorted-list, not true set |
| M12 | one keyword value replaced, same length | fires |

Probing the set-comparison concern specifically: no realistic drift shape is now invisible that previously was. The only newly-tolerated difference is pure reordering, which is the intended relaxation.

**`test/validate-plugin.nu` IS a thin wrapper** — 19 lines, resolves the repo root and shells out to the skill script with `--marketplace`. One implementation, no second copy. The fix landed in the right place; it is the *modes* that are the problem, not the file.

**The zig dedupe holds.** The pointer target exists and supplies exactly the context the author said was missing: `build/references/build.md:3` opens "Documents Zig 0.15+. The `root_module = b.createModule(...)` form below is...", and lines 34–45 carry the full test-step block the fenced duplicate was lifted from. The link uses the house adjacent-qualifier form (`/zig:build` immediately before `references/build.md`) and both the `links` and `ref_depth` checks pass. The remaining prose names all three calls and what each does, so a reader who never opens `build.md` still gets the concepts — they lose copyable code, which is a defensible trade for a reference file.

**The baseline shrink is honest.** `git diff origin/main -- test/quality-baseline.json` removes exactly the two named entries and nothing else — no counts lowered elsewhere, no other keys touched; 49 → 47 confirmed by `jq '.allowed_failures | length'` on both sides. Both entries genuinely pass rather than being suppressed: `agent-substrate-overview` now scores **17/17** with `examples` absent from its failed column, and `dupe/22d40d6a` is gone from the duplicate-group list (3 groups remain: `1182fd06`, `28ccd32d`, `639413bb`).

**Restraint: clean.** All ten changed files map to the three bees plus their version bumps. Nothing out of scope, no speculative abstraction, no new dependency.

**`mise run test` exits 0** on this branch. **gitleaks** (`mise exec gitleaks@latest`): 21 commits, ~3.87 MB, no leaks. **Disclosure lint** passes — the new YAML carries no RFC1918 literals, no `op://` paths, no estate hostnames.

## Could not verify

- Whether any consumer outside this repo invokes the validator in direct-file mode. The impact is bounded by the two documented invocations I did test.
- Whether the `ateom` checkpoint/restore attribution is precisely right — upstream's glossary distinguishes `ateom` (drives the sandbox runtime) from `atelet` (streams snapshots), but no doc states plainly which one executes the checkpoint.

vinnie357:
## Gate 3 re-verification at `35d76d6` — APPROVE

Both blockers are closed and all three minors are correctly fixed. Re-verified in a private clone against `origin/main` = `cfd02dc`; the shared working tree was never checked out. Per scope, I did not re-review the example provenance or the zig dedupe — both were verified against upstream in the first pass and are unchanged.

One new one-byte nit found (trailing newline), detailed at the end. It does not block.

---

### BLOCKER 1 — CLOSED

**The flag is passed on the local path, not merely declared.** `--has-marketplace-context` is declared at line 193 and appears at exactly three other places:

- line 91 — the **local-source** branch of `validate-from-marketplace`: passed
- line 89–90 — the `$is_ext` branch: **not** passed
- line 128 — `validate-plugin-file` (direct mode): **not** passed

Both agreement blocks are gated on it (lines 278 and 448). That is the shape the fix claims.

**All four modes re-run side by side against `origin/main`'s validator**, same files, same invocations that surfaced the regression:

| Mode | PR `35d76d6` | `origin/main` |
|---|---|---|
| 1. Direct-file (`.../scripts/validate-plugin.nu plugins/languages/zig/.claude-plugin/plugin.json`) | 0 errors, 8 warnings, exit 0 | 0 errors, 8 warnings, exit 0 |
| 2. External / GitHub-object (`test/validate-plugin.nu claudio`) | 0 errors, "✓ Plugin 'claudio' is valid!" | identical |
| 3. Local marketplace, pristine (`test/validate-plugin.nu dagu`) | 0 errors, valid | identical |
| 4. Local marketplace, mutated | see below | — |

Modes 1 and 2 are **byte-identical** to main, including the same eight pre-existing skill-path warnings. The regression is gone.

**The feature is still live — a default-false flag did not silently disable it.** I re-ran the full twelve-mutation battery against `test/validate-plugin.nu dagu`, restoring `marketplace.json` between each and confirming a clean tree afterward. **All twelve behave identically to the pre-fix run:**

| # | Mutation | Result |
|---|---|---|
| M1 | delete marketplace `description` | fires, exact text |
| M2 | pristine | clean, exit 0 |
| M3 | reverse keywords (same set) | clean — no false mismatch |
| M4 | extra keyword | mismatch fires |
| M5 | description deleted **and** keywords mismatched | **both** surface; reports, does not abort |
| M6 | strict subset | fires |
| M7 | keywords `[]` | fires as "missing" |
| M8 | keywords `null` | fires as "missing" |
| M9 | description `""` | fires as "missing" |
| M10 | description `null` | fires as "missing" |
| M11 | keyword duplicated on one side only | fires — sorted-list, not true set |
| M12 | value replaced, same length | fires |

### BLOCKER 2 — CLOSED

Both documents now match observed behaviour in all four modes:

- **GitHub-object skip statement retained** — true again, confirmed by Mode 2.
- **Direct-file mode explicitly documented** as skipping the comparison for the same reason (reference section) — confirmed by Mode 1. This is stated more completely than the code was before; good.
- **"absent from either side is not a mismatch" is gone**, replaced by the actual rule: marketplace-side omission flagged when plugin.json defines the field; plugin.json-side omission not compared, leaving only the separate "Missing recommended field" warning. Confirmed against M1/M7–M10 and against the `is-not-empty` gate on `$pj_description` / `$pj_keywords`.
- **Both new error templates quoted verbatim** — I diffed the documented strings against the emitted ones character by character; `marketplace.json entry is missing '<field>' that plugin.json defines: ...` and both `... mismatch — plugin.json is authoritative: ...` forms match exactly.
- **Sorted-list semantics documented** in both files.

Non-blocking observation: the docs say a marketplace entry that "omits" the field is flagged, but the code uses `is-empty`, so a *present but empty* `""` or `[]` is reported as missing too (M7, M9). That is the right behaviour; the wording just undersells it.

### MINORS — all three correctly fixed

1. **`ACTOR_NAME`/`TASK`** — now reads "differing only in `metadata.name` and the `ACTOR_NAME`/`TASK` env values." Correct.
2. **`workerSelector`** — I re-fetched `pkg/api/v1alpha1/actortemplate_types.go` and checked the new wording against the source rather than against my own earlier note. The rewrite matches it on every point: `*metav1.LabelSelector` (exact type), `+optional`, "restricts which worker pools actors from this template may use" → "restricts which `WorkerPool`s this template's actors may schedule onto", "The scheduler only considers pools whose labels match this selector" → "by matching pool `metadata.labels`", and both quoted clauses ("if nil, all pools are eligible", narrow-never-expand) are accurate to the doc comment. The wrong Service analogy is gone and replaced with the correct Deployment/NetworkPolicy framing plus an explicit contrast to a Service's bare `spec.selector` map. `sources.md` now cites the Go type definition, and adds a blob sha for the YAML — stronger provenance than before.
3. **"SORTED LISTS"** — the comment now says `sort` doesn't dedupe and that this is stricter than a set in the safe direction. That is exactly what M11 demonstrates.

### Clobber-residue audit — CLEAN, verified plugin-by-plugin

I did not assume this. I normalised both manifests into name-keyed objects and diffed **all 31 entries in full** (every field, not just versions):

```
$ diff <(jq -S '.plugins|map({(.name):.})|add' main.json) <(jq -S ... pr.json)
   agent-sandboxing  0.2.9  -> 0.2.10
   claude-code       0.2.46 -> 0.2.48
   core              0.2.49 -> 0.2.48
   zig               0.1.5  -> 0.1.6
```

Four version lines, nothing else — every description, keyword array, category, and source is identical to main across all 31 plugins.

The `core` line looked like surviving residue, so I checked whether the PR *changes* it. It does not: the three-dot diff (`merge-base a8112e96...HEAD`) touches exactly four version lines — all-skills, claude-code, agent-sandboxing, zig. `core` is absent from the PR's own diff; the 0.2.48 in the branch is staleness inherited from the merge-base, which main moved past in #198. `elixir` does not differ at all.

Confirmed by trial merge onto `cfd02dc` in my clone — auto-merged cleanly, and post-merge:

```
all-skills 0.4.99   claude-code 0.2.48   core 0.2.49
elixir 0.1.21       agent-sandboxing 0.2.10   zig 0.1.6
```

**#193's core 0.2.49 and #194's elixir 0.1.21 both survive intact.** No residue of the whole-file `checkout --ours/--theirs` resolution reached this head.

### Rebase scope

Nothing beyond version fields. The trial merge required no conflict resolution at all.

### NEW NIT — `.claude-plugin/plugin.json` lost its trailing newline

Almost certainly residue from the manual conflict fix-up, and the one thing the plugin-by-plugin audit surfaced that the version comparison would not:

```
main: 0a 7d 0a   ( ...\n}\n )
PR:   5d 0a 7d   ( ...]\n}  )  ← \ No newline at end of file
```

Only this file is affected — no other file in the diff carries a `\ No newline at end of file` marker. Cosmetic: the JSON parses and every check passes. Worth restoring opportunistically, since `mise update-all-skills` regenerates this file and a newline-only diff will otherwise show up as churn on the next run. You are editing this file at merge time for the all-skills number anyway.

### Gates

- `mise run test` — **exit 0**; "✨ All plugins are valid!"; 47 debt/bug entries, unchanged from the previous head.
- `gitleaks` (`mise exec gitleaks@latest`) — 30 commits, ~3.99 MB, **no leaks**.
- Scope: 12 files, all mapping to the three bees, the two doc corrections, and version bumps. Nothing out of scope.

Versions deliberately not flagged per the re-review brief.

**Verdict: APPROVE.**

PR #201, opened 2026-08-01
docs(core): add anti-fabrication pointers to house-convention skills (claude-skills-144)

vinnie357:
- claude-skills-144 (part a of the approved split): adds an `## Anti-fabrication` section to 4 core skills — documentation, code-review, and twelve-factor are genuinely low-risk / house-convention; **git is not** — it's the one skill in this set making concrete, checkable claims about Git/GitHub API behavior, and its section is treated accordingly: both claims verified against real sources (one cited to GitHub's own docs with a verbatim quote, one honestly labeled as empirically observed rather than documented, after a search turned up no GitHub page stating it as a general principle). Each section names the skill's actual verifiable-claim area (or states honestly that it has none beyond convention) rather than a generic pointer, matching the `rust:async` shape
- regenerates `test/quality-baseline.json` (shrink-only) to drop the 4 now-passing `anti_fab` entries — 49 → 45 debt entries
- does NOT touch the other 40 flagged skills — see the assessment below for why a blanket sweep was rejected
- version bump intentionally omitted — holding per operator's one-at-a-time merge sequencing (core contested by #194/#198 ahead of this PR)

Assessment (claude-skills-144, posted to the bee as `## BEES REQUESTS`): re-measured the anti_fab count directly via the validator (44 today, not the stale 44/45 figures on the bee — the count has now moved 44 → 45 → 44 across three measurements), confirmed the sources.md half is still fully done (28/28 plugins), and triaged the 44 anti_fab failures into 4 true house-convention skills (this PR) vs 40 genuinely high-risk skills where a pointer would be decorative — two structural findings drove the split: `core/bees` already carries 6 verified-against-0.4.0 citations and still fails the check (false negative), and the check never scans `references/`, so a skill's real claims can sit unscanned one level down — the 7 zig skills (18–30-line pointer files) and `wit` (496/500, not pointer-style at all) all route real checkable content through `references/` regardless of SKILL.md size.

Gate-3 review caught that an earlier draft's citation for git's claims was itself unsourced — fixed by fetching and quoting GitHub's actual docs rather than asserting a plausible URL. The reviewer independently searched for documentation of the SSH-bypasses-scope half and found none either — confirming the skill's own hedge (labeled "empirically observed, not centrally documented") was the right call, not excess caution.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.101",
+      "version": "0.4.102",
       "category": "meta",
       "keywords": [
         "skills",
@@ -143,7 +143,7 @@
       "source": "./plugins/core",
       "strict": true,
       "description": "Essential development skills: Git, TDD, code review, gitleaks secret scanning, documentation, restraint, anti-fabrication, twelve-factor apps, the agent loop, bees issue tracking, mise, nushell, and Apple Container",
-      "version": "0.2.49",
+      "version": "0.2.50",
       "category": "development",
       "keywords": [
         "git",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.101",
+  "version": "0.4.102",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/core/.claude-plugin/plugin.json
+++ b/plugins/core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "core",
-  "version": "0.2.49",
+  "version": "0.2.50",
   "description": "Essential development skills: Git, TDD, code review, gitleaks secret scanning, documentation, restraint, anti-fabrication, twelve-factor apps, the agent loop, bees issue tracking, mise, nushell, and Apple Container",
   "author": {
     "name": "vinnie357"
--- a/plugins/core/skills/code-review/SKILL.md
+++ b/plugins/core/skills/code-review/SKILL.md
@@ -17,6 +17,17 @@ Activate when:
 - Suggesting refactoring improvements
 - Checking adherence to coding standards
 
+## Anti-fabrication
+
+This skill follows `core:anti-fabrication`, but genuinely has little to verify against an
+external source: the checklist, review-etiquette, and anti-pattern content is house
+convention and general engineering practice (Google's review guide is cited in
+`sources.md` for the checklist's shape), not versioned claims about a specific tool or
+API. The "Language-Specific Considerations" section is broad, stable idioms (pattern
+matching, ownership, type hints) rather than anything that drifts release to release —
+still, don't assert a language- or framework-specific rule this skill doesn't cover
+without checking that language's own style skill first.
+
 ## Code Review Checklist
 
 ### 1. Correctness and Functionality
--- a/plugins/core/skills/code-review/SKILL.md
+++ b/plugins/core/skills/code-review/SKILL.md
@@ -19,14 +19,12 @@ Activate when:
 
 ## Anti-fabrication
 
-This skill follows `core:anti-fabrication`, but genuinely has little to verify against an
-external source: the checklist, review-etiquette, and anti-pattern content is house
-convention and general engineering practice (Google's review guide is cited in
-`sources.md` for the checklist's shape), not versioned claims about a specific tool or
-API. The "Language-Specific Considerations" section is broad, stable idioms (pattern
-matching, ownership, type hints) rather than anything that drifts release to release —
-still, don't assert a language- or framework-specific rule this skill doesn't cover
-without checking that language's own style skill first.
+This skill follows `core:anti-fabrication`, but has little to verify against an external
+source: the checklist and etiquette content is house convention (Google's review guide,
+cited in `sources.md`, for the checklist's shape), and "Language-Specific
+Considerations" is broad, stable idiom rather than versioned API claims. Don't assert a
+language-specific rule this skill doesn't cover without checking that language's own
+style skill first.
 
 ## Code Review Checklist
 
--- a/plugins/core/skills/documentation/SKILL.md
+++ b/plugins/core/skills/documentation/SKILL.md
@@ -17,6 +17,15 @@ Activate when:
 - Creating architecture or design documents
 - Writing changelogs or release notes
 
+## Anti-fabrication
+
+This skill follows `core:anti-fabrication`. Most of it is house convention — README
+structure, ADR templates, changelog format, comment style — not claims that can go stale
+against an upstream. The one real claim area is the "Documentation Tools" section's
+generator commands (`mix docs` for ExDoc, `cargo doc` for rustdoc, Sphinx, MkDocs, JSDoc,
+TypeDoc): verify the invocation against the tool's current docs before asserting it,
+rather than repeating it from memory — these are stable but not immune to CLI changes.
+
 ## README Files
 
 ### Essential README Structure
--- a/plugins/core/skills/git/SKILL.md
+++ b/plugins/core/skills/git/SKILL.md
@@ -7,6 +7,17 @@ description: Guide for Git operations including commits, branches, rebasing, and
 
 Activate when creating commits, managing branches, creating pull requests, resolving conflicts, or following Git workflows.
 
+## Anti-fabrication
+
+This skill follows `core:anti-fabrication`. The commit/PR format rules are house
+convention, but the "Remote and Authentication Conventions" section makes specific
+behavioral claims about Git and the GitHub API — SSH key auth bypassing OAuth scope
+checks that HTTPS push enforces, and anonymous access to GitHub Releases on a private
+repo returning 404 rather than 401 — verified against Git and GitHub documentation cited
+in `sources.md`. Don't assert `git`/`gh` CLI or GitHub API behavior this skill doesn't
+already cover without checking the current docs; CLI flags and API responses do change
+across versions.
+
 ## Commit Format
 
 Follow the [Conventional Commits](https://www.conventionalcommits.org/) specification:
--- a/plugins/core/skills/git/SKILL.md
+++ b/plugins/core/skills/git/SKILL.md
@@ -10,13 +10,25 @@ Activate when creating commits, managing branches, creating pull requests, resol
 ## Anti-fabrication
 
 This skill follows `core:anti-fabrication`. The commit/PR format rules are house
-convention, but the "Remote and Authentication Conventions" section makes specific
-behavioral claims about Git and the GitHub API — SSH key auth bypassing OAuth scope
-checks that HTTPS push enforces, and anonymous access to GitHub Releases on a private
-repo returning 404 rather than 401 — verified against Git and GitHub documentation cited
-in `sources.md`. Don't assert `git`/`gh` CLI or GitHub API behavior this skill doesn't
-already cover without checking the current docs; CLI flags and API responses do change
-across versions.
+convention, but the "Remote and Authentication Conventions" section makes specific,
+checkable claims about Git and the GitHub API — this is the one skill in this set that
+does, and its two claims are verified two different ways:
+
+- **Anonymous access to GitHub Releases on a private repo returns 404, not 403** (not
+  401 — corrected from an earlier draft of this section). Confirmed against GitHub's own
+  documentation, cited in `sources.md`: "GitHub uses a 404 Not Found response instead of
+  a 403 Forbidden response to avoid confirming the existence of private repositories."
+- **SSH key auth bypasses the `workflow` OAuth scope that HTTPS push enforces.** GitHub
+  documents the `workflow` scope requirement for OAuth apps/PATs pushing Actions workflow
+  files (cited in `sources.md`), but does not document the SSH-side exemption as a general
+  principle — that half is empirically observed, not centrally documented: pushing a
+  `.github/workflows/*.yml` change over HTTPS with a token lacking `workflow` scope is
+  rejected ("Refusing to allow an OAuth App to create or update workflow ... without
+  `workflow` scope"); the identical push over SSH succeeds.
+
+Don't assert `git`/`gh` CLI or GitHub API behavior this skill doesn't already cover
+without checking the current docs, or testing it directly when the docs don't say —
+CLI flags and API responses do change across versions.
 
 ## Commit Format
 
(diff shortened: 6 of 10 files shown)


vinnie357:

**Verdict: REQUEST CHANGES** — narrow. One factual defect in the `core/git` paragraph. Everything else I attacked survived, and the 4/40 split is correct and well-argued.

All findings below were re-derived in a disposable clone (`git clone --reference`, never the shared tree). I inherited no numbers from the PR body or the bee.

---

### What I verified independently

| Claim | Result |
|---|---|
| Baseline 49 → 45 | **Confirmed.** main = 49 debt / 44 `anti_fab`; branch = 45 / 40. Exactly 4 fixed. |
| Exactly 4 keys removed, none added | **Confirmed.** `comm` on the key sets: 4 removed, 0 added; `_comment` byte-identical; surviving 45 entries byte-identical under `jq -S`. Nothing silently lowered. |
| The 4 skills now pass | **Confirmed.** All four score **17/17**, no failed checks. |
| The 4 genuinely failed before | **Confirmed.** Zero `fabricat`/`anti-fabrication` matches in all four on `origin/main` — the baseline entries were not stale, the new paragraphs carry the pass. |
| Restraint / scope | **Confirmed.** Diff is exactly 5 files (42 insertions, 28 deletions). No edits to the other 40 skills, none to the validator. |
| Line caps | **Confirmed.** git 233, twelve-factor 273, documentation 363, code-review 429 — all under 500. |
| No `allowed-tools`, `ref_depth` unaffected | **Confirmed** (both are part of the 17/17). |
| Version files untouched | **Confirmed** — expected, per operator sequencing. |
| Full suite | `mise run test` **exit 0**. `gitleaks detect`: **no leaks found** (21 commits, 3.88 MB). |

---

### The structural argument for the 4/40 split — verified, and stronger than stated

Both premises hold, and I added a third:

1. **`anti_fab` never scans `references/`** — confirmed at `test/validate-skills-quality.nu:2251-2258`. The check tests `$content` / `$content_lower` (SKILL.md body only). `refs` is read at ~line 2237 but consumed solely by checks 8, 9, and 16.

2. **`core/bees` is a false negative** — confirmed, and **understated**. It fails `anti_fab` (16/17) while carrying **six** explicit "Verified against bees 0.4.0" citations (`SKILL.md` lines 85, 109, 144, 157, 181, 368), not four.

3. **(mine) The check cannot distinguish substance from decoration.** I replaced the entire `core/git` paragraph with a bare `## Anti-fabrication` header and nothing else. Result:

   ```
   │  21 │ git │ core │ 224/500 │ 17/17 │  │  │  │
   ```

   An empty header scores a perfect 17/17. A 44-file mechanical sweep would therefore have produced 44 green checks and zero examined claims. (File restored; tree clean.)

**The split is correct.** Point 3 is the decisive one: since the validator cannot verify the substance, the substance the author chose to write *is* the entire deliverable — which is precisely why the one inaccurate sentence below matters.

---

### Findings

#### 1. [blocking] `core/git`: the anti-fabrication paragraph fabricates its own citation

`plugins/core/skills/git/SKILL.md` — the new paragraph states the two auth claims are:

> verified against Git and GitHub documentation cited in `sources.md`

**No GitHub documentation is cited in `sources.md` for the git skill.** The git skill has exactly two sources, in both `sources.md` (lines 7-18) and `sources.toml` (lines 12-37):

- `https://git-scm.com/doc` — Key Topics: *"Git operations, branching, rebasing, conflict resolution"*; its own `sources.toml` note says *"no version the skill content is verified against."*
- `https://www.conventionalcommits.org/` — commit message format.

Neither covers GitHub's OAuth scope enforcement or GitHub Releases HTTP status codes. Every other GitHub URL in `sources.md` belongs to a different skill (mise, nushell, gitleaks, bees, apple/container, ponytail, allium).

**The underlying claims are true — only the provenance sentence is false.** I verified both independently:

- **404-not-401 on private repos:** empirically confirmed. Anonymous `GET https://api.github.com/repos/vinnie357/runex/releases` → **404**, identical to a repo that does not exist. GitHub documents this explicitly: *"GitHub uses a `404 Not Found` response instead of a `403 Forbidden` response to avoid confirming the existence of private repositories."*
- **SSH bypassing OAuth scope checks:** corroborated — the canonical case is pushing `.github/workflows/*` over HTTPS with a token lacking `workflow` scope; SSH keys carry no OAuth scopes, so the check does not fire.

This is the one defect that matters here, because it is the exact failure mode the PR exists to prevent, shipping inside the section that announces the discipline.

**Suggested fix** (either is fine, both are small):

- **Preferred** — add the two real sources to `plugins/core/skills/sources.md` *and* `sources.toml` under `skills = ["git"]`:
  - `https://docs.github.com/en/rest/overview/troubleshooting-the-rest-api` (404-not-403 for private resources)
  - `https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps` (`workflow` scope)

  This is a net improvement: those claims were already uncited on `main`, so the PR surfaced a pre-existing gap rather than creating one.
- **Or** reword to not assert `sources.md` coverage.

*Nit within this finding:* the skill says "404 rather than **401**"; GitHub's docs frame it as 404 instead of **403**. The substance is right; the contrasted code is not the documented one.

#### 2. [suggestion] The PR body's triage label is imprecise for `git`

The body describes all four as *"genuinely low-risk / house-convention."* That fits `documentation`, `code-review`, and `twelve-factor`. It does not fit `git` — `git` is the one skill in this set that makes concrete, checkable claims about external tool behavior, which is why finding 1 was possible at all.

The *treatment* is right (the paragraph names the claims specifically rather than waving at them); only the label is loose. No content change needed — flagging so the framing doesn't get reused as precedent for the remaining 40.

#### 3. [suggestion] Forward-looking: two of the remaining 40 cannot take an 11-line paragraph

For whoever picks up the other 40: an ~11-line section needs the body at ≤489 lines.

- `wasm/wit` — **496/500**. Adding this paragraph breaches the cap (507). Needs a references split first.
- `slidev/interactive` — 485/500 → 496. Fits, barely.

Also, the assessment's characterization of *"the 7 zig and 2 wasm skills as 19–81-line pointer files"* does not hold for `wasm/wit` at **496** lines — that one is a full skill, not a pointer file. (The zig floor is 18, not 19.) Doesn't change the assessment's conclusion; noting it so the number isn't inherited.

---

### On `core/code-review` specifically

Asked to probe whether this paragraph earns its place or is a decorative gesture. **It earns it, and it is honest — but it is the weakest of the four.**

Honest, not empty:
- The Google citation is real and checkable — `https://google.github.io/eng-practices/review/` is in `sources.md` under `## Code Review Skill` and in `sources.toml` line 98. This paragraph is the *only* one of the four whose citation I could confirm.
- Its self-assessment of "Language-Specific Considerations" is accurate. It names pattern matching, ownership, and type hints as stable idioms; all three are present (`SKILL.md` L268 Elixir pattern matching, L294 Rust ownership, L286 Python type hints), and none is version-gated.
- It carries a real operative constraint: *"don't assert a language- or framework-specific rule this skill doesn't cover without checking that language's own style skill first"* — a genuine delegation to `rust:*` / `elixir:style` / `zig:*`, written generically so it creates no maintenance edge.

The weakness is proportion. The load-bearing content is that last clause; the preceding five lines are justification prose, and `/core:restraint` is pointed about this — *"If the explanation is longer than the code, delete the explanation."* I would not block on it: for an anti-fabrication section specifically, stating the scope boundary **is** content, because the boundary is what stops a future author from casually adding versioned claims to a skill that has none. An honest "not much here" was the correct outcome and this delivers it.

---

### What I could not verify

- **Whether the remaining 40 are all genuinely high-risk.** I verified the two structural premises and spot-checked line counts; I did not read all 40 SKILL.md bodies. The split's *reasoning* is sound and independently confirmed — its per-skill application to all 40 was out of scope for this review.
- **The SSH/OAuth-scope claim against a first-party GitHub docs page.** Confirmed via GitHub's own error text and multiple corroborating sources, not a single citable doc paragraph. The 404 claim I confirmed both empirically and from GitHub's docs.

### Version state (re-measured — main moved mid-review)

`main` is now **`a8112e9`** and carries **core 0.2.48** / **all-skills 0.4.98**, matching in `plugin.json` and `marketplace.json`. (An earlier draft of this review read 0.4.97; that was accurate at `72cf8ca` and went stale when main advanced.)

This PR touches **no version files** — confirmed, the diff is 4 SKILL.md files plus `test/quality-baseline.json`. `Check version bumps` red is expected per operator sequencing and is not a finding.

Re-verified against the new `main`: baseline still shows exactly the 4 keys removed, 0 added, and `main`'s count is 49. Main advancing did not disturb this PR.

**One bookkeeping discrepancy, for the merge sequencer:** PR **#197 currently holds all-skills `0.4.98`** (and claude-code `0.2.47`) — not `0.4.100`. `0.4.98` is already taken by `main`, so #197 needs re-bumping before it lands. #198 holds core `0.2.49` / all-skills `0.4.99` as expected. Given that, #201's targets of **core 0.2.50** and **all-skills 0.4.101** hold — but only if #197 is re-bumped to `0.4.100` first.

vinnie357:
## Gate 3 — re-verification at `d381504`

**Verdict: APPROVE.** The blocking finding is properly fixed. The citation is real and verbatim, and the SSH hedge is honest — I tried to overturn it by finding first-party documentation and could not.

Re-derived in a disposable clone against `origin/main` = `cfd02dc`. Nothing inherited from the PR body or the previous round.

---

### 1. The 404/403 citation — sound

| Check | Result |
|---|---|
| URL resolves | **200**, no redirect: `docs.github.com/en/rest/using-the-rest-api/troubleshooting-the-rest-api` |
| Quote verbatim | **Confirmed.** Page source: `GitHub uses a <code>404 Not Found</code> response instead of a <code>403 Forbidden</code> response to avoid confirming the existence of private repositories.` The skill reproduces it exactly, with the `<code>` markup flattened to plain text. |
| Section heading | **Present** — "404 Not Found for an existing resource", as the `sources.toml` note claims |
| `sources.md` entry | Well-formed; URL / Purpose / Date Accessed / Key Topics match the file's existing shape |
| `sources.toml` entry | Well-formed; `test:sources` passes (74 self-test cases, 28 plugins) |
| Header index updated | Yes — both new entry names added to the `sources.toml` pointer line at the top of `sources.md` |

The 401 → 403 correction is right: GitHub's contrast is with 403, and that is what the skill now says.

### 2. The SSH claim — the hedge is correct, and necessary

This was the part to scrutinise hardest. It holds.

**(a) `github-oauth-scopes` is not being stretched.** The page genuinely documents the scope, verbatim: *"Grants the ability to add and update GitHub Actions workflow files."* That is exactly what the `sources.toml` note quotes, and the note explicitly limits the citation to *"only the HTTPS/OAuth-token side."* No overreach.

**(b) The skill text does not imply documentation it lacks.** It states the opposite outright — *"does not document the SSH-side exemption as a general principle — that half is empirically observed, not centrally documented."* A reader cannot mistake the SSH half for a cited claim.

**(c) The three records agree and cannot drift into disagreement.** All three independently say the same thing:

- **skill body**: "does not document the SSH-side exemption as a general principle — that half is empirically observed, not centrally documented"
- **`sources.toml` notes**: "No GitHub docs page found stating the SSH-side exemption as a general principle; that half is described in the skill as observed behavior ..., not as a citable claim"
- **`sources.md` Purpose**: "the SSH-side exemption itself is not documented on this or any GitHub page found, and is described in the skill as empirically observed, not cited, for that reason"

**I could not find a first-party doc stating the SSH exemption — so the hedge stays.** I grepped both cited pages (`scopes-for-oauth-apps` contains **zero** SSH mentions) and ran two targeted searches including a `site:docs.github.com` restriction. Nothing states it.

The search surfaced the *opposite* nuance, which strengthens the case for hedging: SSH keys created by OAuth apps **do** lose repository access when an org enables OAuth App access restrictions. That is a different mechanism from token scopes, but it means "SSH bypasses OAuth" is not a blanket principle GitHub endorses — so asserting the exemption as documented would have been wrong in a second, subtler way. The author's decision to label it empirical is the correct call, not excessive caution.

### 3. Triage rewording and the `code-review` trim

**Skill text — corrected.** `git`'s paragraph now opens by naming itself as *"the one skill in this set that does [make checkable claims]"* and splits the two claims into a bullet each with their differing evidentiary status. That is the corrected framing.

**`code-review` trim — the operative clause survived.** Trimmed 429 → 427 lines, and the scope boundary is intact at line 27: *"Don't assert a language-specific rule this skill doesn't cover without checking that language's own style skill first."* The trim removed justification prose, not payload — exactly the right cut.

**PR body — NOT updated.** See finding A below.

### 4. Mechanical confirmations

- **Baseline untouched by the fix commit**: `git diff d381504^ d381504 -- test/quality-baseline.json` is **empty**. The author's reasoning holds — these were refinements to already-passing skills.
- **Baseline vs current `main` (`cfd02dc`)**: still exactly **4 removed, 0 added** (main 49 → branch 45).
- **`mise run test`: exit 0**, including `test:sources` with both new entries.
- **`gitleaks detect`: no leaks found** (32 commits, 4.01 MB).
- **All four skills 17/17**, all under the cap: git 245, twelve-factor 273, documentation 363, code-review 427.
- **No version files touched** — diff is 6 content files + baseline. `Check version bumps` red is expected per operator sequencing and is not a finding.

---

### Findings (all non-blocking)

**A. [suggestion] The PR body still carries the framing the skill text just corrected.** It reads:

> the 4 core skills identified as genuinely low-risk / house-convention — git, documentation, code-review, twelve-factor

That is the exact label round 1 flagged, with `git` still inside it. The shipped artifact is correct and the body is merge-time metadata, so this changes nothing for users — but it was asked for and is not done. Worth a one-line edit so the rationale in the record matches the rationale in the skill.

Relatedly, the body's assessment paragraph still describes *"the 7 zig + 2 wasm pointer-style skills."* The "19–81-line" phrasing is gone, but `wasm/wit` at **496/500** lines is not a pointer-style skill by any reading.

**B. [nit] Shipped skill content carries its own revision history.** In `git/SKILL.md`:

> returns 404, not 403 **(not 401 — corrected from an earlier draft of this section)**

A reader loading this skill has no access to the earlier draft, so the parenthetical is reviewer-facing noise in a body that costs context on every activation. The git history and this PR already record the correction. Suggest dropping the parenthetical and keeping "returns 404, not 403".

**C. [nit, no action] A carve-out in the cited scopes page.** That page also states workflow files *"can be committed without this scope if the same file (with both the same path and contents) exists on another branch in the same repository."* The skill says "pushing a `.github/workflows/*.yml` **change**", which sidesteps the carve-out correctly. Noting only so it isn't discovered later and mistaken for an error.

### What I could not verify

**I did not reproduce the empirical SSH observation.** Confirming it directly requires pushing a workflow-file change with a deliberately scope-limited token and then over SSH — a mutating operation against a real repo with credentials I do not have. The skill labels this claim as empirically observed rather than cited, which is the honest framing for a claim in that state; I verified the *documented* half and the *labelling*, not the observation itself. Whoever reproduces it should record the result in the `sources.toml` note.

Beyond that, nothing blocked me: both URLs fetched cleanly via curl, and the full suite ran locally.

PR #202, opened 2026-08-01
fix(skills): dedupe cross-skill duplicate blocks in elixir and github plugins

vinnie357:
- Remove `plugins/languages/elixir/skills/phoenix/references/tidewave.md` (near-full duplicate of `elixir:tidewave`'s SKILL.md); point `phoenix/SKILL.md` at `/elixir:tidewave` only
- Port the Cursor/VS Code/Zed editor table (lost with the removed file) into `tidewave/references/mcp-setup.md`, flagged unverified against current upstream in `sources.toml` — Tidewave's docs site ships only Claude Code, Codex, and opencode pages
- Collapse the duplicated "Two validators to run" block in `pr-review/references/registration.md` into a cross-skill pointer at `/github:dependabot-consolidator`'s `references/registration.md`
- Regenerate `test/quality-baseline.json` (shrink-only) to drop the two now-fixed `duplicate_block` entries (`dupe/1182fd06`, `dupe/639413bb`)
- claude-skills-148 (bees-worker/beads-worker shared block): no code change — reasoning recorded on this PR (see comment below); tracker comment text is in the `## BEES REQUESTS` block for the lead to apply
- Bump elixir 0.1.21→0.1.22, github 0.3.10→0.3.11, all-skills 0.4.102→0.4.103 (rebased onto origin/main `9b6c80e` before reading current versions)


vinnie357:
## claude-skills-148 decision: no dedupe, recorded reasoning

*(Edited to correct two points a review round caught — see note at bottom.)*

Group `dupe/28ccd32d` (7 shared 8-line windows, `plugins/core/skills/bees/agents/bees-worker.md` vs `plugins/tools/beads/skills/beads/agents/beads-worker.md`) is left in place. Measured, not guessed:

**Shared content shape** — the 7 sliding windows collapse into two contiguous spans, **21 normalised lines total** (re-derived with the validator's own `normalise-dupe-lines` + window-merge, not eyeballed):
1. PR-mechanics span, **11 lines**: `git add` → `git commit` → `#### PR Creation` → `git push` → `gh pr create` → notify/stop → `#### After Merge` → `gh pr view` → `git checkout main && git pull && git branch -d`. Neither `bees`/`bd` token appears inside this span — the divergent tracker commands sit immediately outside it (`bees close`/`bd close` before, `bees ready`/`bd ready` after).
2. Pattern/Action table span, **10 lines**: the `Create/Add/Implement → Write new files` etc. table plus "For exploration tasks, delegate:" / `Task(subagent_type="Explore"...)`. Not git-related at all.

**Why preloading doesn't pay for itself — and why it's worse than that:**
- `/core:git` SKILL.md is 222 lines (measured: `wc -l plugins/core/skills/git/SKILL.md`). `skills:` frontmatter preloading is **eager** — per `/claude-code:claude-agents`: "The `skills` field preloads full skill content into the subagent's context at startup — not just the description shown during discovery." **Neither `bees-worker.md` nor `beads-worker.md` declares `skills:` today** (checked both frontmatter blocks) — so this isn't redirecting an existing cost, it's 222 lines of entirely new cost, paid on every single invocation regardless of whether that run ever reaches PR creation.
- Span 2 (Pattern/Action table) has no relationship to `/core:git` at all, so preloading it wouldn't address that half of the duplication either way. No existing skill owns this content; inventing a new skill to hold 10 lines fails restraint rung 1.
- **The sharper distinction** (this is the real reason, not "citations are inert"): `bees-worker.md` already cites `/core:agent-loop` in prose (lines 69/106/110) — and that works fine, because those citations are *conditional*: only the complexity-pipeline branch needs to follow them, so paying the read cost only on that branch is correct. The PR-mechanics span is different in kind — it's **main-path**, needed on every single run of both workers, not gated behind a branch. A citation is the right shape for optional content; it's the wrong shape for content every invocation needs.
- **Cross-plugin self-containment rules out both preloading and hoisting.** `bees-worker` ships in `core`; `beads-worker` ships in the separate `beads` plugin. Install only the `beads` plugin and a `skills: [core:git]` preload is a dangling reference to an uninstalled plugin — same failure mode a prose pointer to `/core:git` would have. This also rules out the option nobody named: hoisting the shared span into one file and pointing the other at it — that still requires both plugins present to resolve, which for an agent's main-path operating instructions is unacceptable.

**Verdict:** leave the duplication. `dupe/28ccd32d` stays in `test/quality-baseline.json` unchanged (7 windows) — no baseline edit needed since nothing was fixed.

---
*Edit: corrected span 2 from an eyeballed "~8 lines" to the validator-measured 10 (total 11+10=21, not ~19), and replaced the "prose citations are inert at spawn time" claim — `bees-worker.md` already cites `/core:agent-loop` successfully — with the real distinction: conditional content vs. main-path content, plus the cross-plugin self-containment argument that rules out preloading AND hoisting.*

vinnie357:
## Gate 3 — adversarial review: **APPROVE** (5 findings, none blocking)

Reviewed at `eb0bc4a` against `main` at `3aeaf7f`, in a throwaway clone (no checkout in the shared tree). I tried to break this on eight fronts; the change survived all of them. Every number below is re-derived, not inherited from the PR body.

---

### What I ran

| Check | Result |
|---|---|
| `nu test/validate-skills-quality.nu` on `main` | EXIT=0 — **4** dupe groups, **49** debt entries |
| same, on `eb0bc4a` | EXIT=0 — **2** dupe groups, **47** debt entries |
| `mise run test` (all 7 suites, 106 skills, 626 files) on `eb0bc4a` | EXIT=0 |
| `nu test/validate-plugin.nu elixir` / `github` | 0 errors, 0 warnings each |
| `nu test/check-version-bumps.nu` vs true `main` | fails on elixir + github (expected — see §5) |
| `mise exec gitleaks@latest -- gitleaks detect` | 324 commits, no leaks |
| `git merge-tree` vs `main` | 0 conflicts |

---

### Attacks that failed to land

**1. The deletion of `phoenix/references/tidewave.md` — no dangling references.**
`git grep 'references/tidewave'` over all tracked files returns nothing outside `.bees/issues.jsonl` (a historical tracker export describing the issue itself — correct that it still names the old path). `plugin.json` / `marketplace.json` hits on "tidewave" are the **skill dir** `./skills/tidewave` and the **agent** `./agents/tidewave.md`, both untouched and both still present. `sources.md`/`sources.toml` map the tidewave source entry to `skills = ["tidewave"]` and already name `tidewave/references/mcp-setup.md` as the extraction target — the deleted file was claimed by no source entry. `validate-sources` passes.

**2. Baseline is exactly what's claimed — nothing else moved.**
Entry-level diff of `allowed_failures`: 49 → 47, and the diff is *precisely* two removals with zero additions and zero field mutations (`git diff --numstat` = `0 14`). `dupe/28ccd32d` still `detail_count: 7` / `claude-skills-148`; `dupe/22d40d6a` still `1` / `claude-skills-150`. Both removed groups are **gone from the detected set**, not hiding under the 8-line window — the live scan on `eb0bc4a` reports only the two survivors by name. **No new group minted**; the before/after group sets differ by removal only, which is what the shrink-only ratchet requires.

**3. `ref_depth` on the new pointer — I broke it to prove the qualifier is load-bearing.**
I rewrote the `pr-review` pointer to the unqualified form (`see \`references/registration.md\``) and re-ran the validator:

```
│  53 │ pr-review │ github │ 156/500 │ 16/17 │ ref_depth │ ref_depth │ nested:registration.md │
  github/pr-review:ref_depth
```

That mints a **new** baseline key, which the shrink-only baseline cannot absorb — an unwaivable CI failure, not something `--update-baseline` swallows. The shipped `/github:dependabot-consolidator`'s references/registration.md` form resolves through `cross-skill-qualified` (target dir + path both exist) and scores 17/17. Tree restored; `git status` clean. The other pointer (`phoenix` → `/elixir:tidewave`) carries no path token at all, so it cannot trip `ref_depth`, and it resolves under the `invocations` check. **Both pointers are same-plugin** (`elixir`→`elixir`, `github`→`github`), so neither dangles for a user who installed only one plugin.

**4. Progressive disclosure for `phoenix` is not degraded — it improves.**
The pointer sits at the end of the Tidewave section, after the four inline essentials (dep, plug placement, `claude mcp add`, dev-only warning), and enumerates what's downstream ("full setup, MCP tools reference, CLI app, editor configs, LiveView annotations, troubleshooting"). A reader who follows it gets 338 lines (`tidewave/SKILL.md` 215 + `references/mcp-setup.md` 123) against the deleted file's 181 — including Codex CLI, Gemini CLI, and opencode setup, the standalone CLI app on :9832, and a decision table the old file never had.

**5. Restraint on the deletion — warranted.**
Trimming only the duplicated section would have left a stub whose *purpose* still duplicates a full skill in the same plugin, staler than the survivor. Rung 1 ("does this need to exist?") answers no. Deletion is the correct rung, not an over-reach. One caveat in Finding A below.

---

### Findings

**A. [suggestion] The Cursor / VS Code / Zed setup table died with the file and exists nowhere else in the repo.**
Verified: on `main`, `.cursor/mcp.json` / `.vscode/mcp.json` / `context_servers.tidewave.settings.url` appear only in the deleted file. On `eb0bc4a` they appear nowhere (the `linear` plugin's `context_servers` hit is a different tool's config). The survivor's `mcp-setup.md` covers Claude Code, Codex CLI, Gemini CLI, opencode — not those three.

Mitigating: `sources.toml` records upstream Tidewave as shipping exactly three editor pages (`mcp_claude_code`, `mcp_codex`, `mcp_opencode`), so the lost table was unsourced and may already be stale. I could not verify current upstream coverage for Cursor/VS Code/Zed from this environment.

Smallest fix if you want it back: port the three rows into `tidewave/references/mcp-setup.md`, which is the file that owns per-editor setup. Three lines is far under the 8-line window, so it cannot mint a dupe group. Not blocking — this is a judgment call about scope, and the deletion is defensible either way.

*Also dropped:* "Combine Bandit with Tidewave using a mix alias" for non-Phoenix projects. The survivor's CLI-app section covers that case more accurately (`tidewave_app`, port 9832, plus a "Non-Phoenix Elixir → CLI app" row). Superseded, not lost.

**B. [nit] The 148 span measurement is understated by 2 lines.**
Re-derived with the validator's own `normalise-dupe-lines` + `window 8` + md5, not by eye: **2 contiguous spans, 7 windows, 11 + 10 = 21 normalised lines** (not ~19). Span 1 is 11 — matches exactly, including that neither `bees` nor `bd` appears inside it. Span 2 is **10**, not ~8: the 7 table lines, the "For exploration tasks, delegate:" line, the `Task(subagent_type="Explore"...)` line, and the trailing `## Guidelines` heading. Does not move the verdict — span 2 is not a `/core:git` candidate, so the ratio is 222 : 11 either way.

**C. [question] "Filed as the closing note on claude-skills-148" is not corroborated by the tracked export.**
On this branch, `.bees/issues.jsonl` shows `claude-skills-148` at `status=open`, **0 comments**, `updated_at=2026-07-26`, and `.bees/` is untouched by the PR (`git diff --stat gh/main -- .bees/` empty). Reading it with the "see BEES REQUESTS on the parent task" clause, the write was *requested*, not performed — so "Filed as" claims a completed action.

This has practical teeth: the decision currently lives only in a GitHub PR comment. An agent that later runs `bees ready` sees 148 open with no note and re-derives the whole analysis. Either perform the tracker write before merge (148 is the decision's permanent home) or reword to "to be filed."

**D. [nit] "A markdown citation inside an agent file is inert at spawn time" overstates, and the file contradicts it.**
`bees-worker.md` already uses prose citations three times — `/core:agent-loop` at lines 69, 106, and 110 — and the agent carries `Read` in its `tools:`. A citation is not inert; it is not *auto-loaded*, costing a round-trip and not guaranteed to be followed.

The precise argument is stronger than the one written: those three existing citations point at **conditional** content the agent consults only when it hits that branch, whereas the PR-mechanics span is **main-path** and needed inline on every run that reaches a commit. Worth saying that way — it defends the verdict on its merits rather than on a mechanism claim the file itself falsifies.

**E. [info — expected, not a defect] Version fields, per the lead's instruction to leave them alone.**
`check-version-bumps.nu` against true `main` flags both touched plugins. Required before merge:

| Plugin | Current (both `main` and branch) | Needs |
|---|---|---|
| `elixir` | 0.1.21 | **0.1.22** — `plugins/languages/elixir/.claude-plugin/plugin.json` + `marketplace.json` entry |
| `github` | 0.3.10 | **0.3.11** — `plugins/tools/github/.claude-plugin/plugin.json` + `marketplace.json` entry |
| `all-skills` | branch shows **0.4.99**, `main` is **0.4.100** | **0.4.101** — per CLAUDE.md convention (the checker does not enforce it; no skill dir was added or removed) |

**Trap on that last row:** the branch predates `main`'s `3aeaf7f`, so its local copy of the root version reads 0.4.99. Bumping "from what the branch shows" produces 0.4.100 — identical to `main`, i.e. no bump at all after merge. **Rebase first, then bump.**

Nothing beyond version fields needs a rebase for correctness: `git merge-tree` reports 0 conflicts, and `main`'s only extra commit touches the `claude-code` plugin, `mise.toml`, and root version files — no overlap with `elixir`, `github`, or `quality-baseline.json`.

---

### Independent judgment on the 148 no-dedupe verdict: **correct, and correct for a better reason than the one given**

I set out to show the cost model flatters the conclusion. It does not.

- `/core:git` is **222 lines** — verified by `wc -l`, matching the claim exactly.
- `skills:` preloading is **eager**: per `/claude-code:claude-agents`, it "preloads full skill content into the subagent's context at startup." Not lazy, not deferred, paid per spawn.
- Crucially, **neither worker declares `skills:` today.** I checked both frontmatter blocks. So the 222 lines would be *new* cost, not already-paid cost. Had either already preloaded `/core:git`, the dedupe would have been free and this verdict wrong. It isn't.
- 222 preloaded to save 11 inline ≈ **20:1**, on every invocation whether or not the run ever reaches PR creation. The framing is honest.

**The stronger argument the author did not make — cross-plugin self-containment.** `bees-worker` ships in `core`; `beads-worker` ships in `beads`. A user installing only `beads@vinnie357` has no `core` plugin, so `skills: [core:git]` on `beads-worker` would be a dangling preload and a prose `/core:git` pointer would dangle too. This repo already has a convention for that hazard (the sibling-plugin optional-note work in claude-skills-92). For an agent's **main-path** operating instructions, dangling is not an acceptable failure mode. This also rules out the option not named in the comment — hoisting the span into one file and pointing the other at it — for the same reason.

And extracting span 2 into a new shared home would risk minting a **new** dupe group, which the shrink-only baseline cannot accept at all — the exact lesson already recorded on claude-skills-149.

Verdict stands: leave `dupe/28ccd32d` at 7. The only partial option available is trimming span 1 to drop below the window, which would game the detector rather than fix duplication, and would not clear the group anyway since span 2 alone is ≥ 8 normalised lines.

---

### Could not verify

- Whether upstream Tidewave still documents Cursor / VS Code / Zed setup (no network fetch performed) — bears only on Finding A's severity.
- Remote CI beyond the two reported jobs: `Validate marketplace and plugins` **pass**, `Check version bumps` **fail** (expected, §E). I reproduced both locally.

vinnie357:
## BEES REQUESTS

For the serial writer to apply (not applying directly — single-writer constraint):

**claude-skills-149** — close on merge:
```
bees close claude-skills-149 -r "Both dupe groups fixed in PR #202: dupe/1182fd06 (elixir phoenix/tidewave.md deleted, pointed at /elixir:tidewave) and dupe/639413bb (github pr-review registration.md collapsed to a cross-skill pointer). Baseline shrunk accordingly."
```

**claude-skills-148** — comment only, leave open as recorded DEBT (or close as won't-fix, lead's call):
```
bees comment add claude-skills-148 "Decision: no dedupe. dupe/28ccd32d (bees-worker.md vs beads-worker.md, 7 windows / 21 normalised lines across two spans) stays as-is. Reasoning: skills: preloading of /core:git (222 lines) to save an 11-line PR-mechanics span is disproportionate and is NEW cost (neither worker declares skills: today) rather than reuse. The other 10-line span (Pattern/Action dispatch table) isn't git content and has no existing-skill home. bees-worker ships in core, beads-worker ships in the beads plugin -- cross-plugin self-containment rules out both a skills: preload (dangling if only one plugin is installed) and hoisting the span into a shared file (same dangling-reference problem). Full evidence and measurement: https://github.com/vinnie357/claude-skills/pull/202#issuecomment-5153149053"
```

PR #205, opened 2026-08-04
docs(claude-code): document plugin.json fields, add quoted-source boundary category

vinnie357:
- Document `outputStyles`, `lspServers`, and `experimental` (themes/monitors) as plugin.json fields in the claude-plugins skill, with accepted shapes and an inline-example fixture, verified against `code.claude.com/docs/en/plugins-reference`
- Confirm via a constructed fixture that `validate-plugin.nu`'s invalid-fields denylist already accepts all three — no validator change needed
- Add a fourth stays-prescriptive boundary-test category (quoted source material) to the claude-skills reference doc, with the twelve-factor blockquote as the worked example
- Add a sources.toml/sources.md entry for the plugins-reference doc, and bump claude-code (0.2.48→0.2.49) and all-skills (0.4.103→0.4.104) versions in both manifests


vinnie357:

Independently re-fetched code.claude.com/docs/en/plugins-reference, 12factor.net (3 pages), re-ran the validator with fresh fixtures, and queried the live bees DB. Two blocking findings; the upstream field claims themselves all verify.

### Blocking

**1. The quoted-source worked example is itself a fabricated quote.** `context-engineering-claude-5.md` cites twelve-factor `SKILL.md:263` as a verbatim 12factor.net blockquote and asserts paraphrasing it "would misrepresent what the source actually says." The sentence *"A twelve-factor app never relies on implicit existence of state on the filesystem…"* does not appear anywhere on 12factor.net — it is a mashup of the **dependencies** factor ("A twelve-factor app never relies on implicit existence of **system-wide packages**", 12factor.net/dependencies) and the **processes** factor ("The twelve-factor app never assumes that anything cached in memory or on disk will be available on a future request or job", 12factor.net/processes). Verified by direct fetch of both pages plus an exact-phrase web search (zero hits). The worked example of "never alter a verbatim quote" is a paraphrase mislabeled as a quote — self-refuting, and it shows the category as written checks attribution, not fidelity. **Fix:** the *other* blockquote at `SKILL.md:261` ("The twelve-factor methodology can be applied to apps written in any programming language…") IS verbatim (verified against 12factor.net front page) — cite that one instead, or first fix :263 to the real processes-factor sentence. Also consider adding "after verifying the quote against its source" to the category so a mislabeled paraphrase can't claim immunity. The :263 fabrication itself predates this PR — file a bee against twelve-factor SKILL.md either way.

**2. "Filed as a new bee" is not true (yet).** sources.toml notes and sources.md both state the dependencies contradiction "is filed as a new bee rather than fixed here." The live bees DB's newest issue is claude-skills-216 (/core:security ps -E); no bee about `dependencies` being upstream-valid exists in bees.db or either issues.jsonl. Either file the bee before merge or reword to remove the completed-action claim.

### Verified correct (independent fetch, upstream quotes)

- `outputStyles` string|array, "replaces default `output-styles/`" — upstream manifest table + Path behavior rules ("when the manifest specifies `commands`, the default `commands/` directory is not scanned"; `outputStyles` is in the same replaces-list). The old "needs no manifest field" text was indeed wrong; the fix is right.
- `lspServers` string|array|object — confirmed; "Location: `.lsp.json` in plugin root, or inline in `plugin.json`" supports the defaults-when-absent wording. Note upstream classifies LSP under "own merge rules", not the replaces list.
- `experimental.themes`/`.monitors` both string|array, both replace their default scans — confirmed. Backward-compat claim confirmed in substance: "the top level still works, `claude plugin validate` warns, and a future release will require `experimental.*`". **Nit:** the skill omits that `claude plugin validate` *warns* on top-level placement — "also work unnested" reads more supported than upstream states.
- `dependencies` IS upstream-valid: "array | Other plugins this plugin requires, optionally with semver version constraints | `[{ \"name\": \"secrets-vault\", \"version\": \"~2.1.0\" }]`" — the contradiction with the skill's must-NOT list and the validator denylist is real. `displayName`, `defaultEnabled`, `workflows`, `userConfig`, `channels` all confirmed documented.
- Fixture re-run: all three fields pass `validate-plugin.nu` (0 errors). But so does `{\"notARealField\": \"garbage\"}` — the denylist-only validator accepts ANY unknown field. "No change needed" is correct for the stated AC, but the fixture proves little; worth a bee on the validator gap, and the point-in-time "Verified against…" sentence is PR-body material, not always-loaded skill body.
- marketplace.json diff is exactly the two version lines (jq -S diff). claude-code 0.2.49 > main 0.2.48, all-skills 0.4.104 > main 0.4.103 — passes #203's strict-greater check in any merge order with #203/#204 (both still open; #204 doesn't bump all-skills). Serialize merges; rebase-bump only if another PR claims 0.4.104 first.
- `nu test/validate-sources.nu` clean on the branch; the new sources.toml/sources.md entry is complete and the Entries mention is real content.
- Progressive disclosure: 136→155 lines, well under 500; the two bullets fit the existing component-paths list.

### Scope call on `dependencies`

Deferring the must-NOT-list + denylist fix is acceptable scope discipline — the skill still accurately describes what the *validator does* — but only with the tracked follow-up actually existing. That is why finding 2 blocks.

vinnie357:
## Gate 3 follow-up — blockers CONFIRMED FIXED at f31cb72; one new minor defect in the fix round

**Blocker 1 (fabricated quote): FIXED.** New citation `twelve-factor/SKILL.md:261` verified — line 261 on the new head is the methodology blockquote, and an independent re-fetch of 12factor.net confirms it **byte-exact**: "The twelve-factor methodology can be applied to apps written in any programming language, and which use any combination of backing services (database, queue, memory cache, etc)." twelve-factor SKILL.md is untouched by the PR; the :263 fabrication is tracked as claude-skills-217 (exists in DB). The reworded category genuinely closes the gaming vector, not just caveats it: fidelity is a stated precondition ("confirm the quoted text against the source before invoking this category"), an unverified quote is explicitly named "a fabrication the category would otherwise shield", and the boundary summary protects only "a verified quotation". Verification is an executable check the reader can perform (fetch the source), not an unmakeable judgement. Residual risk — claiming verification without doing it — is generic anti-fab enforcement, no longer a wording gap.

**Blocker 2 (phantom bee): FIXED.** Live DB query: claude-skills-218 = "claude-plugins skill and validate-plugin.nu both treat 'dependencies' as invalid, but upstream documents it"; claude-skills-219 = "validate-plugin.nu accepts any unknown plugin.json field - denylist has no allowlist counterpart". Both cited in sources.md AND sources.toml, each attached to the matching finding (not swapped). Grep of the full PR diff for surviving "filed as / is filed / will file" in added lines: zero.

**Nit (validate warning): FIXED on the reader-facing surface.** SKILL.md now states top-level "works ... today, but `claude plugin validate` already warns on the unnested form, and a future release will require nesting" — matches upstream ("the top level still works, `claude plugin validate` warns, and a future release will require `experimental.*`").

**Regression sweep:** `validate-skills-quality.nu` exit 0, no new violations, `quality-baseline.json` untouched by the PR; `validate-sources.nu` clean; SKILL.md 155 lines, reference 113 — under caps.

**One NEW defect introduced by the fix (small, sources.toml notes only):** the rewritten notes sentence attaches the validate-warning clause to the **outputStyles** parenthetical — "outputStyles (string|array, replaces default output-styles/ scan — reverting to the top-level unnested form triggers a claude plugin validate warning once a future release requires experimental.*)". Wrong twice: outputStyles has no nested/unnested form (the warning belongs to themes/monitors), and the timing is inverted (validate warns **today**; the future release will **require** — the SKILL.md fix states this correctly). One-clause fix: drop the em-dash clause from the outputStyles parenthetical; the later "unnested at the top level ... today, with that same validate warning" clause then needs the warning introduced there instead. Metadata prose, not skill body — lead's call whether it gates the merge.

PR #208, opened 2026-08-04
feat(claude-code): dedupe sources fetch/classify logic, add npm/docker-hub/endoflife-date check methods

vinnie357:
- Extract classify-fetch-error, catch-http-error, classify-staleness, sanitize-notes-cell, check-github-releases, check-hex-pm, check-crates-io, and fetch-latest into a shared sources-lib.nu module, imported by sources-check.nu, sources-stale.nu, and sources-report.nu (claude-skills-211)
- Route hex-pm and crates-io fetch failures through the same classify-fetch-error axes as github-releases via a shared catch-http-error wrapper; distinguish GitHub's 429 secondary rate limit (empirically verified as an unparenthesised "429 Too Many Requests" phrase, unlike the parenthesised 403/404) from a generic error (claude-skills-212)
- Add npm, docker-hub, and endoflife-date check_methods with companion identity fields npm_package, docker_image, and eol_product; extend test/validate-sources.nu's CHECK_METHODS/ENTRY_KEYS/NONE_FORBIDDEN and add conditional-field validation for the three new methods (claude-skills-210)
- Update SKILL.md's schema table and Check Method API Endpoints table, and templates/sources.toml with annotated examples for all three new methods
- Bump claude-code plugin to 0.2.49 in plugin.json and marketplace.json


vinnie357:

Reviewed in a scratchpad clone at 54a7a2b vs origin/main c90f943. Every claim below is from an executed run (nu 0.113.1).

### BLOCKER — F1: `check-npm` is dead on arrival (sources-lib.nu:198)

`$response | get -o "dist-tags.latest"` — a **quoted** string is one flat column name in nushell, not a nested cell path:

```
nested via quoted path: NULL          # {dist-tags: {latest: "9.9.9"}} | get -o "dist-tags.latest"
flat-key via quoted path: flat        # {"dist-tags.latest": "flat"}   | get -o "dist-tags.latest"
```

Live, through the lib: `check-npm express` → **unknown**, `@babel/core` → **unknown**, `@beads/bd` → **unknown** (nonexistent pkg correctly → no-releases, so the HTTP path including scoped names is fine — only the extraction is broken). End-to-end `sources-check.nu` run with `npm_package = "express"`, `current_version = "5.2.1"` (the actual current latest) reports `latest=unknown, stale=unknown` — the method can never detect drift, while the schema, SKILL.md ("reads `.dist-tags.latest`"), and validator all advertise it. Fix verified live: unquoted `get -o dist-tags.latest` returns `5.2.1`.

### MAJOR — F2: doc obligation missed

`references/version-check-methods.md` still documents only github-releases/hex-pm/crates-io/manual (untouched since #35), but the updated SKILL.md endpoint table ends with "See `references/version-check-methods.md` for Nushell parsing examples." The three new methods need entries there (or the pointer scoped).

### FINDING — F3: the "23 sources:validate errors" are NOT transient noise

Reproduced exactly: `272 URLs checked — 249 pass | 0 redirect | 0 fail | 23 error`. Independent re-sweep + curl of each error URL decomposes them:

- 14× `crates.io/*` — persistent **403** (anti-bot; HEAD and GET both 403 without a browser UA)
- 2× `github.com/anthropics/skills/blob/main/...` (skill-creator/SKILL.md, agent_skills_spec.md) — **real 404s**, content moved upstream
- `steveyegge.github.io/beads/` — **real 404**; `www.butunclebob.com/...` — connection dead (000)
- `github.com/vinnie357/runex` — private repo, anonymous 404 (permanent for unauthenticated runs)
- `m3.material.io` — HEAD=405, GET=200 (tool artifact); `molecule.readthedocs.io` — redirect artifact

Root cause: `sources-validate-urls.nu` classifies from `$err.msg`, which nu sets to a generic `"Network failure"` for 404s — the status lives in `$err.debug`. That is exactly the debug-fallback bug class this PR's `catch-http-error` fixes for the three check scripts, but the URL validator wasn't ported. That script is outside this PR's diff, so not a blocker here — but "0 fail / 23 error" is the msg-vs-debug bug hiding real dead URLs, and a follow-up bee should port `catch-http-error` into it (plus a HEAD→GET 405 fallback) and fix the dead anthropics/skills links.

### Verified good (attack surfaces that held)

- **Behavior preservation**: extracted every shared function from all three main-branch scripts and diffed against the lib: `classify-staleness`, `sanitize-notes-cell` byte-identical; `classify-fetch-error` identical for all old inputs + the new 429 matcher; `fetch-latest` identical + three new arms; per-script functions (`is-date-stale`, `priority-weight`, `stale-badge`, `md-row`, all `process-sources-toml`) byte-identical. Offline twin-fixture run of all three scripts old-vs-new: **byte-identical output** (only the printed repo-root path differs).
- **Self-test counts**: reproduced 13+7/25/26 → 42/47/42 and validator 74→80. Genuinely new assertions: +4 fetch-error (hex-404, 429, 429-span, 401), +3 catch-http-error, +9 dispatch, +6 validator conditionals. Caveat: 42 is the shared suite re-run per script (check and report now have zero script-specific cases; notes cases count in scripts that don't render notes), so per-script totals overstate per-script-relevant coverage — disclosed in the wrapper comments, acceptable.
- **Double import**: unfixed shape reproduced — local `run-self-test` calling the bare name hits `nu::shell::recursion_limit_reached (50)`, so the module-qualified fix is justified, and the fixed shape runs clean. Only `run-self-test` collides; `main` and `USER_AGENT` are unexported. Module resolution claim re-verified both ways (script sited elsewhere fails even with cwd=scripts dir; sited beside lib works from cwd=/).
- **429**: live httpbin 429 renders `Error is "429 Too Many Requests"` (no parens) → rate-limited; 401/500 bare shapes → error; a 404 text containing the phrase still classifies no-releases (404 checked first); a 200 body containing the phrase never reaches the classifier (no catch fires).
- **hex/crates via lib, live**: `phoenix` → 1.8.9, `serde` → 1.0.229, hex 404 → `no-releases` (new classification confirmed live).
- **docker-hub**: `ordering=last_updated` empirically returns newest-first (2026-08-03 tag; `-last_updated` returns 2015 — inverse of Django convention, worth a comment but the choice is correct and matches the no-param default). Caveat as designed/documented: for active images (library/node: 8992 tags, rolling tags repushed daily) any version-shaped pin compares stale forever — signal is "a newer tag exists", as the code comment states.
- **endoflife-date**: nodejs → 26.6.0 (matches first cycle), nonexistent → no-releases. Nit: a fully-EOL product reports the newest cycle's latest regardless of EOL status (`centos` → `8 (2111)`) — a freshness tool recommending an EOL version; worth a note or follow-up, not a blocker.
- **Schema conditionals, break-it**: `none`+`npm_package` → b4_none_forbidden; `npm` w/o companion → b4_conditional; `manual`+`eol_product` → legal (matches A-F1 precedent); both companions + `npm` → legal. All reported, none crashed.
- **Nushell traps**: no `2>&1`, no bare `main`, no self-write `save`, no `describe` string-matching in the changed files.
- **`mise test`** and `test/check-version-bumps.nu` pass on the branch clone.
- **Restraint**: extraction pulled exactly the duplicated functions + shared self-test; `catch-http-error` and `USER_AGENT` are the only new symbols, both with ≥6 call sites. The triplication-accepted fallback was dismissed on evidence (resolution verified in two invocation shapes).

### Record correction

The pre-PR copies were **not** byte-identical: `check-github-releases`/`check-crates-io` embedded per-script User-Agent strings (`sources-check.nu/1.0` etc.). The lib unifies them to `claude-skills-sources/1.0 (github.com/vinnie357/claude-skills)` — a benign, undocumented UA change worth one line in the PR body.

### Merge-order

main has claude-code **0.2.48**; PRs **#205, #207, and this one all claim 0.2.49**  — git raises no conflict on identical values, and #203 makes the bump check strictly-greater-than-origin/main. Re-bump matrix for this PR: merges first → 0.2.49 correct; after one sibling → **0.2.50**; after both → **0.2.51** (both plugin.json and marketplace.json).

vinnie357:
## Gate 3 re-verification of 6dbd6d3 — findings CONFIRMED FIXED

Verified in the scratchpad clone at 6dbd6d3; every claim below is from an executed run (nu 0.113.1).

**F1 (BLOCKER, npm parse) — CONFIRMED FIXED, and the test catches it.** Mutation reproduced: re-quoting the path in `parse-npm-latest` makes the self-test fail on exactly the intended case — `❌ npm parse: npm: dist-tags.latest extracted from a real-shaped response: want 5.2.1, got unknown` (rc=1); restored → `✅ sources-lib self-test passed (48 cases)`. Fixtures verified against live fetches, not reverse-engineered: npm `dist-tags` live is `{latest: 5.2.1, latest-4: 4.22.2}` (fixture exact match), docker-hub live `count=8992`, `results[0].name=lts-trixie-slim` (fixture is a key-subset of the real 16-key record — fine for a parse test), endoflife nodejs first two cycles live `[{26, 26.6.0, 2029-04-30}, {25, 25.9.0, 2026-06-01}]` (fixture exact match). End-to-end through the lib: express → **5.2.1**, `@babel/core` → **8.0.1**, nonexistent → **no-releases**, library/node → **lts-trixie-slim**, nodejs → **26.6.0** — all five match the reported values.

**F2 (MAJOR, reference doc) — CONFIRMED FIXED, one inherited nit.** All three sections present (184→280 lines); endpoint tables, the quoted-path gotcha, and the ordering warning are accurate. The three worked examples' bodies produce exactly the stated outputs (5.2.1 / lts-trixie-slim / 26.6.0, run live). Nit, pre-existing: the `def f [args] -> string {` signature shape does not parse on nu 0.113.1 (needs `def f [args]: nothing -> string {`) — but the three NEW sections copied it from the three PRE-EXISTING sections (lines 23/61/95, unchanged since #35), so the whole file shares it. Follow-up sweep material, not this PR's defect.

**Record correction — ACCURATE.** The new header matches my round-1 measurements exactly: `classify-fetch-error`/`classify-staleness` byte-identical across the three main-branch scripts; `check-github-releases`/`check-crates-io` differed only in per-script User-Agent strings; unification to `$USER_AGENT` is now stated as a deliberate behavior change.

**Regression sweep — CLEAN.** Self-tests: check 48, stale 53, report 48, lib 48, validator 80 — all green. Offline twin-fixture byte-diff of all three scripts (main vs 6dbd6d3) on identical input: **OUTPUT IDENTICAL** (modulo the printed fixture path). `check-version-bumps` green locally (merge-order collision handled separately per lead).

### Judgement calls requested

**docker-hub tag-name signal — design gap, recommend a tracked follow-up (not a blocker).** The docs are honest, but the shipped shape is degenerate as a drift comparator: the schema-compliant usage the template itself prescribes (`current_version = "unknown"`, `rolling`) classifies **no-pin forever** — the method cannot flag staleness for its own flagship example (`node:16-buster-slim`). The alternative (a version-shaped pin) compares against whatever tag was repushed most recently on a high-churn image (library/node: 8992 tags, rolling aliases repushed daily) and reads **yes forever**. Neither state ever transitions, so the "latest" column is human-readable context, not a signal. Hub's per-tag endpoint (`/tags/{tag}`) or a tag-filter companion field would make "has my pinned tag's line moved" answerable; file it as a bee rather than leaving it as prose.

**endoflife-date EOL decision — rationale is internally consistent; follow-up warranted.** The "same question as every other method" argument holds: none of the methods filter by supportedness. But claude-skills-176 — the issue that motivated this method — is explicitly about EOL pins, the response already carries each cycle's `eol` field (zero extra network), and the one failure mode a freshness tool should not have is confidently reporting an EOL version as "latest" (centos → `8 (2111)`, live-verified). Sufficient for this PR as documented; recommend a tracked follow-up for an `eol` staleness sentinel (or a documented decision not to), so the method's name and the motivating issue stop pointing at a capability it lacks.

PR #216, opened 2026-08-04
fix(claude-code): sources tooling follow-ups from PR #208/#211 Gate 3

vinnie357:
- fix six worked nushell examples in version-check-methods.md that used a `[args] -> type` signature — does not parse on nu 0.113.1, needs `[args]: input -> type` (claude-skills-227); repo-wide sweep found no other instances
- add an optional `docker_tag` field: `check-docker-hub-tag` compares a specific pinned tag's content digest instead of the newest tag overall, which transitions correctly for a realistically pinned image (claude-skills-225)
- add `check-endoflife-status` alongside the unchanged `check-endoflife-date`: surfaces the newest cycle's own `eol` field so a fully-EOL product (e.g. centos) no longer reports its last patch as "latest" with no sentinel (claude-skills-226)
- add an informational `latest=` line to `fence-freshness.nu`'s report (github-releases and docker-hub only), importing and using `check-github-releases`/`check-docker-hub`; rename the `ok` badge to `exists` (claude-skills-229). Note: `check-github-releases` was not actually a dead import on current main — only `catch-http-error` was imported — but it is genuinely wired up and used now regardless
- register `docker_tag` as a known sources.toml key in `test/validate-sources.nu` and document both docker-hub usages in `templates/sources.toml`

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.113",
+      "version": "0.4.114",
       "category": "meta",
       "keywords": [
         "skills",
@@ -119,7 +119,7 @@
       "source": "./plugins/tools/claude-code",
       "strict": true,
       "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
-      "version": "0.2.55",
+      "version": "0.2.56",
       "category": "tools",
       "keywords": [
         "claude-code",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -256,7 +256,7 @@
       "source": "./plugins/tools/agent-sandboxing",
       "strict": true,
       "description": "Agent sandboxing for AI coding agents: microVM and kernel-isolated execution environments. Index skill (sandbox) routes to substrate-specific paths. Currently covers the kubernetes-sigs/agent-sandbox path on kind, GKE, and kina (Kubernetes-in-Apple-Containers), with Kata, gVisor, and Apple Container microVM substrates. Ships 7 skills + 6 slash commands + nushell scripts. Designed to expand to non-k8s sandboxing paths.",
-      "version": "0.2.11",
+      "version": "0.2.12",
       "category": "tools",
       "keywords": [
         "agent-sandbox",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.113",
+  "version": "0.4.114",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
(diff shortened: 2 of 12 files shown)


vinnie357:
## Gate 3 follow-up findings — resolved

Resumed after the branch stalled post-Gate-3. All three findings addressed in commit 966d7ad.

**F1 (BLOCKER) — docker_tag digest comparator was un-adoptable: FIXED.**
`b3_version_shape` in `test/validate-sources.nu` rejected the ONLY value `check-docker-hub-tag` can legitimately write — a `sha256:<64 hex>` digest never matches `VERSION_SHAPE_RE` (the colon isn't in its alphabet). Added `DOCKER_DIGEST_RE`, scoped the acceptance to entries with `docker_tag` set (a non-docker_tag entry with a coincidentally digest-shaped `current_version` still fails, so a typo doesn't silently pass). Corrected the false fixture comment claiming "VERSION_SHAPE_RE never applies to docker-hub's digest-comparator usage" — the rule applies unconditionally; the fixture just never actually carried a digest before. Added a fixture with a real digest that now passes, plus a negative fixture proving the scoping holds. `--self-test` 80 → 83 cases.

Adopted `docker_tag` on a real entry: `plugins/tools/agent-sandboxing/skills/sources.toml`'s new `debian-13-slim-base-image`, tracking the actual `FROM debian:13-slim` pin in this plugin's own `templates/Dockerfile.claude-code:7` (previously untracked). Verified end-to-end, not just schema-level:
- `fetch-latest` against the real entry returns the live digest, matching `current_version` (`classify-staleness` → `no`).
- Same entry with a deliberately wrong `current_version` → `classify-staleness` → `yes`.
- `mise test:sources` and `mise test` both clean with the new entry live in the corpus.

**F3 — docker_tag / check-endoflife-status undocumented: FIXED.**
Added `docker_tag` to `skill-update/SKILL.md`'s schema table and a paragraph on the per-tag endpoint + digest semantics; added a paragraph documenting `check-endoflife-status` (the EOL-aware counterpart `check-endoflife-date` wraps) in the Check Method API Endpoints section. `SKILL.md` stays at 267 lines (limit 500).

**F2 (BLOCKER) — stale rebase: FIXED.**
main had moved through #212/#213/#215 since this branch's base. Rebased onto current `origin/main` (verified #214 still open, not landed, before and after). Rebuilt `marketplace.json` from `origin/main` and applied only this PR's version fields via `jq`, diff-verified the blast radius was exactly the intended entries. Version bumps: `claude-code` 0.2.55→0.2.56, `all-skills` 0.4.113→0.4.114, and `agent-sandboxing` 0.2.11→0.2.12 (needed once the new `sources.toml`/`sources.md` entries were added — caught by `mise run test:version-bumps`, not assumed).

**Confirmed sound from the original review, not reworked:** the `docker_tag` digest-parse mutation test and `parse-endoflife-cycle-eol` mutation tests both still fail correctly when broken; the `check-github-releases` dead-import correction in fence-freshness.nu was accurate and is unchanged here.

**Gate 1:** `mise test` (exit 0), `mise test:sources`, and `mise run test:version-bumps` all green; gitleaks (`--source . --no-git -v`) clean.
**Gate 2:** both checks (`Validate marketplace and plugins`, `Check version bumps`) pass on run [30955955651](https://github.com/vinnie357/claude-skills/actions/runs/30955955651), headSha `966d7ad` confirmed matching the PR.

Not merging — awaiting operator approval per Gate 3 policy.

PR #221, opened 2026-08-04
test(claude-code): close self-test mutation-coverage gaps in validate-plugin.nu

vinnie357:
- Add reject-direction fixture coverage for validate-skill-md (9 cases), is-kebab-case (3 call-site cases), is-semver top-level-version (2 cases), validate-agent-md non-model checks (8 cases), cleanup-temp (2 cases), and CLI-level cases reaching main/validate-plugin-file/validate-from-marketplace/setup-external-plugin's non-network branch
- Fix cli_invalid_json_exit_1 fixture: "not valid json {{{" parses cleanly as a JSON5-style bare string under nu's own `from json`, so the intended catch branch was never exercised — replaced with genuinely truncated JSON that nu's parser rejects
- Fix cli_marketplace_valid_local_plugin_exit_0 fixture: the marketplace entry was missing a `description` field that plugin.json defines, tripping the intentional claude-skills-170 deletion-is-a-failure check — added the matching description
- Verify by mutation per function (neuter to always-pass, confirm failures, restore, confirm green): is-kebab-case (3/3 reject cases fail), is-semver (2/2), validate-skill-md (8/9 non-baseline cases fail), validate-agent-md non-model checks (7/8 non-baseline cases fail)
- Bump claude-code 0.2.59 -> 0.2.60 and all-skills 0.4.119 -> 0.4.120

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.119",
+      "version": "0.4.120",
       "category": "meta",
       "keywords": [
         "skills",
@@ -119,7 +119,7 @@
       "source": "./plugins/tools/claude-code",
       "strict": true,
       "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
-      "version": "0.2.59",
+      "version": "0.2.60",
       "category": "tools",
       "keywords": [
         "claude-code",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.119",
+  "version": "0.4.120",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/claude-code/.claude-plugin/plugin.json
+++ b/plugins/tools/claude-code/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-code",
-  "version": "0.2.59",
+  "version": "0.2.60",
   "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/claude-code/skills/claude-plugins/scripts/validate-plugin.nu
+++ b/plugins/tools/claude-code/skills/claude-plugins/scripts/validate-plugin.nu
@@ -1151,7 +1151,7 @@ def self-test [] {
     }
     {
       name: "top_level_version_malformed_warns"
-      why: "claude-skills-238 — is-semver (the plugin.json top-level 'version' field, distinct from the dependency RANGE check above) had zero fixture coverage; mutating it to always-true passed the pre-fix 30-case suite unchanged"
+      why: "claude-skills-238 — is-semver (the plugin.json top-level 'version' field, distinct from the dependency RANGE check above) had zero fixture coverage; mutating it to always-true passed the pre-fix 34-case suite unchanged"
       plugin: { name: "my-plugin", version: "not-a-version", description: "test fixture", license: "MIT" }
       expect_errors: 0
       expect_warnings: 1
@@ -1230,7 +1230,7 @@ def self-test [] {
   # — no case in $cases above carries a `skills` entry, so a mutation
   # replacing the whole function body with an unconditional error (or with
   # `{ errors: [], warnings: [] }`, the always-pass direction the bee is
-  # actually about) passed the pre-fix 30-case suite unchanged. Called
+  # actually about) passed the pre-fix 34-case suite unchanged. Called
   # directly against real temp SKILL.md files, the same pattern as
   # agent_model_cases above, rather than threaded through a full plugin.json
   # + skills-array + on-disk-directory round trip — that indirection buys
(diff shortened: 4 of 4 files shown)


vinnie357:
## Gate 3 — adversarial review (separate opus agent, read-only against a scratch clone)

Verdict: 1 blocker, 2 real-but-minor, 1 nitpick. All four PR-body mutation claims reproduced exactly by the reviewer independently.

**Blocker (fixed, commit db37dae):** the claude-skills-170 marketplace description/keywords-agreement check itself (`validate-plugin.nu:362`, `:534`) had zero reject-direction fixture coverage. Reviewer proved it by neutering both `has_marketplace_context` blocks to `if false` — self-test stayed green, and a real marketplace-description-drift scenario (edited `.claude-plugin/marketplace.json`) silently validated as OK. This was the bee's exact failure mode surviving inside the PR that closes the bee. Fix: added `desc-mismatch-plugin` and `keywords-mismatch-plugin` fixtures + two new CLI cases (`cli_marketplace_description_mismatch_exit_1`, `cli_marketplace_keywords_mismatch_exit_1`). Re-verified the same `if false` mutation now fails both new cases before restoring.

**Real-but-minor #1 (not fixed here, flagged for follow-up):** `validate-plugin-file`'s invalid-JSON catch only guards against a parse *exception* — content that nu's `from json` parses leniently as a scalar (bare string, number, bool, list) instead of raising still crashes with an unhandled `nu::shell::only_supports_this_input_type` at line ~140 rather than the documented "Invalid JSON syntax" message. Confirmed by the reviewer across 7 scalar variants. `SKILL.md` claims the tool "checks JSON syntax" without documenting this gap. Out of scope for this PR (the fixture fix for `cli_invalid_json_exit_1` was itself correct — the old fixture just wasn't testing what it claimed to). Filing a follow-up bee (see BEES REQUESTS below, applied by team-lead — I hold read-only bees per this session's dispatch).

**Real-but-minor #2 (not fixed here, flagged for follow-up):** reviewer's systematic sweep found 11 more always-pass-neuterable checks with zero reject-direction fixtures — the entire skills/commands/agents array-type and path-existence surface (e.g. "'keywords' must be an array", "Skill directory ... missing SKILL.md file", "Command path not found"). None were in this PR's claimed scope (the bee named four specific functions; this PR also closed the CLI-reachability gap and, per Gate 3, the marketplace-agreement gap). claude-skills-238 is not fully closed by this PR — filing a follow-up bee for the remaining surface.

**Nitpick (fixed, commit db37dae):** two comments said "pre-fix 30-case suite" — measured against `origin/main`, it's 34 cases. Both corrected.

**Restraint verdict:** proportionate. Reviewer proved (not assumed) the ~370 total added lines duplicate nothing in `mise run test:plugins` — that integration test only exercises the AGREE direction by construction (every fixture plugin in this repo already passes), so it contributes zero mutation-detection power; verified by running the same neutering mutation against it and observing no change. The decision to decline an automated mutation-probe harness (bee's own open question) stands, citing the existing repo-wide manual `why:`-commented-fixture convention (`test/validate-core-list.nu` uses the identical pattern) as precedent — though the reviewer notes the 11 remaining gaps argue for *finishing* manual coverage rather than *building* the harness, which is consistent with declining the harness now.

**Documentation verdict:** no doc updates obligated by this PR's actual changes (self-test fixtures only, no production behavior changed). The `SKILL.md` "checks JSON syntax" overstatement is pre-existing and tracked under real-but-minor #1 above, not introduced by this PR.

## BEES REQUESTS (queued for team-lead / bees-manager — I hold read-only bees this session)

1. New bee: `validate-plugin-file`'s invalid-JSON catch doesn't guard against nu's lenient JSON5 scalar parsing (bare string/number/bool/list content crashes with an unhandled nu shell-type error instead of "Invalid JSON syntax"). Fix: after `open`, check `($plugin | describe | str starts-with "record")` and treat non-record as the same invalid-JSON error path. Add a fixture. `SKILL.md`'s "checks JSON syntax" claim and `references/validation-and-troubleshooting.md` should be updated to reflect actual behavior once fixed.
2. New bee: close the remaining always-pass-neuterable surface in `validate-plugin.nu`'s skills/commands/agents array-type and path-existence checks (11 checks identified by Gate 3 mutation sweep on PR #221 — see reviewer's Finding 3 for the full list). claude-skills-238 is not fully closed until these have reject-direction fixtures too.

vinnie357:
## Gate 3 round 2 — team-lead's independent adversarial pass

team-lead ran their own mutation sweep against the 66-case suite and found two surviving mutations plus one optional nit, all in the exact functions this PR claims to cover:

**Fixed — `is-kebab-case` accepting underscores** (stub: `^[a-z0-9_]+(-[a-z0-9_]+)*$`). Survived because the only reject fixture, `My_Plugin`, differs from valid kebab-case in TWO dimensions at once (case AND underscore) — the uppercase alone fails it even under the mutant, so the fixture never isolated which rule did the rejecting. Added `plugin_name_snake_case_errors` (`my_plugin` — lowercase, underscore-only) which differs in exactly one dimension. Re-ran the exact mutation: now fails with `expected 2 errors, got 1`.

**Fixed — `is-semver` as starts-with-digit** (stub: `^[0-9]`). Survived because both reject fixtures (`not-a-version`, `v1.0.0`) start with a non-digit and every accept fixture is a full `X.Y.Z`. Added `top_level_version_two_segment_warns` (`1.0` — starts with a digit, so the stub wrongly accepts it, but is missing the patch segment, so the real check must still warn). Re-ran the exact mutation: now fails with `expected 1 warnings, got 0`.

**Fixed (optional nit, taken) — agent `tools` YAML-list branch message-specificity.** Neutering that branch to `if false` left `expect_errors: 1` still true because the `else if $tools_type != "string"` fallback also errors on a list, masking which branch actually fired. Extended the `agent_md_cases` loop with an `expect_error_contains` assertion (mirroring the existing `cli_cases` pattern) and pinned `tools_as_yaml_list` to the substring `"comma-separated string, not YAML array"`. Re-ran the exact mutation: now fails with the specific-message mismatch instead of passing on count alone.

Generalizable lesson recorded in the fixtures' `why:` comments: a reject fixture must differ from the accept fixtures in exactly one dimension, or it proves "something was rejected" without proving "this rule rejected it."

Suite now 68 cases (was 66). Full `mise run ci` green, gitleaks clean, `mise run test:version-bumps` green (live-read immediately before push, unchanged at claude-code 0.2.59→0.2.60 / all-skills 0.4.119→0.4.120 since neither this PR nor #222 has merged yet). Gate 2 re-confirmed: headSha f316489 matches both the PR and the CI run exactly.

Everything else from round 1 stands — see the earlier Gate 3 comment above.

PR #223, opened 2026-08-05
test(claude-code): close remaining array-type and path-existence mutation gaps

vinnie357:
- Add reject-direction fixtures for the 11 always-pass-neuterable checks Gate 3 found on PR #221: keywords/skills/commands/agents-must-be-an-array (4), missing recommended description/license (2), skill-path-not-found, skill-directory-missing-SKILL.md, sources.md-missing, command-path-not-found, agent-path-not-found (5)
- Two on-disk fixture roots isolate the path-existence checks along exactly one dimension each: one directory that genuinely exists but lacks SKILL.md, one that genuinely doesn't exist at all, one with skills/sources.md present vs one without
- Verify every check by mutation, twice: the obvious always-pass neuter (11/11 killed) AND a non-equivalent mutation per check — wrong field, wrong constant, off-by-one path, inverted exemption logic (11/11 killed, closing two that survived on the first fixture design and required pre-creating parent directories to distinguish "file missing" from "whole directory missing")
- No mutation-probe harness added — declining stands per the restraint reasoning already recorded on PR #221, unchanged
- Bump claude-code 0.2.61 -> 0.2.62 and all-skills 0.4.121 -> 0.4.122

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.121",
+      "version": "0.4.122",
       "category": "meta",
       "keywords": [
         "skills",
@@ -119,7 +119,7 @@
       "source": "./plugins/tools/claude-code",
       "strict": true,
       "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
-      "version": "0.2.61",
+      "version": "0.2.62",
       "category": "tools",
       "keywords": [
         "claude-code",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.121",
+  "version": "0.4.122",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/claude-code/.claude-plugin/plugin.json
+++ b/plugins/tools/claude-code/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-code",
-  "version": "0.2.61",
+  "version": "0.2.62",
   "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
   "author": {
     "name": "vinnie357"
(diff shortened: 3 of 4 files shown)


vinnie357:
## Gate 3 round 2 — team-lead's independent adversarial pass: REQUEST CHANGES, addressed

All three findings independently reproduced and fixed:

**1. Surviving mutation — "right check, wrong report" (line 472).** Swapping the command-path warning's literal to `"Agent path not found: ($command)"` kept severity+count identical and passed 79/79, because all 11 new fixtures asserted counts only. Added `expect_error_contains` / `expect_warning_contains` to all 11, pinning full message text (not just the field name — the exact string including the path, e.g. `"Command path not found: commands/does-not-exist.md"`). Re-ran the exact mutation: now fails with `expected a warning containing 'Command path not found: ...', got: ["Agent path not found: ..."]`.

**2. Crash reachable WITHOUT mutation — `agents: ["agents"]` where the entry resolves to a directory.** Reproduced independently on unmutated code: `validate-agent-md`'s `open $agent_path --raw` raises an uncaught `nu::shell::io::is_a_directory`, and since self-test accumulates failures and prints only at the end, this silently discards every already-recorded result. Fixed with a `path type` guard — generalized to `!= "file"` rather than `== "dir"` specifically, because while verifying the fix against the two mutations that had "died by crash" in round 1, I found `path type` returns EMPTY (not `"dir"`, no error) for a path a broken existence-check wrongly claims exists — the `== "dir"` version alone would still crash under that mutation. `!= "file"` catches both. Added `agent_path_is_directory_errors` as the regression-guard fixture. Re-ran the round-1 dirname mutation against `agent_path_missing_warns`: previously crashed with `file_not_found`, now fails cleanly with a full assertion diff, exit 1, no crash.

**3. Fixtures cementing a validator/doc divergence — commands/agents "must be an array".** `plugin-schema.md:21-22` and `SKILL.md:50/52` document both fields as accepting a bare string (a directory to scan), but the validator rejects any non-array value, and my fixtures asserted that rejection. Chose RECORD over silent-resolve after checking what a correct fix requires: the string form means "scan this directory for files" (skills does this for its own array entries, but nothing in this validator implements directory-scanning at all) — a real feature addition, not a coverage-gap fix, and a rushed partial fix (naively coercing the string to a one-element list) would have been actively wrong, since it would treat a valid directory as a single invalid file entry. Documented the divergence explicitly in both fixtures' `why:` comments with exact doc line citations, and am filing a follow-up bee (see BEES REQUESTS on the coordination thread) to decide: implement the documented directory-scan behavior, or narrow the docs to array-only.

Self-test: 80/80 green (79 -> 80, the new crash-regression fixture). `mise run ci` EXIT_CODE=0. gitleaks clean. `mise run test:version-bumps` green, live-read immediately before push (origin/main unchanged at ca6b5b1, claude-code/all-skills already correctly bumped to 0.2.62/0.4.122 from the prior push). Gate 2 re-confirmed: headSha d365908 matches both the PR and the CI run exactly.

PR #224, opened 2026-08-05
fix(claude-code): prevent validate-skill-md crash on a mutated exists-guard

vinnie357:
- Follow-up to #223 (merged): team-lead's Gate 3 reviewer found a second, more severe twin of the agents-as-directory crash — validate-skill-md's `open $skill_md_path --raw` crashes with an uncaught nu shell I/O error whenever the SKILL.md-exists guard is bypassed
- Reproduced independently before fixing: a guard mutation checking `skill_md`'s dirname (which is `skill_path` itself, already confirmed to exist by the outer branch) makes the original `not (path exists)` guard PERMANENTLY inert — every missing-SKILL.md case would crash, not a contrived one
- Same fix pattern as the agents case in #223: guard on `(path type) != "file"` before calling into the function that opens the file
- Re-ran the exact mutation against the fix: fails cleanly with a normal self-test assertion diff and exit 1, no crash
- No new fixture needed — the existing `skill_dir_exists_no_skill_md_errors` case already exercises the unmutated guard; this fix closes the crash surface a mutation could still reach
- Bump claude-code 0.2.62 -> 0.2.63 and all-skills 0.4.122 -> 0.4.123

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.122",
+      "version": "0.4.123",
       "category": "meta",
       "keywords": [
         "skills",
@@ -119,7 +119,7 @@
       "source": "./plugins/tools/claude-code",
       "strict": true,
       "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
-      "version": "0.2.62",
+      "version": "0.2.63",
       "category": "tools",
       "keywords": [
         "claude-code",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.122",
+  "version": "0.4.123",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/claude-code/.claude-plugin/plugin.json
+++ b/plugins/tools/claude-code/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-code",
-  "version": "0.2.62",
+  "version": "0.2.63",
   "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/claude-code/skills/claude-plugins/scripts/validate-plugin.nu
+++ b/plugins/tools/claude-code/skills/claude-plugins/scripts/validate-plugin.nu
@@ -447,29 +447,47 @@ def validate-plugin-content [
           $warnings = ($warnings | append $"Skill path not found: ($skill)")
         } else {
           let skill_md = ($skill_path | path join "SKILL.md")
-          if ($skill_md | path expand | path type) != "file" {
-            # claude-skills-244 Gate 3 round 2 — same crash class as the
-            # agents branch above, independently found and confirmed by
-            # team-lead's reviewer: validate-skill-md's `open $skill_md_path
-            # --raw` crashes with an uncaught nu shell I/O error whenever
-            # this guard wrongly believes SKILL.md exists. `!= "file"`
-            # (not `not (path exists)`) on purpose — `path dirname` of
-            # skill_md is skill_path itself, ALREADY confirmed to exist by
-            # the outer branch, so a mutation checking the dirname instead
-            # of skill_md leaves the ORIGINAL `not (path exists)` guard
-            # permanently inert (every missing-SKILL.md case would crash,
-            # not just a contrived one) — reproduced and verified directly.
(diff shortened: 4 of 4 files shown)


vinnie357:
## Gate 3 round 3 — third crash instance + symlink false-positive, both fixed

Folded per team-lead's authorization (same file, same defect class).

**Symlink false-positive (priority — affects the guard #223 already merged).** `path type` reports `"symlink"` for a symlink regardless of what it points at; `open` follows symlinks fine. Verified all 6 relevant states with `path expand | path type` before applying: real file → `file` (accept), real dir → `dir` (reject), symlink→file → `file` (accept, was wrongly rejected before this fix), symlink→dir → `dir` (reject), broken symlink → never reaches the guard at all (`path exists` is already `false` for it — commented in-code so it isn't later "fixed" as a gap), nonexistent path → empty/reject. Applied `path expand |` to both guards (agents, already-merged in #223; skills, this branch's own #224 addition).

Added two ACCEPT-direction regression fixtures with real on-disk symlinks (`^ln -sf`, not mutations): `skill_symlink_to_real_file_accepts`, `agent_symlink_to_real_file_accepts`. Verified revert behavior directly, per the ask: reverting both guards to bare `path type` makes both fixtures fail by an ordinary self-test assertion mismatch (`expected 0 errors, got 1: [...]`), exit 1, **no crash** — confirmed by actually reverting, running, and observing the output before restoring.

**Third `open` instance.** Swept every `open` call in the file (6 total). Two (`open $plugin_path` at what are now lines 148/236, inside `validate-plugin-file`/`validate-plugin-content`) are wrapped in try/catch — verified directly that nushell's try/catch DOES catch `is_a_directory` (`try { open <a-real-directory> } catch { |e| print $e.msg }` → caught cleanly), and re-verified against the current file post-fix (direct-file mode against a directory target → clean "Invalid JSON syntax" error, exit 1, no crash) — unchanged, unaffected by this round's edits. The third: `validate-from-marketplace`'s `open $marketplace_path` (line 50, no try/catch) after a `path exists` check that's true for directories too. Reproduced with zero mutation and no plugin.json at all: `--marketplace <a-directory>` crashed with uncaught `is_a_directory`. Guarded with the same `path expand | path type != "file"` pattern, clean error + exit 1 confirmed. Added `cli_marketplace_path_is_directory_exit_1` reusing an existing fixture directory (no new setup).

Three instances of one class in this file — the real story, as flagged: nobody swept after the first two until this round.

Self-test: 83/83 green (80 -> 83: 1 new CLI case + 2 symlink-accept fixtures). `mise run ci` EXIT_CODE=0. gitleaks clean. `mise run test:version-bumps` green, live-read immediately before push (main unchanged at cb46320). Gate 2 re-confirmed: headSha a7f84fa matches both the PR and the CI run exactly. Diff is single-file, validate-plugin.nu only — nothing from 243/245 folded in.

vinnie357:
## Gate 3 round 4 — the fix wasn't pinned, fixed

Independently verified before touching anything: reverted the skills guard back to bare \`not (path exists)\` and confirmed the suite stayed **83/83 green** — my PR body's claim that \`skill_dir_exists_no_skill_md_errors\` "already exercises the unmutated guard" was true but didn't distinguish which guard shape was running, exactly as flagged. That fixture's on-disk state (SKILL.md simply absent) passes under both the old and new guard identically.

**Fix:** restructured the skills branch into two explicit checks instead of one combined \`!= "file"\`:
1. \`not ($skill_md | path exists)\` → the existing "missing SKILL.md file" message, for the common case.
2. \`($skill_md | path expand | path type) != "file"\` → a new, distinct "SKILL.md that is a directory, not a file" message, matching the agents branch's existing wording per the nit.

This is also defense-in-depth: even if check 1 alone were mutated (e.g. checking the wrong path), check 2 still catches it, since \`path expand | path type\` on a path that doesn't really exist resolves to empty, not "file" — verified directly.

**New regression fixture** \`skill_md_is_directory_errors\`: a real on-disk directory literally named "SKILL.md" (not a mutation). Verified it's reachable with **zero mutation on truly unmutated pre-#224 code** — checked out \`cb46320\` (main before this whole PR chain) and ran the exact fixture against it: crashes immediately with \`nu::shell::io::is_a_directory\`, confirming this is reachable outright, same as the agents twin, not merely "under a bypassed guard" as my PR body originally framed it.

**Verified both revert directions directly, per the ask:**
- Full revert to bare \`not (path exists)\` (no type check at all): this fixture **crashes** with an uncaught I/O error, exit 1.
- Revert of only the message split (type check intact, generic message restored): this fixture **fails cleanly** by an ordinary self-test assertion mismatch — \`expected an error containing 'SKILL.md that is a directory, not a file', got 'missing SKILL.md file'\` — exit 0 for the reverted run, reported failure in the self-test output.

Neither revert direction passes silently now.

Self-test: 84/84 green (83 -> 84). \`mise run ci\` EXIT_CODE=0. gitleaks clean. \`mise run test:version-bumps\` green, live-read immediately before push (main unchanged at cb46320, versions already correctly bumped at 0.2.63/0.4.123). Gate 2 re-confirmed: headSha e2ee46f matches both the PR and the CI run exactly.

PR #228, opened 2026-08-05
fix(claude-code): guard YAML/marketplace parsing, narrow commands/agents docs to array-only

vinnie357:
- claude-skills-246: REVERSED after Gate 3 found the initial (b) decision wrong — verified independently at code.claude.com/docs/en/plugins-reference before reversing: the manifest field table documents `commands`/`agents` as `string | array` with bare-string examples, and explicitly states a string "Replaces the default: commands, agents..." — upstream genuinely supports the string form
- Validator now accepts a bare string for `commands`/`agents`, existence-checked as a single path (not coerced into the array-entry loop, which enforces "must be a file" and calls into content validation) — avoids the naive-coercion hazard the original #223 investigation flagged, since the string form may legitimately be a directory this validator doesn't scan
- Restored `plugin-schema.md` and `SKILL.md` to `string | array`, matching upstream
- Flipped `commands_not_array_errors`/`agents_not_array_errors` to accept-direction, added wrong-type reject fixtures (a number, invalid under both accepted shapes) and clean-resolve accept fixtures
- claude-skills-247: guarded both `from yaml` calls (`validate-skill-md`, `validate-agent-md`) against two distinct shapes each — content that raises, and content that parses to a non-record without raising (the claude-skills-243 defect in YAML rather than JSON, which a try/catch alone cannot catch)
- Guarded the marketplace `open`, a third call site #227's record guards never covered, and replaced the bare `.plugins` cell-path access that crashed on a marketplace missing that key
- Fixed a fifth crash shape on the `plugins`-access line this PR rewrote: a `plugins` key present but wrong-typed (`"hi"`, or a list with non-record entries) crashed uncaught — guarded with a list/table type check
- Fixed a sixth shape on that same line, found by Gate 3 after the fifth was fixed: an entry that IS a record but has no `name` key still crashed, because nushell's `where name == x` raises `column_not_found` rather than skipping such a record. The round-1 two-stage form (record pre-filter, then bare `where name ==`) did not protect the name access at all. Replaced with a single closure doing the record check and a `get -o name` comparison together, which returns null instead of raising
- Added fixtures for both sixth-shape cases: a nameless record alone, and a nameless record FOLLOWED by a valid entry — the second is what proves the filter skips rather than aborts, since a filter that aborts and one that skips look identical when the bad entry is alone
- Added a genuinely-unparseable marketplace JSON fixture (`{"name": }`) — the existing lenient-scalar fixture only exercised the record-type check, never the try/catch's raise path, so neutering the try/catch alone stayed green until this fixture closed that gap
- Replaced four direct-call fixtures with CLI-subprocess-isolated equivalents: direct calls have no subprocess boundary, so a reverted guard crashed the whole self-test script instead of failing one case
- Fixed a fixture isolation nit (`no-plugins-key` marketplace fixture now carries `owner`, matching the valid fixtures, so it differs from them in exactly the one dimension under test)
- Every new/changed guard mutation-verified: neutered, confirmed the exact case fails by clean stdout assertion (not crash), restored, confirmed 104/104 green
- Bump claude-code 0.2.65 -> 0.2.66 and all-skills 0.4.126 -> 0.4.127 (unchanged from the original push — still one above live main)

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.126",
+      "version": "0.4.127",
       "category": "meta",
       "keywords": [
         "skills",
@@ -119,7 +119,7 @@
       "source": "./plugins/tools/claude-code",
       "strict": true,
       "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
-      "version": "0.2.65",
+      "version": "0.2.66",
       "category": "tools",
       "keywords": [
         "claude-code",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.126",
+  "version": "0.4.127",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/claude-code/.claude-plugin/plugin.json
+++ b/plugins/tools/claude-code/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-code",
-  "version": "0.2.65",
+  "version": "0.2.66",
   "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
   "author": {
     "name": "vinnie357"
--- a/plugins/tools/claude-code/skills/claude-plugins/SKILL.md
+++ b/plugins/tools/claude-code/skills/claude-plugins/SKILL.md
@@ -48,8 +48,8 @@ A manifest lives at `.claude-plugin/plugin.json` inside the plugin directory.
 Relative to the plugin root, `./`-prefixed. Each has a dedicated skill for its own file format:
 
 - `skills` — array of directories, each containing a `SKILL.md` (see `claude-skills`)
-- `commands` — array of `.md` files or directories (see `claude-commands`). A bare string was previously documented as a directory-to-scan shorthand; `validate-plugin.nu` never implemented that scan (claude-skills-246), so this is array-only.
-- `agents` — array of files (see `claude-agents`). Same correction as `commands` above.
+- `commands` — string or array of `.md` files or directories, replacing the default `commands/` scan (see `claude-commands`)
+- `agents` — string or array of files, replacing the default `agents/` scan (see `claude-agents`)
 - `workflows` — string or array of workflow script files or directories, replacing the default `workflows/` scan
 - `hooks` — string path to a hooks.json, or an inline hooks object (see `claude-hooks`)
 - `mcpServers` — string path to an MCP config, or an inline object
--- a/plugins/tools/claude-code/skills/claude-plugins/SKILL.md
+++ b/plugins/tools/claude-code/skills/claude-plugins/SKILL.md
@@ -48,8 +48,8 @@ A manifest lives at `.claude-plugin/plugin.json` inside the plugin directory.
 Relative to the plugin root, `./`-prefixed. Each has a dedicated skill for its own file format:
 
 - `skills` — array of directories, each containing a `SKILL.md` (see `claude-skills`)
-- `commands` — string or array of `.md` files or directories (see `claude-commands`)
-- `agents` — string directory or array of files (see `claude-agents`)
+- `commands` — array of `.md` files or directories (see `claude-commands`). A bare string was previously documented as a directory-to-scan shorthand; `validate-plugin.nu` never implemented that scan (claude-skills-246), so this is array-only.
+- `agents` — array of files (see `claude-agents`). Same correction as `commands` above.
 - `workflows` — string or array of workflow script files or directories, replacing the default `workflows/` scan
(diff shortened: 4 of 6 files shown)


vinnie357:
## Gate 3 REQUEST CHANGES — resolved

Independently verified the upstream citation before reversing anything, per the explicit ask not to take it on faith. Fetched code.claude.com/docs/en/plugins-reference directly: the manifest field table shows \`commands\`/\`agents\` as \`string | array\` with bare-string examples (\`"./custom/cmd.md"\`), and further down: "Replaces the default: \`commands\`, \`agents\`..." for the string form. This confirms upstream genuinely supports it — the original (b) decision (narrow our docs to array-only) was wrong, converting an honest "known divergence" comment into a codified falsehood, exactly as flagged.

**Reversed:**
- Validator accepts a bare string for commands/agents, existence-checked as a single path — NOT coerced into the array-entry loop (which enforces file-type and calls into content validation), avoiding the exact naive-coercion hazard the original investigation warned about
- Restored plugin-schema.md and SKILL.md to \`string | array\`
- Flipped the two \`_not_array_errors\` fixtures to accept-direction, added wrong-type reject fixtures (a number — invalid under both accepted shapes) and clean-resolve accept fixtures proving the string form is genuinely usable, not just type-accepted

**Blocker 2 fixed** — a fifth crash shape on the exact marketplace-access line this PR rewrote: \`plugins\` present but wrong-typed. \`{"plugins": "hi"}\` and \`{"plugins": [42, "x"]}\` both reproduced with zero mutation before fixing. Guarded with a list/table type check plus per-entry record tolerance — verified directly that \`where name == ...\` raises on the first non-record entry rather than silently skipping it, so entries are pre-filtered to records before the comparison.

**Minor fixed** — confirmed the mutation gap directly: neutering only the marketplace try/catch (keeping the record check) left the suite green, because the shipped \`"not json {{{"\` fixture parses leniently to a string and only exercises the record-type path. Added \`{"name": }\` (verified to genuinely raise) to close it.

**Nit fixed** — the no-plugins-key fixture now carries \`owner\`, matching the valid fixtures, isolating exactly the one dimension under test.

**New finding, out of scope for this PR:** upstream also documents \`skills\` as \`string | array\` (same table), and this validator's \`skills\` field currently rejects a bare string too (verified directly). Same defect class, third field, not part of 246/247's mandate — flagged in \`## BEES REQUESTS\` below rather than expanded into here.

Self-test: 102/102 green (95 -> 102: 2 flipped fixtures re-verified + 6 new: 2 wrong-type, 2 clean-resolve, 1 unparseable-marketplace, 2 plugins-wrong-type-shapes). Every changed/added guard mutation-verified failing by clean stdout assertion, not crash — confirmed for all of them individually, not asserted. \`mise run ci\` EXIT_CODE 0 via ci-runner, gitleaks clean, version-bumps confirmed still valid against live main (unchanged since original push). Gate 2 re-confirmed: headSha 8109233 matches both the PR and the CI run exactly.

## BEES REQUESTS (queued — I hold read-only bees this session)

1. New bee: \`skills\` field has the identical string|array documented-but-unimplemented gap that 246 just fixed for commands/agents — upstream documents it as \`string | array\`, this validator's skills handling only accepts an array (rejects a bare string), verified directly. Same fix shape as 246: accept the string, existence-check only, no coercion into the array-entry loop (which validates individual SKILL.md content per entry).

vinnie357:
## Gate 3 round 3 — sixth crash shape, fixed

Independently verified before touching anything: \`[{foo: 1}] | where name == "x"\` raises \`column_not_found\` — nushell does NOT silently skip a record missing the compared column, confirmed both in isolation and against the actual validator on unfixed code.

**The two false claims are corrected.** The prior code comment said "a malformed entry is tolerated (skipped), not crashed on" — true for non-record entries, false for a record missing \`name\`. Rewrote the guard as a single closure that checks record-ness AND reads \`name\` via \`get -o\` (returns null rather than raising) in one step, replacing the two-stage \`where {record} | where name ==\` that let a nameless record through the first filter only to crash on the second. The comment now documents all three sub-shapes explicitly and states plainly that the old claim was wrong for this one — same correction applied to the PR body.

**Two new fixtures**, per the ask: a nameless-record-alone case (must not crash, falls through to "not found"), and a nameless record FOLLOWED BY a valid matching entry — proving the filter genuinely skips the bad entry and keeps evaluating rather than aborting the whole list. Building the second one caught a real fixture-isolation bug of my own: the marketplace file needs to live at \`<root>/.claude-plugin/marketplace.json\` for \`marketplace_dir\` resolution to work (verified against the one WORKING marketplace fixture's structure) — my first attempt used a flat file like the earlier "fails before reaching resolution" fixtures, which silently broke source lookup. Fixed with a self-contained subtree mirroring the working fixture's layout.

Mutation-verified: reverted to the two-stage filter, confirmed both new fixtures fail by clean stdout assertion (self-test completes, reports both) — not a crash — restored, confirmed 104/104 green.

**Confirmed, not touched:** the string→missing-path warn (not error) behavior matches the array path's existing convention. The wrong-typed \`source\` crash on a matching entry — same pre-existing lines on main — is out of scope, filed separately per the note.

Self-test: 104/104 green (102 -> 104). \`mise run ci\` EXIT_CODE 0 via ci-runner, gitleaks clean, version-bumps confirmed unchanged and still valid (0.2.66/0.4.127 against live main 0.2.65/0.4.126). Gate 2 re-confirmed: headSha a8b13d7 matches both the PR and the CI run exactly.

PR #246, opened 2026-08-07
fix(security): make gitleaks invocation shadow-proof and fix non-git fail-open

vinnie357:
- Add `--no-git` when the scan path is not a git work tree, in both the native and container arg builders — previously `gitleaks detect --source=<non-git-dir>` logged a git fatal, scanned 0 bytes, printed "no leaks found" and exited 0, so `gitleaks.nu` reported a green gate having scanned nothing
- Detect work trees via `git rev-parse --is-inside-work-tree` rather than a literal `.git` directory check, so worktrees and submodules resolve correctly
- Add native-binary resolution (`mise which gitleaks` then PATH) ahead of the container fallback in the `templates/mise.toml` gitleaks task, matching the precedence `gitleaks.nu` already documents and the decision recorded in root `mise.toml:11`
- Name the shadow-proof invocation at the point of first teaching in `security/SKILL.md` instead of only in a later subsection, and make every example use the resolved binary path
- Replace deprecated `detect`/`protect` examples with the current `git`/`dir` subcommands, noting both still work as unlisted aliases
- Name the safe invocation in `git/SKILL.md` Gate 1 and the PR workflow, and record that the PreToolUse hook already scans staged changes automatically
- Add `test/validate-gitleaks-invocation.nu` covering the fail-open regression, `--no-git` arg construction, and template native-first ordering; wire it into `mise run ci`
- Bump core to 0.2.58, all-skills to 0.4.145, and regenerate the skills-quality baseline


vinnie357:

Reviewed by a separate agent on the strongest available model, briefed to defeat the change. Three blockers, independently reproduced by the lead before dispatching fixes.

### Blockers

**B1 — test 3 compares two prose matches; it passes on a template with zero native resolution.** `test/validate-gitleaks-invocation.nu:124` uses marker list `["mise which gitleaks" "which gitleaks" "gitleaks.nu" "native"]`. The body slice starts at the task header, so bare `native` matches the `description` line, and `container run` matches the prose "container runtime" in a comment. Demonstrated by mutation: a template with no native resolution and an unconditional `container run` still passes. Confirmed.

**B2 — the fail-open is not closed generally, only the non-git-directory instance.** A zero-commit git repo containing an untracked secret returns `--is-inside-work-tree: true`, so `--no-git` is not added, gitleaks scans 0 bytes, and the script reports "No secrets detected" at exit 0. Reproduced by the lead. `run-gitleaks-native` prints only stdout on the success path, discarding the `scanned ~N bytes` line on stderr that would make a zero-scan visible.

**B3 — the template rewritten by this PR still ships the unfixed arg shape.** `grep -c -- "--no-git"` on `templates/mise.toml` returns 0. Two arg builders fixed, a third edited in the same commit left broken — and B1 means nothing catches it. Confirmed.

### Non-blocking

- **F4 (medium)** — `git/SKILL.md:99,117` state the PreToolUse hook scans on every commit. Skill-frontmatter hooks are scoped to that skill's lifecycle, so it fires only while `/core:security` is loaded, and the hook fails open when no scanner is available. An instruction was turned into an assurance an agent can rely on to skip scanning.
- **F5 (medium)** — four `detect --source=` blocks remain in `security/SKILL.md` (~301, 316, 331, 385); the PR body claims the replacement file-wide.
- **F6 (medium)** — test 1 asserts a bare `exit_code == 1`, which also matches "no gitleaks binary found".
- **F7 (low, restraint)** — ~45 lines of resolution logic duplicated into the template; it already drifted within this PR (B3).
- **F8 (low, informational)** — bare repo adds `--no-git` and scans packed objects; not worth fixing.

### Verified as holding

`--no-git` present in both `gitleaks.nu` arg builders; `git rev-parse --is-inside-work-tree` used as claimed; template is native-first in code; `detect`/`protect` confirmed working unlisted aliases on 8.30.1; version bumps consistent across `plugin.json` and `marketplace.json`; baseline regen correct; disclosure lint clean; `hooks/check-secrets-before-commit.sh` genuinely immune (passes `--no-git`, fails closed on unexpected exit codes); fixture token still detected at runtime with the source file clean.

Remediation dispatched. Not merging until re-review.

vinnie357:
## Gate 3 round 2 — REQUEST CHANGES

Re-review of `5f8bef4` by the same adversarial reviewer. `mise run ci` green and 16/16 self-tests; neither catches what follows. Blockers reproduced by the lead before dispatch.

### Blockers

**R1 — the tightened tests are defeated again, by putting the markers in a comment.** Round 1's fix moved the slice past the `description`, but Cases 3 and 5 still substring-match over text including comments. A template with no native resolution and no `--no-git`, with both marker strings in a comment, passes both. The defects B1 and B3 were filed for survive the tests written to prevent them. Confirmed.

**R2 — the bytes invariant's comment overclaims.** A repo with one commit plus an uncommitted live secret scans 20 bytes of history, satisfies `scanned_bytes > 0`, and reports clean — the secret never examined. Reproduced by the lead. Bytes-scanned proves *something* was scanned, not that the requested scope was covered. Operator decision: correct the claim to match what the check delivers (zero-scan and unparseable-summary cases). The working-tree-vs-history design question is filed as `claude-skills-268`.

**R3 — `scripts/gitleaks.sh` is untouched and has all three defects.** Absent from the diff; `--no-git` count 0, native-resolution count 0, bytes-verification count 0. `SKILL.md:260` presents it as an alternative while warning only about the native gap. Confirmed by the lead. Operator decision: bring it to parity.

### High / medium

- **R4** — the template received `--no-git` but not the bytes invariant; the zero-commit fixture is still green there. It is what agents copy into new projects.
- **R5** — reusing `Secrets detected!` for the unverified branch is an anti-fabrication violation, and the stated rationale (external callers gate on it) is unsupported: repo-wide grep finds only this PR's own tests as consumers. It has already cost test strength — Case 4 cannot distinguish detection from non-verification. Being split to a distinct `Scan unverified!` marker.
- **R6** — `parse-scanned-bytes` takes the FIRST `scanned ~` match, so an earlier stderr line containing that substring overrides the real summary. Fails OPEN. Confirmed by probe.
- **R7 (UNVERIFIED)** — if a container runtime merges stderr into stdout, every clean container scan would print the unverified marker and exit 1. Reviewer could not test without starting a restricted runtime.
- **R8** — the invariant scrapes an undocumented log string while `gitleaks = "latest"` is unpinned. Filed as `claude-skills-269`.
- **R9** — the three runtime examples teach `git /code` only; a non-repo or zero-commit mount hits the same fail-open.

### Confirmed resolved since round 1

F4 (wording verified accurate on both hook scoping and fail-open), F5, F6 partially, F7, B3's `--no-git` half, and the hook's remediation line. No regressions: a clean repo with committed content still exits 0, and `main` independently re-derived clean at 369 commits.

Round 3 remediation dispatched.

vinnie357:
## Gate 3 round 4 — REQUEST CHANGES (narrow)

Re-review of `3a60028`. **The three structural classes are closed** — first round the reviewer could not defeat a guard. Remaining findings are surface reconciliation.

### Confirmed resolved
- **T1** — all four historical mutation shapes (description, comment, bare string literal, caret-bearing literal in a dead branch) are structurally dead. The tests execute TOML-parsed task bodies as subprocesses; a comment or string literal cannot alter subprocess behavior. The offset/marker/comment-stripping machinery is gone.
- **T2** — every round-3 divergence input now agrees across all three implementations.
- **T3** — one implementation, dynamic task enumeration.
- **T4/T5/T6** — realistic decoy handled, EXIT trap correct, report-path divergence fixed with a named `restraint:` boundary.

### Change requests
- **U1 (medium-high)** — `mise gitleaks:docker` / `gitleaks:colima` do not use Docker or Colima. Native resolution runs first and `GITLEAKS_RUNTIME` is only read in the unreachable container fallback. Verified by running `gitleaks:docker` with the Docker daemon down and getting a clean native scan. The mechanism is sound and native-first is correct (claude-skills-209); the task descriptions and `SKILL.md:439-440` are what need reconciling.
- **U2 (medium)** — the template ships no `[tools]`, so `mise run gitleaks` fails with `env: nu: No such file or directory` (exit 127) in a fresh project. Precedent already merged into this branch's base: `5f85531` (PR #245).
- **U3 (medium)** — a lead judgment call was wrong on the facts. Two full-shaped summary lines CAN occur in one run: a file whose name carries the shape produces a `WRN skipping file:` line matching the anchor. The last-match choice remains correct, but only because gitleaks emits all warnings before the summary — an ordering invariant that was nowhere documented. Being named in all three implementations, with an explicit ordering fixture restored.
- **U4 (low)** — `safe-minimal-path`'s comment claims it excludes real mise; real mise leaks in via `which git`'s directory. Protection is actually the stub-dir prepend, which is load-bearing and unmarked.

### Known blind spot, recorded
The conformance test proves the three implementations do not drift. It does not prove they are right — all three agree on `No secrets detected` for a repo with one commit plus an uncommitted secret. That is `claude-skills-268`, filed and deferred by operator decision.

### Also outstanding
Gate 2 failed remotely on `[tasks."gitleaks:colima"]` while local CI was green — an environment-dependent test. Being reproduced locally and diagnosed; not rerun.

vinnie357:
## Gate 3 round 5 — REQUEST CHANGES (narrow, 2 items)

Re-review of `de79fbd`. Both gates green including Linux.

**The reviewer states the security substance is converged, not merely quiet** — it could not defeat any scanner guard this round. No fail-open remains in any of the three implementations, and the ordering dependency is named and pinned. Neither remaining item touches the scanner.

### Change requests

- **V1 (medium)** — the Linux colima assertion degraded to "exited nonzero and didn't claim clean". Confirmed by evaluating the predicate against real result records: a crash on an undefined variable, and taking the wrong arm (`❌ Docker CLI not found`), both pass. Delete the colima arm or put a syntax error in it and CI stays green; only a macOS developer sees red. It still catches the dangerous silent-clean case. Fix is to assert the gate's distinctive string rather than a nonzero exit.
- **V2 (medium)** — the `[tools]` block added for standalone operation breaks any consuming `mise.toml` that already declares nushell: TOML rejects the duplicate key regardless of value, taking down every task in that project. It collides with this marketplace's own beads template, which declares the identical key at the identical version. Fix is merge-don't-append guidance in both the template and SKILL.md — not removing `[tools]`, since standalone operation is verified working.

### Lower

- **V3 (low)** — the ordering fixture pins a known-wrong verdict; its failure message reads `expected true, got false`, so someone who correctly hardens the parser sees what looks like their own regression and reverts it.
- **V4 (low)** — colima labels are accurate on precedence but never state Colima is macOS-only.

### Confirmed resolved

U1 across all five surfaces, U2 standalone operation (independently reproduced), U3 ordering invariant named in all three implementations with a live-reproduced fixture, U4 PATH comment corrected and the prepend marked load-bearing. The colima root cause was verified as the OS gate, not native resolution.

### Review boundaries, stated by the reviewer

Recorded so coverage is judged rather than assumed: **no real container runtime was ever exercised** across five rounds (Docker daemon down, `container run` restricted — all container verification is via stubs or reading, apart from one live Apple Container test); **no Linux execution** (V1 is analysed, not run; Gate 2's green is the only Linux evidence); and never examined — `gitleaks.sh`'s argument-parsing loop, the `--config`/`--baseline` paths end-to-end, the `gitleaks:stop*` teardown tasks, `check-secrets-before-commit.sh` beyond round 1, and `SKILL.md` outside the gitleaks sections.

vinnie357:
## Gate 3 round 6 — APPROVE

Re-review of `2250f5a` by the adversarial reviewer. V1–V4 all sound, no regressions, and three of the four stated coverage boundaries closed this round.

Regression matrix identical across all three implementations: clean repo → exit 0 `No secrets detected`; zero-commit + secret → exit 1 `Scan unverified!`; non-git + secret → exit 1 `Secrets detected!`.

### Boundaries closed this round
- **`gitleaks.sh` argument-parsing loop** — every malformed shape exercised; all fail closed.
- **`--config` / `--baseline` end-to-end** — both validate existence and error before scanning; behaviour identical in nu and sh.
- **`gitleaks:stop*`** — all four bodies read; they stop runtimes only, gated on `which` plus a status probe. No unfiltered delete patterns.
- **`check-secrets-before-commit.sh`** — closed by diff: one changed line since main (the round-3 remediation-message fix), so the round-1 review stands.

### Residual, none blocking
- **W1 (low-medium)** — the template and `SKILL.md:434` give contradictory merge instructions, and neither covers a consumer whose nushell pin is *older* than 0.113.1. Being fixed. Measured, not assumed: nushell 0.107.0 ran the template task and 21/21 self-tests correctly, so there is no known breakage — the gap is documentation, not function.
- **W2 (low)** — the colima gate-message extraction matches any line containing the phrase, so a *comment* carrying it causes a false RED. Fails closed. Being fixed.
- **W3 (nit)** — `gitleaks.sh`'s arg loop dereferences `$2` unchecked. Every malformed shape across both scripts exits 1 and none claims a scan, so it fails closed. Filed as `claude-skills-270` rather than touching bash arg parsing at approval time.

### What this approval covers
Verified by repeated execution on macOS against real gitleaks 8.30.1: no fail-open in any of the three implementations for non-git paths, zero-commit repos, zero-byte scans, unparseable summaries, or empty directories; identical verdicts across all three on every constructed input; arg-builder and flag handling; documentation accuracy; and test strength — the four historical template-mutation shapes are structurally dead, and the colima predicate is satisfiable only by the real gate firing.

### What it does not cover
- **No real container runtime, across six rounds.** Docker's daemon was down throughout and `container run` is restricted; every docker/container/colima claim rests on stubs and reading, apart from one live Apple Container test.
- **No Linux execution by the reviewer.** Gate 2 green is the Linux evidence.
- **`claude-skills-268`** — git mode scans committed history, so a secret sitting uncommitted in a repo with any history is still reported clean by all three. Filed by operator decision, not fixed here. That is what "green" means for this gate.
- **`claude-skills-269`** — unpinned gitleaks, coupling to an undocumented log-line format.
- **Older nushell** beyond the single 0.107.0 spot-check.

PR #247, opened 2026-08-07
docs(security): document gitleaks.nu's committed-history-only scan scope

vinnie357:
- State that `gitleaks.nu`'s git-mode scan walks committed history and does not examine uncommitted working-tree changes
- State what a clean result establishes: nothing in the scanned history matched, not that the repository holds no secrets
- Name the `check-secrets-before-commit.sh` PreToolUse hook as the staged-case coverage, with both its preconditions — it fires only while `/core:security` is loaded, and fails open when no scanner is available
- Give the manual escape hatch for checking uncommitted content (`dir` mode) and explain why it stays manual rather than becoming the automatic gate
- Closes claude-skills-268


vinnie357:
## Gate 3 — REQUEST CHANGES

Doc-only scope confirmed: one SKILL.md hunk plus consistent version bumps (core 0.2.59, all-skills 0.4.146 across both `plugin.json` files and `marketplace.json`). CI green, 21/21 self-tests, disclosure clean.

The central claim was verified against the document's own fixture — one committed file, one uncommitted, one gitignored `.env`, each with a live-shaped token: git mode reports "No secrets detected", `dir . -v` finds 2. Claims about what a clean result establishes, and the manual `dir` command with its rationale, both hold by execution.

### Blocking

**X1 — the section overstates its own compensating control.** It says the PreToolUse hook scans staged content "on every `git commit`". The hook reads `git diff --cached`, so it only sees what is already in the index when the hook fires; `-a` and the pathspec form stage *during* the commit, after it has run.

| command | hook | exit |
|---|---|---|
| `git commit -m x` | No staged files to scan | 0 |
| `git commit -a -m x` | No staged files to scan | 0 |
| `git commit -am x` | No staged files to scan | 0 |
| `git commit <path> -m x` | No staged files to scan | 0 |
| control — secret staged | **SECRETS DETECTED** | **2** |

Reproduced independently by the lead: with a tracked file modified but not staged, `git diff --cached --name-only` is empty. A section whose purpose is calibrating trust must not overstate the control it points at.

**This is pre-existing hook behaviour, filed as `claude-skills-271` (P1)** — the fix belongs in the hook. This PR corrects the claim and references the issue.

### Also

- **X2** — the note sits under `### gitleaks.nu` only. `### gitleaks.sh` is a sibling immediately after it, and `## Mise Tasks Template` is a different top-level section; both share the identical git-mode behaviour, enforced by #246's conformance test. Same one-artifact-not-its-siblings pattern that PR #246 spent six rounds on, in documentation form.
- **X3** — "inside a git repository" is false for a bare repo (takes the `--no-git` path, never walks history) and silent about the non-git case, where coverage is *inverted* — a full content scan that catches the secret. Verified across seven repo shapes; holds for repos with commits, worktrees, submodules, and subdirs.
- **X4** — a more precise version of this note already exists at line 293 from #246's R9 fix, covering both shapes the new note gets wrong. Restraint question raised: amend and reference that one rather than carry two notes at different precision.
- **X5** — `### Basic Usage` gives the document's first runnable command carrying this same limitation, ~96 lines above the note.

vinnie357:
## Gate 3 — APPROVE

Round 4 review of `a1ae590`. The corrected boundary could not be falsified in either direction, and the parenthetical ("not even the 'no staged files' message") is literally true — `git -C <dir> commit` produces zero bytes of hook output.

**Why the enumeration trap is now closed structurally:** earlier revisions listed *instances* with no rule, so an unlisted instance falsified the text. It now states the **rule** — the trigger regex is anchored at position 0 — and offers five shapes as illustrations of it. A sixth (alias, `GIT_DIR=` prefix, subshell) is covered by the rule rather than contradicting it.

No regressions across four revisions: one `#### Scan scope` heading, four cross-reference sites intact, Y2's scoping intact, line 297's note untouched, versions consistent, CI green, disclosure clean.

### A third hook bypass class was found and routed to claude-skills-271

**It is the worst of the four, because it emits a positive clean verdict over a file it never read.** A staged file with a non-ASCII name: `git diff --cached --name-only` emits git's quoted form (`"caf\303\251.txt"`, `core.quotePath=true` being the default), the hook feeds that literal to `git show ":$file"`, it fails, and `|| true` swallows it. Result: `scanned ~0 bytes … no leaks found` → **exit 0, commit allowed**. Control with the same token in an ASCII filename → exit 2, blocked. Independently reproduced by the lead.

The shape — zero bytes scanned, reported clean — is exactly the fail-open PR #246 spent six rounds closing in `gitleaks.nu`, `gitleaks.sh`, and the template. **The hook is the one implementation that never received the bytes-verification invariant.** Adding it there closes this class and any future variant where the export step silently produces nothing.

It does not block this PR: the section already states coverage is partial, hedges explicitly, and directs readers to the manual `dir` scan, so no reader's behaviour changes. The fix belongs in the hook.

### What this approval covers

The scan-scope note is accurate for every repo shape constructed — working tree with commits, zero-commit, bare, non-git, worktree, submodule, subdirectory. Three-implementation parity holds as written. Both documented bypass classes are correct and reproduce. The three prescribing sites are real. One note, four entry points.

### What it does not cover

**The gate itself is unfixed.** `claude-skills-271` is open and this PR ships documentation, not a fix. The hook still misses `git add . && git commit`, `git -C … commit`, `-a`/`-am`/pathspec, and silently reports clean on a staged non-ASCII filename.

`claude-skills-268` remains the underlying limitation by operator decision: `git` mode scans committed history, so an uncommitted secret in a repo with history is reported clean. That is what a green gate means here.

No real container runtime was exercised in any round of either PR.

PR #248, opened 2026-08-07
fix(security): close four commit-hook bypasses and add the bytes invariant

vinnie357:
- Replace the position-anchored `^git[[:space:]]+commit` trigger, which never fired for `git add . && git commit` (the form `/core:git`, `/core:bees` and CLAUDE.md all prescribe), `cd … && git commit`, leading whitespace, an env-var prefix, or a global option such as `git -C <dir> commit`
- Read staged paths with `-z` null separation so a non-ASCII filename is no longer emitted in git's quoted form, fed to `git show` where it fails, and silently swallowed — that path produced an affirmative "no secrets detected" over a file never read
- Export tracked working-tree modifications alongside index content so `git commit -am`, `git commit -a -m` and `git commit <path>` are scanned; that content stages during the commit and is invisible to `git diff --cached` when the hook fires
- Port `gitleaks_verified_clean` from `scripts/gitleaks.sh`: refuse to report clean unless the scan covered nonzero bytes, treating an unparseable summary as no evidence rather than innocence. The hook was the only implementation missing the invariant PR #246 added to `gitleaks.nu`, `gitleaks.sh` and the template
- Stop echoing gitleaks stdout, which contains matched secret values, through the hook's own output
- Extend `test/validate-security-hook.nu` to 19 cases covering all four bypass classes, both control directions, and the no-evidence case; make nine stub fixtures emit realistic gitleaks output instead of silence
- Closes claude-skills-271


vinnie357:
## Gate 3 — REQUEST CHANGES

Two blockers, both outside the 19-case suite, both reproduced by the lead.

### A1 — `git add . && git commit` still misses a brand-new secret file

The trigger now fires; the export then finds nothing. An untracked file is invisible to both passes:

```
untracked brand-new.txt with a live token:
  git diff --cached --name-only      -> []
  git diff --name-only               -> []
  git ls-files -o --exclude-standard -> [brand-new.txt]   <- only this sees it
```

Measured end-to-end: `git add . && git commit`, `git add .bees/ && git commit`, `git add -A && git commit`, and an untracked file in a subdirectory all report "No staged or modified-tracked files to scan" and **exit 0**. Controls hold — a tracked-modified secret and a staged secret both block at exit 2.

A brand-new file is the most common way a secret enters a repo, and it is exactly what `git add .` stages. **This is the headline case of the issue, still open.**

### A2 — routine deletions and empty files are now blocked

```
git rm gone.txt && git commit -m "remove x"    -> Scan unverified!  exit 2
touch empty.txt && git add empty.txt && commit -> Scan unverified!  exit 2
```

A path is listed so the file-found flag is set, but `git show ":$file"` fails on a deleted path and an empty file exports zero bytes. The invariant — correct in the scanning scripts, where zero bytes means the scan failed — reports "unable to verify no secrets are present", which is false: there was nothing to scan.

This is the disable-the-hook risk. Over-broad *matching* isn't it — a spurious block only occurs when a real secret is staged. A gate that blocks `git rm && git commit` is.

Both defects share one seam: the export set is chosen by path and content volume is never considered. Distinguishing "nothing to scan" (allow) from "scanned nothing unexpectedly" (block) fixes A2; a third `git ls-files --others --exclude-standard -z` pass fixes A1. Part 7 currently locks in A2's behaviour, so it needs a carve-out rather than deletion.

### Verified as holding

The #247 non-ASCII class is closed (blocks at exit 2 under both `bash` and `/bin/bash` 3.2.57). The invariant port is **byte-identical** to `scripts/gitleaks.sh` and all four implementations agree on every PR #246 fixture. Secrets are no longer echoed — zero occurrences of the token prefix in hook output on a detecting run. The `mapfile` rejection was correct (`mapfile: command not found` on bash 3.2.57). No false negatives found against the new trigger: every real commit form constructed matches, including multi-line commands.

**B1 (low, accepted)** — the unanchored trigger matches unrelated commands (`git log --grep=commit`, a PR body mentioning commit). With no finding pending these cost only scan latency; when one is pending the block is factually true. That is the fail-toward-scanning trade, deliberately taken.

### Not examined

The container/docker/colima branches (Docker down, `container run` restricted — read, not executed); mutation-testing of the 19 existing cases; `cp` following a staged symlink to its target; export performance on a large repo.

vinnie357:
## Gate 3 — APPROVE

Re-review of `39ddcc2`. A1 and A2 both closed, and each verified by **mutation** rather than by running the suite.

### Mutation evidence

- Dropping `--exclude-standard` from the hook turns Part 13 RED (`want exit 0, got 2`).
- Making the untracked pass yield nothing turns Parts 9, 10 and 14 RED (`want exit 2, got 0`).

Both new mechanisms detect their own removal — which matters here, because a green suite in this PR was previously resting on a compromise that reopened the headline bug.

### Verified behaviour

**Blocks correctly**: staged secrets; unstaged tracked secrets; untracked new files at root and in new nested directories; non-ASCII filenames; `git stage` as well as `git add`; every trigger shape from mechanism 1. Sparse checkout tested rather than skipped — a secret inside the cone blocks at exit 2. A nested repo's exit 0 was checked and is *correct*: the parent records only a gitlink, so there is nothing to scan.

**Allows correctly**: clean commits; pure staged deletions; empty staged files; gitignored files at rest; non-commit commands with nothing pending.

**`HAS_EXPECTED_CONTENT` probed in both directions** — staged-empty then written non-empty on disk blocks (exit 2, the unstaged pass catches it); staged non-empty then truncated on disk blocks (the index blob still has content); staged symlink allows correctly. The metadata check running *before and independent of* the export is what makes this hold.

**No path collisions**: the same path staged with one secret and worktree-modified with a different one produces `staged/dual.txt` + `unstaged/dual.txt`, 2 findings, exit 2. Neither version lost.

`gitleaks_verified_clean` remains byte-identical to `scripts/gitleaks.sh`; all four implementations still agree.

### Residual, filed not fixed

**C1 — `git add -f <gitignored> && git commit` goes unscanned.** Filed as `claude-skills-272`. `--exclude-standard` is correct for the default staging set (dropping it reintroduces the `.env` false positive `claude-skills-268` rejected), but `-f` is the one moment a gitignored secret genuinely enters history. The gap predates this PR and is narrow: a force-added file lands in the index, so a *separate* `git add -f` then `git commit` is already scanned today — only the compound form escapes.

### What this approval does not cover

- **The container / docker / colima branches — never executed in any round of #246, #247 or #248.** Docker's daemon is down and `container run` is restricted; those arms have only ever been read.
- The suite was mutation-tested at two points, not twenty-five.
- `cp` following a staged symlink to its target.
- Performance of three export passes on a large repo, now that the hook runs on any command matching `git.*commit`.
- `claude-skills-268`'s underlying limitation is unchanged and out of scope by decision.

PR #249, opened 2026-08-07
docs(security): document the force-add gitignored gap instead of chasing it

vinnie357:
Documents a bounded gap rather than fixing it. Two predicates were attempted and each blocked legitimate commits, so the widening was reverted on operator decision.

**The gap:** a secret force-added to a gitignored path in the *same compound command* as the commit is not scanned. The untracked pass uses `--exclude-standard`, which correctly hides the file at the moment the hook fires — before the `add -f` has run.

**Why it is narrow:** once force-added, the file is in the index, so a *separate* `git add -f` followed by `git commit` is caught by the existing staged pass. Verified, and pinned as Part 20.

**Why it is not fixed.** Dropping `--exclude-standard` unconditionally means every commit scans every gitignored file — the false positive `claude-skills-268` decided against, and on a repo with `node_modules` or `target/` it copies and scans the whole ignored tree. Two predicates were tried to widen the scan only on a force-add:

| attempt | defeated by |
|---|---|
| `-f` as a bare substring | `git commit --fixup=HEAD`, `git commit --file=msg.txt` |
| `-f` as a whitespace-bounded token | `rm -f tmp.txt && git commit`, `grep -f patterns.txt src && git commit`, `docker build -f Dockerfile . && git commit`, `tar -x -f archive.tgz && git commit` |

Each blocked ordinary commands in any repo keeping a real secret in `.env`, reporting SECRETS DETECTED against a file the user deliberately ignored. The underlying reason: a command string does not carry the structure needed to say whose flag a `-f` belongs to.

**Changes**

- Revert the fourth export pass and its predicate; the three passes from #248 remain unconditional and unchanged
- Record both attempted predicates and their exact defeating cases in a comment at the removal site, so a third attempt starts from the evidence
- Convert Parts 15-17 to documented-limitation pins asserting the gap, with instructions to flip the expectation if a safely-scoped detector ever closes it rather than treat it as a regression
- Keep Parts 22/23 as regression guards against reintroducing a predicate
- Document the gap in `/core:security` alongside the existing scan-scope material: what is not scanned, that the separate-command form is, why it is not fixed, and the manual `dir` check for anyone force-adding a gitignored file

Closes claude-skills-272


vinnie357:
## Gate 3 — APPROVE

No findings. The revert is provably clean: `git diff 38dffd9..HEAD -- check-secrets-before-commit.sh` is **46 insertions, 0 deletions, 0 non-comment added lines** — the executable body is byte-identical to PR #248's merged state, so the three passes cannot behave differently. Every `force-added`/`--force` match sits inside the comment block.

**No regressions**, verified by live re-execution as well as by diff: untracked file (2), untracked in subdirectory (2), staged deletion (0), staged empty file (0), `git stage` synonym (2). `HAS_EXPECTED_CONTENT` intact; `parse_scanned_bytes` and `gitleaks_verified_clean` still byte-identical to `scripts/gitleaks.sh`.

**The limitation pins are honest on two independent grounds.** Structurally, Parts 15-17 hand the stub a scanner rigged to report a leak (exit 1) and assert exit 0 — so passing can only mean the scanner was never invoked, which is exactly the short-circuit being pinned. The Part-13-under-the-gate failure mode is structurally excluded. By mutation, dropping `--exclude-standard` flips 15, 16, 17 red along with 13, 18, 21, 22, 23, 24.

**Documentation accurate in both directions**: the compound form produces the exact quoted message at exit 0; the separate-command form blocks at exit 2. No overstatement of remaining coverage.

### What this does not cover

**`claude-skills-272` is an accepted OPEN gap, not a closed one.** `git add -f <gitignored> && git commit` in one command still commits a secret unscanned. Read "#249 merged" as *documented and pinned*, not fixed. The manual mitigation is `$(mise which gitleaks) dir <path> -v` before force-adding.

`claude-skills-273` — ~30 of 35 cases still unprobed for passing-for-the-wrong-reason. The container/docker/colima branches remain never-executed across #246-#249. `cp` following symlinks and export performance on a large repo are unexamined. `claude-skills-268`'s scan-scope limitation is unchanged by decision.

PR #250, opened 2026-08-15
fix(elixir): ports skill {:env, list} extends the environment, not replaces it

vinnie357:
- Correct the `{:env, env_list}` semantics in 6 places (5 in SKILL.md, 1 in references/erlang-ports.md) — OTP documents that it EXTENDS the spawned process's environment, it does not replace it
- Rewrite the Security section: a short `env:` list does not sandbox the child, since it inherits everything the BEAM holds; withholding requires pairing a name with `false`
- Delete the `System.get_env() |> Map.to_list()` merge example, which was built on the inverted premise
- Retitle the Common Pitfalls entry to name the real trap
- Add an Environment Inheritance section to references/erlang-ports.md with the OTP doc quote, a deny example, and a minimal allowlist helper with its limitation stated
- Bump elixir 0.1.22 → 0.1.23 and all-skills 0.4.148 → 0.4.149

Evidence: OTP 28.4.2 `erts-16.3.1/src/erlang.erl:7600-7610` — "The environment of the started process is extended using the environment specifications in `Env`." Identical wording in OTP 27.3.4.13 and 28.0.1, so no version drift. Typespec at :7796 confirms `false` / `[]` as unset forms.

Verified by loading the skill via `claude -p --plugin-dir` (returns EXTEND; the 0.1.22 cache returns REPLACE, establishing the positive control). `mise test` exit 0, gitleaks clean.


vinnie357:
## Gate 3 adversarial review — APPROVE (no blocking findings; 1 minor content fix suggested)

Reviewed at pinned head `6af15e2`. Note: `gh pr diff 250` initially served stale content (it still contained "binaries are accepted in practice" and named `Port.info/1` as an exposure surface — both already corrected at head). Everything below is verified against `6af15e2` via `git show`, no checkout.

### Primary-source verification (attack 1 — overcorrection)
- `erlang.erl:7600-7602` (OTP 28.4.2, erts-16.3.1): "The environment of the started process is **extended** using the environment specifications in `Env`." Quote in references/erlang-ports.md matches verbatim. CONFIRMED.
- Typespec `:7796` matches the reference's quoted spec character-for-character. CONFIRMED.
- "There is no replace-all option": read the full `Opt` union (`:7786-7809`) — no env-clearing form exists. CONFIRMED.
- "Binaries raise `ArgumentError`": empirically confirmed, all four shapes — `{binary, binary}`, `{binary, charlist}`, `{charlist, binary}`, `{charlist, ""}` each raise `badarg` "invalid option in list" on OTP 28.4.2 (boolean-token marker probes, no env listing emitted). The quoted error string matches exactly. CONFIRMED.
- Typespec asymmetry claim: `cd`/`args`/`arg0` declare `string() | binary()`, `env` declares `os:env_var_name()` = `nonempty_string()` (os.erl:83). CONFIRMED.
- `System.cmd/3` contrast: `system.ex:1235-1246` — `{k, nil} -> {to_charlist(k), false}`; passing `false` hits `String.to_charlist(false)` which raises. CONFIRMED.

### Empirical semantics (marker probes, boolean output only)
- Short `env:` list → parent marker INHERITED (extend confirmed; replace would have shown ABSENT).
- `{name, false}` → ABSENT. `{name, ~c""}` → ABSENT (unset, not empty — supports the "empty value cannot be passed through" sentence).
- Duplicate names: later entry wins (`[{m, false}, {m, ~c"1"}]` → INHERITED; reversed → ABSENT). So the allowlist helper's `denied ++ extra` ordering is correct on OTP 28.4.2.

### Findings
1. **[suggestion, minor] SKILL.md:73** — "pair its name with `false` (`\"\"` and `[]` also unset)". In this Elixir-syntax context `""` is a binary, and a binary Val raises `ArgumentError` (probe-confirmed). Line 340 and the reference correctly attribute `""` to Erlang notation (≡ `[]`); line 73 stands alone and will send a reader to a badarg. Suggest `~c""` or "the empty charlist `[]`". Fail-loud, so minor — not blocking.
2. **[nit] references/erlang-ports.md** — "The asymmetry is **deliberate**": intent is not verifiable from a typespec; the observable is only that the specs differ. Suggest "The asymmetry is in the typespec itself".
3. **[suggestion] allowlist helper** — when a name in `extra` is also in `denied`, correctness depends on later-entry-wins for duplicate names, which OTP does **not** document (verified empirically on 28.4.2 only). Worth either a one-line note or filtering `denied` by the names in `extra`.
4. **[plausible, unverified] Windows** — env names are case-insensitive there; the allowlist compares case-sensitively (`Path` vs `PATH` would be denied). No Windows host available to verify; noting, not asserting.

### Attack 5 — the framing judgement: UPHELD
The child's environment is byte-identical whether the parent's vars are re-passed or merely inherited (probe: marker INHERITED with a short list), so "exposes sensitive OS environment variables" to the *child* was overstated — the deletion rationale (teaches the inverted model + materializes secrets into BEAM terms) is the correct one. The materialization surfaces are real: my own failing `Port.open` probe echoed the **entire env list into the `ArgumentError` message** — direct evidence for the "exception messages" surface. `:sys.get_state/1` is conditional on the list being stored in state, appropriately hedged with "can surface". `Port.info/1` (present in the stale revision) is correctly gone: `port_info/1` and `/2` specs (`erlang.erl:12077-12142`) contain no env item.

### Attacks 2 & 4 — residual inversion / completeness: CLEAN
Grepped every file in `plugins/languages/elixir/skills/ports/` on the branch for `replac|sandbox|isolat`: all 6 occurrences corrected; remaining "isolation" hits are crash/memory isolation (correct, unrelated). No new wrong occurrence at head.

### Attack 6 — conventions: CLEAN
- elixir `0.1.23` in plugin.json AND marketplace.json; all-skills `0.4.149` in both. Verified at head.
- Frontmatter: name/description/license only, no `allowed-tools`.
- References one level deep; no real hostnames/IPs/op:// paths anywhere in the diff.
- Remote CI was **pending** at review time — Gate 2 must still report green before merge.

Reviewer: gate3-reviewer (fable), did not author the change.

vinnie357:
## Gate 3 addendum — final-state confirmation at head `157d09c`

Spot-confirmed pinned head `157d09c6a88ff585c2fc393b815ddb1af84a80ec` (matches `gh pr view --json headRefOid`; verified via `git show`, not `gh pr diff`). All four findings from the review above are dispositioned:

1. **FIXED** — SKILL.md:73 and :340 now state the unset forms as `false` and the empty **charlist** (`~c""`/`[]`), with the explicit warning that Elixir's `""` is a binary and raises. Matches my probes exactly (`~c""` → unset; `""` → `ArgumentError`).
2. **FIXED** — "The asymmetry is deliberate" → "The asymmetry is in the typespec itself". Grep for `deliberate` across the skill dir at head: zero hits.
3. **FIXED** — reference now notes `denied ++ extra` relies on later-entry-wins for duplicate names, correctly scoped as "observed on OTP 28.4.2, but not documented", with the filter-`denied`-explicitly remedy.
4. **FILED** — Windows case-sensitivity concern tracked as bees claude-skills-277 rather than documented unverified. Correct disposition.

**New OTP 26 paragraph: CONFIRMED accurate.** Independently fetched the published OTP 26 doc (erlang.org/docs/26, `open_port/2` env entry): type union is `Val = os:env_var_value() | false` — no `[]` — and the removal form is described as "the atom `false` (in analogy with os:getenv/1), which removes the environment variable". The extend sentence ("The environment of the started process is extended...") is present verbatim in OTP 26, so the paragraph's claim that extend semantics are not version-dependent while the empty-string unset form is documented only from OTP 27 onward is correct on both halves. The paragraph is properly scoped to what the docs say (behavior on ≤26 untested), and "use `false` rather than an empty charlist" on OTP 26 is the right conservative advice.

Versions remain correctly synced at head (elixir 0.1.23, all-skills 0.4.149 in both manifests — one bump per unreleased version is correct for review-round fixes).

**Verdict stands: APPROVE.** Remote CI on `157d09c` was version-bumps ✓ pass / marketplace-validate pending at time of writing — Gate 2 green required before merge.

PR #264, opened 2026-08-27
feat(core): ship the plain-technical output style

vinnie357:
- Add `plugins/core/output-styles/plain-technical.md` — the first output style in this repo
- `keep-coding-instructions: true` — the style changes how Claude writes, not what work it does; the default `false` would strip coding instructions other core skills presuppose
- `force-for-plugin: false`, written explicitly — `core` is the base plugin, and forcing would silently cancel a user's chosen Concise/Explanatory/Learning style
- Set `outputStyles` on the root manifest so `all-skills` ships the style too, not just `core`
- Document the opt-in path and the plugin requirement in `core:technical-english`

**What was actually verified.** Positive control: `claude -p` with `{"outputStyle":"Explanatory"}` returns Explanatory's system-prompt text, proving `-p` honors the setting. Test: this style file placed at `.claude/output-styles/` and selected by name returns its own sentences verbatim from a live system prompt, including the fork-exception line added during review.

**What was not verified.** Discovery via an *installed* plugin — the `core` convention scan or the root `outputStyles` field — is untested and untestable until 0.2.72/0.4.162 publish. `--plugin-dir` does not wire plugin output styles at all, so there is no headless path to it; only project-level discovery could be exercised. The root `outputStyles` path is confirmed to resolve on disk and to survive `mise update-all-skills`, but that is structural, not behavioral.

Output-style files are an unvalidated surface repo-wide; filed as claude-skills-310.


vinnie357:

Reviewed at `5566a77`; **request-changes**. All findings verified independently by the lead before acting. Fixed at `6f50af7`.

| # | Sev | Finding | Verified | Disposition |
|---|---|---|---|---|
| 1 | major | Style does not apply in a headless probe, and the PR body wrongly claimed headless confirmation was impossible | **confirmed, and the reviewer's positive control was decisive** | **PR body corrected.** `claude -p` *does* honor `outputStyle` — my original claim was wrong. Isolated further: the identical file at `.claude/output-styles/` loads and returns its own text verbatim, so `--plugin-dir` simply does not wire plugin output styles. Not a dead file |
| 2 | major | Issue AC (interactive verification) unmet | confirmed | **superseded** — behavioral verification now done headlessly with a positive control, which is stronger than the AC asked for |
| 3 | major | `all-skills` ships the skill but no output style, so its users follow an instruction to a style they do not have | confirmed (`has("outputStyles")` false, no root `output-styles/`) | **fixed** — `outputStyles` set on the root manifest; verified it survives `mise update-all-skills` (the script upserts, so extra fields persist) |
| 4 | minor | "Does not reach subagents" omits the fork exception | confirmed | **fixed** — and the new sentence verified present in a live system prompt |
| 5 | minor | Restates six skill rules with drift already present (four tenses vs five; procedural/explanatory vs procedure/description), then claims it does not restate them | confirmed | **fixed** — tense list and terms aligned; the false "does not restate" claim scoped to what genuinely is not restated |
| 6 | minor | `output-styles/` is an unvalidated surface — zero hits across `test/`, `mise.toml`, `.github/` | confirmed | **filed as claude-skills-310** with break-it-to-prove-it criteria |
| 7 | nit | Reply-shape bullets overlap Concise's documented behavior | confirmed, no conflict | left as is |

**Clean, verified by the reviewer:** all four frontmatter fields match the live docs table; explicit default-valued `force-for-plugin` is not prohibited; `keep-coding-instructions: true` matches the docs' stated use case; self-application clean including the 16-word description; version bumps and full suite green; disclosure lint clean across 662 files.

**Note the reviewer raised:** `/core:technical-english` was not loadable via the Skill tool — the installed plugin cache predates it — so it read the skill from the repo. Expected for a skill merged minutes earlier, and the right fallback.

Finding 1 is the one that mattered: it caught me asserting a verification limit I had not actually tested, and its positive control disproved it.

vinnie357:
## Gate 3 re-review — approve at `6f50af7`

All seven original findings verified fixed with independent probes by the reviewer, including a live `claude -p` run quoting the new fork-exception sentence back from the system prompt.

Three new items raised on re-review, all non-blocking; two fixed at `c1489ef`:

| # | Sev | Item | Disposition |
|---|---|---|---|
| 1 | minor | PR body's "working correctly on the install path" overstated the evidence — project-level discovery was tested, plugin-install discovery was not | **fixed** — PR body now separates what was verified from what was not, and says plainly that install-path discovery is untestable until 0.2.72/0.4.162 publish |
| 2 | nit | `jq` dropped the trailing newline from the root manifest (last byte `}`, while sibling manifests end in `0a`) | **fixed** — newline restored, JSON still valid, suite green |
| 3 | nit, uncertain | With both `core` and `all-skills` installed, the same style arrives from two plugins; picker dedup for same-named styles across plugins is undocumented | **noted, not changed** — settled only by installing both; recorded here so it is not rediscovered |

**Reviewer's side-effect check, as requested:** `plugins/core/.claude-plugin/plugin.json` is untouched and still has no `outputStyles` key, so `core`'s convention scan is unaffected — the root field governs `all-skills` only. `update-all-skills --dry-run` retains the field, confirming the upsert claim.

Finding 1 is the second time this PR that a verification claim of mine outran the evidence. The first was corrected in the previous round; this one narrowed "install path" to what was actually exercised. Both were caught by the gate, not by me.

PR #263, opened 2026-08-27
feat(core): add technical-english skill and reconcile the hedging contradiction

vinnie357:
- Add `core:technical-english` — a two-register writing standard: strict mode for procedural text, flavored mode for explanatory prose
- Resolve the live contradiction between `anti-fabrication:23` (required hedges) and `claude-skills:70` (banned them) with one rule: hedge about the world, never about the instruction
- Add word budgets to the worker-report and review-findings contracts — "one sentence" alone permitted a 69-word sentence; ci-evidence has no prose field and needed none
- Add three sources entries (ASD-STE100 as shape only, Zinsser, OpenSTE) with an explicit no-conformance-claim boundary
- Skill passes its own sentence-length rule and scores 20/20 on `test:skills-quality` at 98 lines
- Not added to the mandatory core list; no prose gate ships here — Vale adoption is claude-skills-306

PR #265, opened 2026-08-27
fix(validate-skills-quality): flag bare sibling links in ref_depth

vinnie357:
Closes claude-skills-309.

- `ref_depth` short-circuited on `str contains "references/"`, so a bare sibling link — `[attribution.md](attribution.md)` — was never examined despite violating the same one-level rule
- Add a second detection branch: a markdown link whose target exactly matches a sibling reference basename, with anchor and title forms handled
- Thread a `sibling_basenames` parameter through the predicate; the real call site computes it from the other files in the same `references/` directory
- Document the accepted residuals — a target with a directory component (`[x](./attribution.md)`), an angle-bracket target, and nested brackets still escape, matching the function's existing minimalism

**Found a real pre-existing instance**, not just the synthetic case: `plugins/wasm/skills/wit/references/patterns.md` linked `[syntax-reference.md](syntax-reference.md)`. Fixed rather than baselined, per `quality-baseline.json`'s "Do not add entries for new code; fix the skill instead."

On the score count: `main` and this branch both sit at 88/107 perfect. The widened check alone drops it to 87 by exposing the `wit` violation; the corpus fix restores 88. The delta is a caught bug, not a score change.

**Adversarial TDD.** Tests were authored and frozen by a different agent than the implementer. `git diff` on the test file across the implementation commits touches only the predicate body and comment block; no assertion changed.

**Proven by breaking it**, twice independently: appending a bare sibling link to a real reference file makes `test:skills-quality` exit 1 naming `core/technical-english:ref_depth` and the offending file. Reverted; tree clean.

This is the bug that let PR #263 ship a reference-to-reference link while flagging the reverse direction.


vinnie357:

Round 1 at `c7d2780`: approve with 2 minor, 2 nit. Round 2 at `13190b5`: **approve**, all verified fixed by execution.

| # | Sev | Finding | Disposition |
|---|---|---|---|
| 1 | minor | Anchor and title forms escaped the new branch, while the old `references/` branch already flagged `references/foo.md#anchor` — the two halves disagreed on the most realistic evasion | **fixed via the pair** — new cases frozen at `b1dcca6` by the test author, stripping implemented at `13190b5`. Reviewer's 13 probes confirm: combined `[x](attribution.md#a "T")` flags, and `[x](attribution.md "T#not-a-fragment")` flags — title-then-fragment order is correct, since fragment-first would strand `attribution.md "T` and miss it |
| 2 | minor | `scheme://` rejection was provably dead under exact matching | **deleted** — implementer isolated the logic in a standalone script and ran the frozen inputs before editing; reviewer re-confirmed it stays inert under strip-then-match |
| 3 | nit | Root manifest ended `}\n\n` | **fixed** — my bug, a `printf '\n'` after `jq` had already added one |
| 4 | nit | "87 → 88" read as a main→HEAD delta; main is also 88 | **fixed** — PR body now states the widened check alone drops it to 87 by exposing the `wit` violation, and the corpus fix restores it |

Round-2 nit: title stripping handles only the double-quote form, so `[x](attribution.md 'Title')` escapes. Reviewer marked it no-action, but an incomplete residual list is the same defect class this check exists to catch — **documented at `cb636e2`**.

**Verified by execution, not reasoning:** all three documented residuals reproduce (directory-prefix, angle-bracket, nested-bracket targets each escape); both positive cases fail red against the pre-fix implementation for the right reason; the two negative cases pass either way and are structural guards against a future basename-extraction widening, not harmful passengers.

**Adversarial boundary held across both rounds** — `git diff` on the test file touches only the predicate body and comment; zero assertion lines changed.

**Break-it proof reproduced independently by me** on the anchor form specifically, since that was the new behavior: appending `[attribution.md](attribution.md#sources)` to a real reference file exits 1 naming `core/technical-english:ref_depth`. Reverted clean.

PR #266, opened 2026-08-27
feat(test): adopt Vale as the prose gate

vinnie357:
Closes claude-skills-306.

- Pin `vale = "3.18.0"` in `mise.toml` — pinned, not `latest`, since rule semantics shift between minors and CI would absorb it silently
- Add `.vale.ini` selecting a register by path: **Strict** for `agents/` and `commands/`, **Flavored** everywhere else
- Six rules across `styles/Strict/` and `styles/Flavored/`
- Add `mise run test:prose`, wired into `mise run test`
- Update `core:technical-english`, whose "no prose gate runs in CI today" line is now false

**Findings on the current corpus: 169** — 69 sentence-length, 69 qualifiers, 30 superlatives, 1 hedged directive, and **0 clutter**.

**Clutter ships at `error` and gates the build today.** The rest report at `warning` while they burn down (claude-skills-307). Proven both directions: a real clutter phrase added to `README.md` makes `mise run test:prose` exit 1; clean, it exits 0.

**False positives handled where our own docs teach the rules.** The banned-superlative lists in `CLAUDE.md` and `core:anti-fabrication`, and the clutter examples in `technical-english/references/flavored-prose.md`, are the rule definitions — suppressed inline rather than counted. Without that, Clutter could never reach zero and the gate would rest entirely on false positives.

**`Qualifiers` was rebuilt after measurement.** As first written it produced 227 findings, 156 of them `rather` — almost all "rather than", a legitimate construction. A rule whose most frequent match is a false positive teaches authors to ignore the gate. Dropping that one token leaves 69 real findings.

`AGENTS.md` is a symlink to `CLAUDE.md`; the task filters symlinks so the same bytes are not scanned twice.

No workflow change needed, and this is verified rather than assumed: the CI log shows `vale@3.18.0` installing on Linux and `test:prose` reporting real findings during the run.

Vale replaces the hand-rolled checks originally planned. Its `consistency` and POS-tagged `sequence` extension points do term-consistency and passive-voice detection that a regex cannot do soundly.


vinnie357:
## Gate 3 adversarial review (vale-reviewer, fable)

## SCOPE
Vale prose gate: six rules, two registers, `test:prose` task, symlink filter, CLAUDE.md suppression, skill update, CI availability, count and proof claims.

## FINDINGS
1. [major] PR body — the headline count does not reproduce at the PR HEAD (574810f). Measured: **112** findings (69 sentence-length, **39** superlatives, 3 clutter, 1 hedged), not 123/50. The 11-superlative delta is exactly the PR's own two mitigations: un-suppressing CLAUDE.md:314 adds 4, and scanning the AGENTS.md symlink double-counts CLAUDE.md's unsuppressed 7 (112 − 3 + 7 + 7 = 123). The number was measured before the mitigations it ships with.
   Fix: restate the corpus count as 112 (69/39/3/1).
2. [major] PR body — the error-path proof count is wrong: promoting `Flavored.Clutter` to `error` exits 1 on **3** findings (`complete` capture: exit=1, lines=3), not 6; the task's two `vale` invocations print the same 3 findings twice.
   Fix: state 3 findings, or note the double-print.
3. [major] styles/Flavored/Qualifiers.yml:4 — `level: suggestion` under `MinAlertLevel = warning` means the rule can never report: a fixture with "very/quite/really" produced zero output through the gate and fires only with `--minAlertLevel suggestion`. The updated skill's "It checks … weak qualifiers" (SKILL.md:92) and the `.vale.ini` comment "Every rule ships at `warning`" are both false as shipped.
   Fix: raise Qualifiers to `warning`, or delete it and drop "weak qualifiers" from the skill sentence and the .ini comment.
4. [minor] .vale.ini:24,33 — `*` crosses `/` in Vale section globs (fixture-verified: `plugins/*/agents/*.md` alone matches `plugins/tools/qa/agents/` and `plugins/a/b/c/agents/`), so both `plugins/*/*/…` sections are dead config. Coverage is complete either way — depth-4 dirs like `plugins/tools/github/skills/pr-review/agents/` also get Strict.
   Fix: drop the two doubled sections.
5. [minor] styles/Strict/SentenceLength.yml — fires on headings, list items, and table rows (fixture-verified), not only sentences; 27 of the 69 corpus findings are list items. It also applies one 25-word cap where the skill's strict rule 6 caps procedure sentences at 20. Boundary itself is exact (25 passes, 26 fires); code spans/blocks are skipped.
   Fix: accept and document, or scope the burn-down to prose lines.
6. [minor] The rule-definition false-positive class is suppressed only in CLAUDE.md. `plugins/core/skills/anti-fabrication/SKILL.md` contributes 9 superlative findings quoting the banned list, and all 3 clutter findings sit in `technical-english/references/flavored-prose.md:23` — deliberate teaching examples, so the error-promotion proof gates entirely on false positives, and Clutter can never burn to zero without suppressing them. bees 307 already mandates exemptions the shipped rule lacks.
   Fix: extend the inline suppression to the two teaching files (or add the 307 exemptions now).
7. [nit] mise.toml test:prose comment — "git ls-files lists CLAUDE.md, a symlink to AGENTS.md" inverts the direction; `AGENTS.md -> CLAUDE.md`.
   Fix: swap the two names.

Clean categories, each tool-verified: symlink predicate works (483→482 files; AGENTS.md excluded, CLAUDE.md scanned); suppression scope exact (line 314 silenced, findings at 579/602/677 after `= YES` still fire); Nominalization `$1` capture produces correct verbs ("evaluate", "validate"); Clutter's empty replacement renders without breakage; overlapping Strict sections do not duplicate findings; CI is sound (`mise install` runs unconditionally after cache restore, cache key hashes mise.toml, `.vale.ini` pulls no packages so no `vale sync`; vale resolves via `aqua:vale-cli/vale`); `mise run test` exits 0 with test:prose executing; `mise run test:version-bumps origin/main` green, versions synced (core 0.2.72→0.2.73, all-skills 0.4.163→0.4.164 in both manifests); disclosure lint green inside `mise test`.

## VERDICT
request-changes — the gate mechanism works as designed, but the PR's headline count, proof count, and the shipped skill's qualifier claim all fail reproduction and must be corrected before merge.

vinnie357:
## Gate 3 re-review at 9329d43 (vale-reviewer, fable)

## SCOPE
Re-verify all seven findings at 9329d43, the rebuilt Qualifiers rule, the new gate exit path, suppression scope, and the 169-count claim.

## FIX VERIFICATION
1. **verified-fixed** — corpus at 9329d43 measures exactly 169 (69 SentenceLength, 69 Qualifiers, 30 Superlatives, 1 HedgedDirective, 0 Clutter); PR body split matches finding-for-finding.
2. **verified-fixed** — the double-print is gone; the second pass runs `--output=JSON` through `complete` and speaks only through its exit code.
3. **verified-fixed** — Qualifiers at `warning` now reports (69 findings through the gate); `rather` absent from the token list; `.vale.ini` and skill text no longer claim "every rule ships at warning". Note for burn-down: `actually` is 51 of the 69 (74%), but sampled instances are genuine hedges per the skill's own teaching, unlike `rather than` — no action demanded.
4. **verified-fixed** — the depth-3 sections are gone; the replacement comment ("`*` crosses `/`… a `plugins/*/*/agents/*.md` twin would be dead config") matches my fixture evidence.
5. **partially disputed — both measurements reproduce, and they don't conflict.** My heading/table claim REPRODUCES as capability: under the shipped `.vale.ini` + `styles/` at 9329d43, a file containing only a 26-word heading fires (`onlyheading.md:1:3:Strict.SentenceLength`) and a 26-word table cell fires (`onlytable.md:1:3`). Your corpus classification also reproduces: I re-checked all 69 corpus findings and zero land on `#`- or `|`-prefixed lines. The corpus has no over-25-word headings or table rows in Strict files today; the rule fires on them the day one appears. What differs is fixture-vs-corpus, not vale behavior. List items firing deliberately: accepted, documented. But see NEW FINDINGS — the skill sentence shipped to record this codifies the wrong side.
6. **verified-fixed** — Clutter genuinely at zero: the three suppressed regions contain only rule-definition/teaching lines (anti-fabrication SKILL.md:44–54 is the banned-word lists; flavored-prose.md:24 quotes the clutter phrases as examples; corpus-wide there is exactly one clutter-phrase-bearing line outside a suppression and it's in a fenced code block, which Vale skips). No `= NO`/`= YES` pair leaks: findings before and after each region still fire (anti-fab:38, flavored-prose:27, CLAUDE.md:579/602/677). Clutter at `error` proven both directions with the VERBATIM task script extracted from mise.toml, run in a scratch git repo: tracked violation file → prints the finding + "Prose gate failed", exit 1; violation removed → exit 0. The `| ignore` regression is dead.
7. **verified-fixed** — mise.toml comment now reads "AGENTS.md is a symlink to CLAUDE.md", matching `ls -la`.

Mechanical: `mise run test` exit 0 (test:prose runs, gate silent); `mise run test:version-bumps origin/main` green (core 0.2.72→0.2.73, all-skills 0.4.163→0.4.164); disclosure lint green inside `mise test`.

## NEW FINDINGS
1. [major] plugins/core/skills/technical-english/SKILL.md:94 — "Sentence length is measured per sentence and **skips headings** and code" is false: an isolated 26-word heading fires `Strict.SentenceLength` under the shipped config (fixture above), and so does a table cell, which the sentence omits. "Skips code" is true; the list-item clause is true. This codifies the disputed half of F5 as mechanism fact when only the corpus half was measured.
   Fix: state it as it is — e.g. "skips code; headings, table rows, and list items all count" — or scope the claim to the corpus.

## VERDICT
request-changes — every prior finding is genuinely fixed and the gate provably blocks, but the one sentence added to settle F5 asserts behavior the fixture disproves; correct it and this approves.

vinnie357:
## Gate 3 final review at a84bdbd (vale-reviewer, fable)

## FIX VERIFICATION
**verified-fixed** — SKILL.md:92 now reads "Sentence length skips code. Headings, table rows, and list items all count…", which matches the fixture evidence exactly (code span and fenced block do not fire; heading, table cell, and list item each fire in isolation). Corpus count unchanged at 169; `mise run test:prose` exit 0 at a84bdbd.

## SWEEP
Swept every changed file for corpus-derived behavior stated as mechanism: **none found.** Each remaining claim is one of two kinds, both sound:
- Corpus claims explicitly framed as corpus state: Qualifiers.yml ("156 of 227 corpus hits were `rather`"), Clutter.yml ("At zero findings on this corpus"), `.vale.ini` ("Clutter is already at zero and gates today") — all reproduce.
- Mechanism claims, each now tool-verified: `*` crosses `/` in section globs (fixture); AGENTS.md → CLAUDE.md symlink direction (`ls -la`); "new clutter fails the build" (verbatim-script scratch-repo proof, exit 1/0 both directions); "skips code" (fixture); and the mise.toml comment that `| ignore` would discard the exit code and silently disable the gate — confirmed directly: `nu -c '^false | ignore; print "reached"'` prints and exits 0.

## VERDICT
**approve** — all eight findings across three rounds are fixed and re-verified, the gate provably blocks and passes, and no unverified mechanism claim remains in the changed files.

vinnie357:
## Gate 3 — adversarial review (fable, did not author), three rounds

**Approve at `a84bdbd`.** Eight findings across three rounds, all fixed and re-verified by execution.

### Round 1 at `574810f` — three majors, all reproduction failures on my claims

| Claim | Reality |
| :--- | :--- |
| 123 findings | **112** — measured before applying my own suppression and symlink filter |
| Error proof on 6 findings | **3** — the task ran Vale twice, so I counted duplicate prints |
| "checks weak qualifiers" | **Never fired** — `suggestion` under `MinAlertLevel = warning` is invisible |

Plus: dead depth-3 globs (`*` crosses `/`), an inverted symlink comment, and the sharpest one — **Clutter's only 3 findings were teaching examples in our own skill**, so the error-path proof gated entirely on false positives.

### A regression I introduced while fixing round 1

Suppressing the double-print with `| ignore` discards the exit code along with the output. A real violation returned **exit 0** — the gate was live, reporting, and completely toothless. Caught only by re-running the break-it test after what looked like a cosmetic edit. Now captures status via `complete` and re-raises.

### Round 2 — a false mechanism claim, and I was wrong

I disputed the reviewer's heading/table finding using a corpus scan showing zero such findings. Both measurements were true and never conflicted: its fixture proved the rule **can** fire on headings and table cells; my scan proved no such heading exists in Strict files **today**. Capability versus corpus.

I then wrote that corpus observation into the skill as mechanism — "skips headings and code" — which is false. Verified before conceding: an over-cap heading fires at `h.md:1:3`, a table cell at `t.md:3:3`. Corrected.

### What changed as a result

- `Qualifiers` rebuilt after measurement: 227 findings, **156 of them `rather`** (almost all "rather than"). Dropping that token leaves 69 real ones. A rule whose most frequent match is a false positive teaches authors to ignore the gate.
- Teaching-example suppression dropped **Clutter to zero**, so it ships at `error` and gates the build — rather than everything being advisory with an unproven error path.

### Verification worth noting

The reviewer proved the gate by extracting the task script **verbatim** from `mise.toml` into a scratch git repo, rather than trusting an in-place test — the right paranoia for a check whose failure mode is silently passing. It also confirmed the `| ignore` comment describes real nushell behaviour by running `nu -c '^false | ignore; print "reached"'`.

Final sweep for other corpus-inferred mechanism claims: **none found**.

PR #267, opened 2026-08-27
feat(vale): add the vale plugin skill

vinnie357:
Closes claude-skills-311.

- New `plugins/tools/vale/` plugin, matching the one-plugin-per-tool pattern used by dagu, ansible, proxmox, and playwright
- `SKILL.md` (109 lines) plus three references: extension points, rule authoring, and how this repo is wired
- `sources.toml` + `sources.md` for the new plugin
- Cross-linked from `core:documentation` and `core:technical-english` — those point at the tool skill; `technical-english` keeps ownership of the register rules

**The install example is verified to run**, not just written. All three commands executed verbatim in a clean scratch project containing only the pinned `mise.toml`:

```
mise install                   → mise all tools are installed
mise exec -- vale --version    → vale version 3.18.0
mise registry | grep '^vale'   → vale  aqua:vale-cli/vale  asdf:pdemagny/asdf-vale
```

A genuine cold install — no cache, fresh runner — is evidenced in PR #266's CI log: download, checksum, extract, installed. Locally that path is masked by mise's download cache, so the CI run is the honest proof.

**Behaviour is fixture-verified, and what was not tested says so.** Seven of the eleven extension points fired against purpose-built fixtures. The other four are marked **Documented, untested** with their `docs.vale.sh` source, rather than asserted. Also verified by fixture: `*` crosses `/` in section globs, suppression comments scope correctly, fenced code is skipped, `text.frontmatter.<field>` targets exactly that field, only `error` exits non-zero, and `sequence` does POS-tag matching.

**The `ref_depth` check from #265 caught a violation in this very skill** — `extension-points.md` referenced a sibling reference. Fixed; skill now scores 20/20.

Carries the three measured lessons from #266: a rule is only as good as its false-positive rate (227 findings, 156 of them `rather`); a document teaching a rule will trip it, and suppression is correct there; and `| ignore` in nushell discards the exit code and silently disables a gate.


vinnie357:
## Gate 3 — adversarial review (fable, did not author), three rounds

**Approve at `49c7cac`.** Five findings, all verified fixed, plus a clean final sweep.

### Round 1 — one major, and it was the dangerous one

The skill claimed Vale sections "combine rather than replace." **They do not.** Reproduced independently: two style sets in separate sections, and a file matching both got only the last section's rules — flipping the section order flipped the winner. Not most-specific-wins.

The failure mode is silent: anyone adding `[plugins/*/hooks/*.md] BasedOnStyles = Strict` on the strength of that sentence would lose every Flavored rule on those files and see nothing wrong. This repo's config is correct only because both strict sections spell out `Flavored, Strict` — and the skill documented the wrong reason it works.

That is the second corpus-inferred mechanism claim this week, after "skips headings" in #266. Same shape: a true observation about this repo generalized into a false statement about the tool.

Also round 1: three YAML snippets labelled "This repo's `styles/...`" were abridged (3 of 6 tokens, 2 of 12 swaps, a missing `ignorecase`), and one entry's "Verified" rested on clean CI runs — which cannot verify anything, since a typo'd swap key also produces clean runs.

### Round 2 — my fix introduced two fresh errors

| Finding | Detail |
| :--- | :--- |
| Count error | I wrote "eleven swaps" for a file carrying **twelve** — a count error inside a fix for count errors |
| Over-broad rule | "Sections do not merge" is wrong for **identical** globs: two `[sub/*.md]` sections accumulate, both style sets fire. Distinct globs override; identical globs merge |

The identical-glob case was the exact edge case I asked the reviewer to check, and I was wrong about it.

### What the review added beyond my own fixtures

Its three-section test (`[*.md]`→A, `[sub/*]`→B, `[sub/*.md]`→C, only C fired) confirmed last-match-wins generalizes past two sections, and its order-flip put the *less* specific section last and watched it win — settling "not most-specific" in a way my two-section fixture could not.

It also built the cold-install control I could not: a scratch project with its own `MISE_DATA_DIR` and `CACHE_DIR` and an empty installs dir, proving a genuine download of vale 3.18.0 from nothing. My local attempts kept hitting mise's download cache.

And during the final sweep it caught its **own** probe error — it had tested `[^\s.!?]+` rather than the skill's `[^s.!?]+`, a different regex — then re-ran and confirmed the skill's claim holds.

PR #269, opened 2026-08-27
test(validate-skills-quality): validate output-style frontmatter and outputStyles paths

vinnie357:
Closes claude-skills-310.

`output-styles/` was an entirely unvalidated plugin surface. Pass 2 covered `agents/`, `commands/`, and `hooks/hooks.json`; a malformed style file shipped silently. I know that because I shipped this repo's first output style in #264 and nothing objected.

Four rules, all wired into the Pass-2 corpus loop:

| Check | Fires when |
| :--- | :--- |
| `fm_schema` | frontmatter carries a key outside the four documented ones (`name`, `description`, `keep-coding-instructions`, `force-for-plugin`) |
| `missing_name` | no `name:` |
| `missing_desc` | no `description:` |
| `missing_path` | a `plugin.json` sets `outputStyles` to a path that does not exist |

**Adversarial TDD.** Tests frozen at `3f31a96` by a different agent than the implementer. Both boundary commands are clean — `git diff 3f31a96 -- <file>` shows only implementation, and `git status --porcelain -- <file>` is empty. Both are required: `A..HEAD` compares commits and cannot see an uncommitted edit to a frozen test.

**Proven by breaking each rule**, then reverting. Reproduced independently: adding an undocumented key to the real `plain-technical.md` fails the run — and it fails via the regression guard that reads the *shipped* file, which proves that guard is not a passenger.

### A coverage gap this work exposed — filed as claude-skills-312

`check-output-styles-path` cannot see the only manifest in this repo that sets `outputStyles`. The root `.claude-plugin/plugin.json` is skipped at `validate-skills-quality.nu:3386` when building the registry, because `all-skills` re-lists every other plugin's skills and would double-count them. The exclusion is right for aggregation; the side effect is that the meta-plugin's own manifest is exempt from every plugin-level check — including the one added here, and including the field #264 set on it.

Breaking `plugins/core/.claude-plugin/plugin.json` produces a finding. Breaking the root one produces nothing. Filed rather than fixed here, since covering it means special-casing the meta-plugin and belongs behind its own frozen tests.


vinnie357:

**Approve at `1cc9a04`**, with one minor dispositioned at `ba29ec4`.

### Round 1 — a crash, not a missed check

`check-output-styles-path` **threw** `Can't convert to string` on `outputStyles: ["./output-styles/"]`. The plugins reference documents the field as **string or array**, so this was a crash on valid input — and because it throws rather than returning a finding, it would abort the entire validator run, not merely skip a rule.

Fixed through the pair rather than filed: four cases frozen at `b65aa3e`, normalization via `[$value] | flatten` at `1cc9a04`. The test author wrapped each new call in `try/catch` so a crash surfaces as one failed assertion instead of killing the suite before later ones print.

Proven three ways — direct call on all three array shapes, the frozen cases, and on the **real corpus** with `plugins/core`'s manifest set to a mixed `["./output-styles/", "./missing/"]`, which reports `missing_path` instead of throwing. The self-test calls the function directly and cannot distinguish "handles arrays" from "handles the one shape the test passes."

### Round 2 — a fabricated tracker citation

Two shipped comments cited **`claude-skills-313`**, which does not exist — `bees show` returns not-found and it has zero matches in the export. An invented issue ID in shipped text is precisely the fabricated-citation class this repo lints prose for, sitting inside the validator that does the linting. Both now cite `claude-skills-310`.

### What the review verified that I could not assume

- **The try/catch wrapper cannot mask a broken implementation.** It ran the frozen tests against the *pre-fix* body: all four array cases fail, including the empty-array case (`path join []` crashed too), so none is a passenger. Every crash path lands in `catch { null }` and each case checks `== null` before the short-circuiting `or`.
- **Undocumented shapes still throw** (nested array, record, int, mixed-type array) — judged consistent with the rest of the validator, since a record-valued `skills` would break the registry loop the same way. Not a finding.
- **Requiring `name`/`description` is defensible.** Both are optional upstream; the house rule is scoped to this repo's own styles and documented in the constant's comment.

### On claude-skills-312

Priority stays P3 — the residual is a silent missed check on one rarely-edited manifest, not an abort. Recorded on the issue: `mise update-all-skills` regenerates the root manifest, so if that tooling ever emits `outputStyles`, its output lands on the unvalidated path.

PR #270, opened 2026-08-27
test(validate-skills-quality): check the root manifest outputStyles path

vinnie357:
Closes claude-skills-312.

The root `.claude-plugin/plugin.json` was exempt from every plugin-level check, because `validate-skills-quality.nu` skips `all-skills` when building the plugin registry. That exclusion is correct — the meta-plugin re-lists every other plugin's skills and would double-count the whole corpus — but it left the one manifest in this repo that actually sets `outputStyles` unvalidated.

Found while shipping #269: breaking `plugins/core`'s manifest produced a finding, breaking the root one produced nothing. Confirmed against `origin/main` during review — the same break exits 0 there with zero root-manifest mentions.

- `check-root-manifest` delegates to the existing `check-output-styles-path`. **No new checking logic** — that function already handles the string and array forms after #269
- One call site in `main()`, outside the registry loop, keyed `root-manifest`
- The `all-skills` registry exclusion is untouched; skill count unchanged at 108

**Adversarial TDD.** Tests frozen by a different agent than the implementer; both boundary commands clean.

**Proven by breaking it:** the real root manifest with a bad `outputStyles` now yields `root-manifest/.claude-plugin/plugin.json:missing_path`.

### The fourth test case shipped inert, and was fixed

Gate 3 found that case 4 — the guard stopping someone from "fixing" this gap by deleting the exclusion — **could never fail**. It searched the validator's own source for the exclusion literal, but that string appears twice: once as the real exclusion, once inside the guard's own text. `str contains` matched itself.

Verified independently by deleting the real exclusion: the suite still printed `✅ Root-manifest self-test passed (4 cases)` and exited 0.

Now counts occurrences and requires at least two — the guard's own self-reference plus the real line. Re-verified by the same deletion test, which now fails with `found 1 occurrence(s), need >= 2`.

I had called this guard "brittle to rewording" when it was first reported. That was too generous: it was not brittle, it was inert, and its comment claimed otherwise.

### Follow-up filed

**claude-skills-313** — a malformed manifest crashes the validator rather than reporting a finding, aborting the run and masking every other result. Pre-existing across all 31 plugin manifests, not introduced here; `mise test:plugins` fails malformed JSON separately, so CI is red either way.


vinnie357:

**Approve at `6a59c8e`**, with the residual nit dispositioned at `0862fc7`.

### The finding: a guard that could never fail

Case 4 exists to stop someone "fixing" this gap by deleting the `all-skills` exclusion — which would double-count every skill in the corpus. It searched the validator's own source for the exclusion literal.

That literal appears **twice**: once as the real exclusion, once inside the guard's own text. `str contains` matched itself, so the guard always passed. Verified by deleting the real exclusion — the suite still printed `✅ Root-manifest self-test passed (4 cases)` and exited 0.

A guard documented as load-bearing, structurally incapable of firing, shipped inside the gate whose purpose is catching exactly that.

Fixed by counting occurrences and requiring at least two. Re-verified by the same deletion: `found 1 occurrence(s), need >= 2`.

### The reviewer then defeated the fix — in one edit

I framed a bypass as needing a coordinated rewrite of both the needle and the real line. It is cheaper than that. Replacing the exclusion with `# removed: if ($plugin.name == "all-skills") { continue }` keeps the literal, keeps the count at 2, and passes. Reproduced independently.

That is a plausible careless edit, not only a deliberate one. Not blocking, because the reviewer proved the real backstop holds: with the exclusion gone, a full validator run reports **216 skills instead of 108** and fails on duplicated `anti_fab` findings. The corpus scan is load-bearing; this check is an early tripwire.

But the comment claimed the guard "only passes when the real exclusion is also still present," which is false. Given this PR exists because a guard's comment asserted behaviour the code did not have, shipping a second such comment was not an option. It now states the residual plainly and names the corpus run as the actual backstop.

### Also verified by the review

- **Rewording trips it**: a semantically identical `if "all-skills" == $plugin.name` fails the guard. A loud false-fail on harmless refactors, with a message naming what to update.
- **Cases 1-3 byte-identical**; the amendment touches only case 4's block and comment.
- **The gap was real**: the same break exits 0 on `origin/main` with zero root-manifest mentions.

### Follow-up

**claude-skills-313** — a malformed manifest crashes the validator rather than reporting, aborting the run and masking every other finding. Scoped to both unguarded `open` sites, not just the one added here. P3: `mise test:plugins` fails malformed JSON separately, so CI is red either way.

PR #271, opened 2026-08-27
fix(validate-skills-quality): report malformed manifests instead of crashing

vinnie357:
Closes claude-skills-313.

A malformed `plugin.json` aborted the entire validator run. `Total skills` never printed and every other finding in that run was masked — one bad brace hid the whole corpus scan.

Same class as the `outputStyles` array crash fixed in #269: **a crash aborts, a finding reports.**

- New `read-plugin-json` helper returning `{status, data, file}` with `status` ∈ `ok` | `invalid` | `missing`. Absent and malformed are deliberately distinct states — the existing code already skipped absent manifests, and collapsing the two would change that behaviour
- All four bare `open` sites now route through it: `check-root-manifest`, the Pass-1 registry loop, the Pass-2 skills loop, and the Pass-2 outputStyles check
- A malformed manifest emits an `invalid_json` finding naming the file, then that manifest is skipped

**Proven at two sites**, which fail at different phases so one proof would not cover the other:

| Broken manifest | Result |
| :--- | :--- |
| `plugins/core/.claude-plugin/plugin.json` | `core/.claude-plugin/plugin.json:invalid_json`, run completes, `Total skills: 94` |
| root `.claude-plugin/plugin.json` | `root-manifest/.claude-plugin/plugin.json:invalid_json`, run completes, `Total skills: 108` |

The 94 is correct, not a regression: core's 14 skills cannot register from a manifest that will not parse. The point is that the run **finishes and reports** rather than dying.

**Adversarial TDD.** Tests were authored and frozen at `a2e0c1e` before implementation began, and the implementer was a separate agent invocation with no shared context. Note that git cannot evidence that separation — every agent commit carries the same identity — so the verifiable part is the boundary: `git diff a2e0c1e -- <file>` shows only implementation, and `git status --porcelain -- <file>` was empty at commit time. The commit sequence is consistent with the claim; the separation itself rests on the process, not on the repository.

### On the issue's original scope

I filed claude-skills-313 saying "two unguarded call sites", taking the count from a review report without checking. There are **four**. Corrected on the issue before work started, with a note that the original count was unverified, and the AC now tells the implementer to confirm the count themselves rather than trust the list. The test author independently verified four and confirmed no fifth — the other nearby `open` calls are already try-guarded or are not manifests.

### Two sites are unreachable today, and are guarded anyway

The Pass-2 loops iterate `$registry`, which Pass 1 has already filtered to parse-clean manifests, so invalid JSON cannot reach them in practice. Both were still routed through the helper rather than left as bare `open` calls, and they skip silently instead of re-emitting a duplicate of the finding Pass 1 already filed. The reasoning is documented inline at both sites.


vinnie357:

**Approve at `2ffaf13`.** Two findings, both verified **pre-existing siblings** rather than defects introduced here; filed as follow-ups.

### The change delivers its scope

- Broken `plugins/core` manifest → `core/.claude-plugin/plugin.json:invalid_json`, run completes, `Total skills: 94`
- Broken root manifest → `root-manifest/.claude-plugin/plugin.json:invalid_json`, run completes, `Total skills: 108`

The 94 is jq-verified as 108 − core's 14 skills. Exit 1 comes from the injected finding via the baseline hard-failure path, not a lingering crash; a clean run still exits 0. Also confirmed: `--update-baseline` is shrink-only and refuses `invalid_json` keys, so the finding can never be baselined away.

### The structural change was clean

The implementer moved the `mut failing_keys` / `failing_counts` / `surface_results` declarations earlier so Pass 1 could append — the riskiest edit in the diff and not in the brief. Verified: **clean-tree full-run output is byte-identical** between `origin/main` and this branch, both exit 0. Every write after the moved declarations is an append or accumulator merge, and the ratchet is membership-based, so the new Pass-1-first ordering cannot change verdicts.

### Findings — both pre-existing, both filed

**claude-skills-314 [major]** — a manifest that is *valid JSON but not a record* still aborts. Reproduced independently: `[1,2,3]`, `"juststring"`, `null`, and an empty file all die before `Total skills` prints. A UTF-8 BOM does the same by a different route — nushell `open` returns a *string* rather than raising, so a try/catch never fires and the BOM case needs its own test. Confirmed pre-existing: BOM and `null` crash identically on `origin/main`.

**claude-skills-315 [minor]** — `marketplace.json` is still opened bare at line 3654. Lower severity: with no parseable marketplace there is no corpus at all, and the error does name the file.

### A claim in my PR body that could not be verified

The reviewer noted that "tests frozen by a different agent than the implementer" is **unverifiable from git** — every agent commit carries the same identity. The commit sequence is consistent with it, but the repository cannot evidence it.

Corrected in the body: the separation rests on the process, and the *verifiable* part is the boundary — `git diff a2e0c1e -- <file>` shows only implementation, `git status --porcelain` empty at commit time. Worth recording because I have asserted that separation on every adversarial-TDD PR this session, and it is a process claim wearing the clothes of a checkable one.

PR #272, opened 2026-08-27
fix(validate-skills-quality): reject manifests that parse to a non-record

vinnie357:
Closes claude-skills-314.

PR #271 made a manifest that fails to **parse** report `invalid_json` instead of aborting. A manifest that parses to something other than a **record** still aborted, at the first field access in the Pass-1 registry loop.

One post-parse type check in `read-plugin-json`: if the parsed value is not a record, return `invalid`.

**Proven on the real corpus for all five shapes.** Each written to `plugins/core/.claude-plugin/plugin.json`, each reporting `core/.claude-plugin/plugin.json:invalid_json` and **still printing `Total skills: 94`**:

`[1,2,3]` · `"juststring"` · `null` · empty file · BOM + valid JSON

The 94 is correct, not a regression — core's 14 skills cannot register from a manifest that is not a record. The point is that the run **finishes and reports**.

### Two measured facts that shaped the fix

**The BOM case never throws.** `open` on a BOM-prefixed `.json` returns a **string** and raises nothing — the BOM breaks `from json`'s auto-detect and nu falls back to raw text. A wider `try/catch` cannot reach it. That case alone proves the fix had to be a post-parse type check rather than a broader catch.

**`describe` on a record returns the full field signature**, e.g. `record<name: string, version: string>` — not the bare word `"record"`. A check written as `== "record"` would be false for **every** well-formed manifest, turning the five cases green while rejecting the entire corpus. The check uses `str starts-with` and says why inline.

Both were measured by the test author and confirmed independently before implementation. The loudest guard against getting the second one wrong is the skill count: it stays **108**, where a bad check collapses it.

### Verification notes

The BOM fixture needs `save -f --raw`; without `--raw`, nushell pretty-prints the bytes as the text `"[239, 187, 191]"` instead of writing them. Confirmed with `xxd` that the first three bytes are `ef bb bf` followed immediately by `7b`.

Exit codes were checked by redirecting to a file rather than piping through `tail` — a pipe returns the pipe's status, not the command's.

**Adversarial TDD.** Tests authored and frozen at `4a70021` before implementation, by a separate agent invocation. Git cannot evidence that separation — all agent commits share one identity — so the checkable part is the boundary: `git diff` shows only `read-plugin-json`'s body, and `git status --porcelain` was empty at commit time.

Sibling still open: **claude-skills-315**, `marketplace.json` is opened bare at line 3654.


vinnie357:
## Gate 3 adversarial review (nr-reviewer, fable — did not author the change)

**Verdict: approve.** Attacked on eight fronts in a scratch clone at `a660493`; the change survived all of them. One minor pre-existing sibling found, out of this PR's scope.

### Evidence

**A. Type-check soundness (nu 0.113.1).** Probed 17 value shapes through `describe | str starts-with "record"`: table → `table<...>`, list-of-records → `table<...>`, list, string, nothing, binary, range, closure, cell-path, float, bool, duration, filesize, datetime all fail the check; only records (incl. empty record → bare `record`) pass. No non-record value slips through on this version.

**B. No over-rejection.** Clean corpus at PR head: exit 0, `Total skills: 108`, zero invalid findings. Four unusual-but-valid manifests (deeply nested record, null field, empty `skills` array, unknown extra fields) all return `ok` from `read-plugin-json`.

**C. BOM claim verified independently.** `open` on BOM+valid-JSON: `{threw: false, type: string}` — never throws, exactly as the PR states. The post-parse type check is the only fix shape that catches it.

**D. Five-shape corpus proof reproduced.** `[1,2,3]`, `"juststring"`, `null`, empty file, BOM+valid each written to `plugins/core/.claude-plugin/plugin.json`: every run reports `core/.claude-plugin/plugin.json:invalid_json`, prints `Total skills: 94`, and completes (exit 1 from the finding). BOM fixture verified via xxd: `efbb bf7b`. 94 + core's 14 skills = 108, confirmed by `jq '.skills | length'`.

**E. No passenger tests.** Against the `4a70021` stub: cases 1–5 fail RED (all report `got status: ok`), 6–8 pass — as designed. Mutation-tested the three green guards, one edit each: `str starts-with` → `== "record"` caught by case 6 (all three assertions fire); missing-branch → `invalid` caught by case 7; catch-branch → `ok` caught by case 8. Every frozen case is load-bearing.

**F. Siblings intact.** Malformed `{ broken` corpus manifest still reports `invalid_json` + `Total skills: 94` (#271 preserved). Broken root `outputStyles` still yields `root-manifest / plugin-json / missing_path` with the run completing at 108 (#312 preserved).

**G. PR-body claims all check out**, including the `save -f --raw` note: `0x[EF BB BF] | save` to a `.json` path without `--raw` produces exactly 23 bytes of literal text `[\n  239,\n  187,\n  191\n]` (xxd-confirmed). Impl-commit boundary `git diff 4a70021..b83b865` touches only `read-plugin-json`'s body — no test edits.

**H. Mechanical.** `mise run test` exit 0 in the scratch clone (includes test:disclosure); `--self-test` exit 0, "Non-record-manifest self-test passed (8 cases)"; remote CI: version bumps pass, validate pass; no baseline entries in the diff; no attribution in any of the three commits.

### Finding (minor, pre-existing, out of scope)

A manifest that IS a record but lacks `name` — e.g. `{"foo": 1}` — passes the new check and still aborts the whole run at `test/validate-skills-quality.nu:3937` (`where ns == $plugin_json.name`, `column_not_found`); `Total skills` never prints. Pre-existing (records passed straight through before this PR too), same crash class one shape over — the record-shaped analogue of what 314 was to 313. Recommend a follow-up bees issue alongside the already-filed claude-skills-315.

vinnie357:

**Approve at `a660493`.** One finding, already filed independently before the review landed.

### The soundness question was the whole review

The fix is a single `describe | str starts-with "record"` — a **prefix match on a type name**, so anything describing as `record`-adjacent would slip through and crash at the field access.

The reviewer probed **17 value shapes**; I had probed 10 separately. Same result: only records pass. Tables describe as `table<...>`, lists-of-records also `table<...>`, and string, null, binary, range, closure, and cell-path are all rejected. An empty record `{}` describes as the bare word `record` and passes, which is correct — it is a record.

The inverse held too: clean corpus at `Total skills: 108`, zero `invalid`, and deep-nested, null-field, empty-skills, and extra-field manifests all `ok`.

### Every frozen guard is load-bearing

Cases 6-8 pass before the fix, which makes them look like passengers. The reviewer mutation-tested all three:

| Guard | Mutant | Caught by |
| :--- | :--- | :--- |
| well-formed → `ok` | `== "record"` instead of `str starts-with` | case 6 |
| missing stays `missing` | move the type check ahead of the exists branch | case 7 |
| malformed stays `invalid` | bypass the #271 try/catch | case 8 |

That mutant for case 6 is the one that matters — `== "record"` is false for **every** well-formed manifest, so it would turn the five target cases green while rejecting the entire corpus.

### The BOM reasoning checked out

Independently confirmed: `open` on BOM + valid JSON returns `{threw: false, type: string}`. It never raises, so a wider `try/catch` could not have reached it. That case alone is why the fix had to be a post-parse type check.

### Sibling behaviours intact

Malformed `{ broken` still reports `invalid_json` plus totals (#271), and a broken root `outputStyles` still yields `root-manifest/missing_path` (#312).

### The finding — already filed as claude-skills-316

A manifest that **is** a record but lacks `name` still aborts at `where ns == $plugin_json.name`. I found this while probing soundness and filed it before the review arrived; the reviewer found it independently. Convergent discovery on the same gap.

It is the third shape of one class, and out of scope here — 314's acceptance criteria and frozen tests are explicitly about non-record shapes, so widening would have meant rewriting a frozen contract mid-flight:

| Issue | Shape | State |
| :--- | :--- | :--- |
| 313 | fails to parse | merged |
| 314 | parses to a non-record | this PR |
| 316 | is a record, missing required fields | filed |

PR #273, opened 2026-08-27
fix(validate-skills-quality): report manifests missing a required name

vinnie357:
Closes claude-skills-316. **Third of four shapes of one crash class** — see the correction below; an earlier draft of this body claimed it was the last.

| Issue | Manifest shape | State |
| :--- | :--- | :--- |
| 313 | fails to parse | merged |
| 314 | parses to a non-record | merged |
| **316** | is a record, missing `name` | this PR |
| 317 | `name` present but not a string | filed |

`{}` and `{"foo": 1}` both aborted the run with `column_not_found`; `Total skills` never printed.

### The field audit was measured, not guessed

The issue's AC required enumerating what the validator actually dereferences unguarded. `name` is the **only** one, at three sites, all in the Pass-1 registry loop — `:3923`, `:3937`, `:3940`. Every other `$plugin_json` access uses `get -o`, which returns null rather than throwing. Verified independently by me and by the test author; both lists agree.

So the fix requires `name` and nothing else. No schema validation for `version` or `skills` — nobody asked, and neither can crash the run.

**That conclusion was too strong, and the review falsified it.** The audit enumerated unguarded *dereferences* and missed a second axis: a **string-typed function parameter** also throws. A `name` of `123` or `["core"]` passes the presence check, then aborts at `evaluate-agent-file`'s typed `$plugin_name`. Pre-existing — the same crash occurs on `origin/main` — but "nothing else can crash the run" was false when written. Filed as **claude-skills-317**, and the helper's contract comment now records the type axis as a known residual.

### A distinct status, not a reused one

`status` is now a closed set of four: `ok` | `invalid` | `missing` | `missing_required_field`.

Reusing `invalid_json` was rejected on the grounds that a parse failure and a schema gap are different states — the same distinction the helper already draws between `missing` and `invalid`. A maintainer reading `invalid_json` on a file that is perfectly valid JSON would be misled.

### The frozen test that constrained the implementation

Case 3 uses `{"foo": 1}` — a record **with** fields but no `name`. A check written as "is the record empty?" satisfies `{}` and fails this one. The shipped check is `($parsed | get -o name) == null` — a presence test, which is what the frozen case forced.

**Proven on the real corpus:** both shapes report `core/.claude-plugin/plugin.json:missing_required_field` and still print `Total skills: 94`.

**Siblings intact:** `{ broken` still reports `invalid_json` (#313), `[1,2,3]` still reports `invalid_json` (#314), and a deleted manifest still skips silently. A fourth status did not disturb the first three.

**Adversarial TDD.** Tests frozen at `537b70f` before implementation by a separate agent invocation. Git cannot evidence that separation — all agent commits share one identity — so the checkable part is the boundary: `git diff` touches only `read-plugin-json` and the call sites, `git status --porcelain` empty at commit time.

Remaining sibling: **claude-skills-315**, `marketplace.json` is still opened bare at line 3654.


vinnie357:

**Approve at `fdfd602`** after amendment. One major finding, in the PR's claims rather than its code.

### The finding: my field audit's conclusion was falsified

The PR stated `name` is the only unguarded dereference — true, verified independently at three sites — and concluded **"so the fix requires `name` and nothing else."** That conclusion is false.

A `name` that is present but non-string passes the presence check and still aborts the run. `{"name": 123}` and `{"name": ["core"]}` both die with `cant_convert` at `evaluate-agent-file`'s **string-typed `$plugin_name` parameter** — not at a dereference at all.

That is the sharper half of the diagnosis. I found the crash while probing; the reviewer found *why*. **A dereference audit structurally cannot see a type axis.** Pre-existing — the same crash occurs on `origin/main` — but the claim was false when written.

Filed as **claude-skills-317**. The helper's contract comment now records the type axis as a known residual, and the PR body's "third and last shape" is corrected to third of four.

### My reason for filing rather than widening was also wrong

I said folding a type check in would mean rewriting a frozen contract. The reviewer checked: **false.** Extending the check to `(get -o name | describe) != "string"` passes all seven frozen cases unmodified — I reproduced this in a scratch clone, zero failures. No frozen assertion constrains the type axis.

The sound reason is TDD discipline: a widened check would be a branch **no RED test forced and no test pins** — the untested-code shape the adversarial split exists to prevent. Same disposition, sounder footing. Corrected on 317 so the next implementer does not inherit a constraint that does not exist.

### Attack B came back clean, and it was the one that mattered

A fourth status could silently fall through a branch written for three. Every site was located and checked: `check-root-manifest` handles all four explicitly, the Pass-1 loop handles all four, and both Pass-2 sites use catch-all comparisons (`!= "ok"`, `== "ok"`) rather than three-way enumeration — so a fourth or fifth status skips safely. Structurally, a flagged plugin never enters `$registry`, so Pass-2 cannot reach one regardless, and the Pass-1 finding is already filed so the skip loses nothing.

### Also verified

- **Case 2 is not a passenger** — mutating the new branch to `file: null` kills case 2 alone, exactly as the test author claimed.
- All four other green guards mutation-killed individually.
- `{"name": ""}` does **not** crash; it degrades to `/mise`-style finding keys and exits 1 via baseline mismatch. Left for 317 to decide deliberately rather than by accident.
- Siblings intact: #313, #314, #312, and deleted-manifest all behave as before.

### Two smaller fixes

The helper's contract comment still enumerated three statuses after the fourth was added below it. And a self-test comment cited "around line 3892" for a loop that had shifted to :4051 — stale at commit time, from the comment's own insertion. Both corrected; the line number is now a name rather than a number.

Four shapes of one crash class, three closed:

| Issue | Shape | State |
| :--- | :--- | :--- |
| 313 | fails to parse | merged |
| 314 | parses to a non-record | merged |
| 316 | record missing `name` | this PR |
| 317 | `name` present, not a string | filed |

PR #274, opened 2026-08-27
fix(validate-skills-quality): halt cleanly on an unusable marketplace.json

vinnie357:
Closes claude-skills-315. Last unguarded read in the file.

`marketplace.json` was opened bare, so malformed JSON aborted the run with a raw nushell parser trace and `Total skills` never printed.

### A claim in an earlier draft of this body was false — corrected

That draft argued `{"plugins": null}` produced a silent vacuous pass on `main`, printing `Total skills: 0` with everything green. **It does not.** Measured on `origin/main`: it exits **1** with `No skills found to validate.` A downstream empty-corpus guard already catches it.

The `for` loop does iterate zero times silently, as the test author measured — but the run does not survive to a green finish. Neither of us checked the baseline before building an argument on it.

What is actually true on `main`, measured:

| Shape | exit | `nu::shell::` traces | prints totals |
| :--- | :--- | :--- | :--- |
| `{ broken` | 1 | **3** | no |
| `[1,2,3]` | 1 | **1** | no |
| `{}` | 1 | **1** | no |
| `{"plugins": null}` | 1 | 0 | no |
| `{"plugins": []}` | 1 | 0 | no |

So the real defect is narrower than claimed: three shapes leak a **raw parser trace**, and the two null/empty shapes are caught but **misdiagnosed** — `No skills found to validate.` describes a symptom, not the unusable manifest that caused it. Nothing ever passed vacuously.

The fix still earns its place on both counts: no shape leaks a trace, and every unusable shape now names the file and the reason. But it is an error-message and diagnosis fix, not the safety fix the earlier draft described.

### Behaviour

A new `read-marketplace-json` mirrors `read-plugin-json`'s status contract. `main` branches on status and halts with a message naming the file. All four unusable shapes now behave identically — exit 1, filename in stdout, **no `nu::shell::` trace**, and `Total skills` never printed:

| Shape | Status |
| :--- | :--- |
| `{ broken` | `invalid` |
| `[1,2,3]` | `invalid` |
| `{}` | `missing_required_field` |
| `{"plugins": null}` | `missing_required_field` |
| deleted file | `missing` |

Absent `plugins` and `plugins: null` collapse to one status, matching how `read-plugin-json` already treats a missing `name` — `get -o` cannot distinguish them and neither is usable.

### Scope declined, with a reason

`plugins` present-but-wrong-type is **not** covered. It still throws rather than silently passing, and `validate-marketplace.nu` already schema-enforces it in the same `mise test` gate. Guarding it here would be scope creep beyond this file's own read.

### Dereference audit

`$marketplace.plugins` is the only dereference; per-plugin fields come from the loop variable. Measured, not assumed.

**Adversarial TDD.** Tests frozen at `10830ed` before implementation by a separate agent invocation. Git cannot evidence that separation — all agent commits share one identity — so the checkable part is the boundary: `git diff` touches only the new helper and `main`'s branch, `git status --porcelain` empty at commit time.

Case 7 is a full subprocess run rather than a unit test, since a halt cannot be observed by calling a helper.

This closes the crash class for reads. Still open: **claude-skills-317**, a plugin manifest whose `name` is present but not a string.


vinnie357:

**Approve at `a161201`.** Two findings, both in what was claimed rather than what was built.

### F1 — a fabricated measurement, in three places

An earlier draft argued `{"plugins": null}` produced a **silent vacuous pass** on `main` — `Total skills: 0`, exit 0, everything green. Measured on `origin/main`: it exits **1** with `No skills found to validate.` via a pre-existing empty-corpus guard.

I caught this myself and corrected the PR body. **The reviewer caught the half I missed** — the same fabrication was baked into two shipped code comments, where it would have outlived the PR description entirely and misled whoever read that code next.

All three now state the measured behaviour, and the helper's doc comment says outright that the claim was fabricated and that nobody measured the baseline before arguing from it. A quietly-reworded comment teaches nothing.

The reviewer re-measured every row of the corrected table on `origin/main` rather than accepting it:

| Shape | exit | `nu::shell::` traces | totals |
| :--- | :--- | :--- | :--- |
| `{ broken` | 1 | 3 | no |
| `[1,2,3]` | 1 | 1 | no |
| `{}` | 1 | 1 | no |
| `{"plugins": null}` | 1 | 0 | no |
| `{"plugins": []}` | 1 | 0 | no |

So the real defect is narrower than first claimed: three shapes leak a raw parser trace, and two are caught but **misdiagnosed**. Nothing ever passed vacuously. This is a diagnosis fix, not a safety fix.

### F2 — a mutation survivor: the record check was unpinned

Deleting `describe | str starts-with "record"` left **all seven frozen cases green**. Case 3's `[1,2,3]` throws into the catch and lands on `invalid` regardless, so it never exercised that dimension.

Case 8 (`[{"plugins": []}]` → `invalid`) closes it. Since the check already exists the case was green on arrival and had **no RED phase**, so it was proven the inverse way — mutate, watch it die, revert. Reproduced independently: exactly one failure, Case 8, 19 other suites green.

### The sweep that should have happened the first time

One survivor means the suite was never mutation-tested, so the reviewer mutated **every branch** — seven mutations, zero survivors:

| Mutation | Killed by |
| :--- | :--- |
| `missing` → `invalid` | Case 1 |
| record check → `if true` | Case 8 alone |
| `missing_required_field` → `ok` | Cases 4 + 5 |
| non-record else → `ok` | Cases 3 + 8 |
| catch → `ok` | Case 2 + Case 7 |
| main guard → bare `open` | Case 7 |
| `exit 1` dropped | Case 7 |

Case 7 — the subprocess test — was the sole killer of the two mutations that revert `main` itself. A unit test could not have caught either.

### Method note

Both of us hit the same trap: a scratch clone without `git init` makes the validator exit 128, which reads as "the mutant survived" rather than "the harness broke." Worth knowing for anyone mutating this file.

PR #275, opened 2026-08-28
docs(claude-agents): bracket every segment of the agent spawn naming convention

vinnie357:
- Give the spawn name shape as `<issue>-<role>-<model>-<n>` with all four segments bracketed, plus a table naming what belongs in each
- State that no segment is a literal, and record the Agent tool's `name` pattern the assembled string must satisfy
- Replace every example carrying the literal `task-` prefix with a cross-issue sequence that demonstrates the counter
- Resolve the contradiction between the PER-SESSION sentence and its per-model example: the counter is one global series per session
- Reduce the two restatements in `agent-loop` to an example plus a pointer, so the rule lives in exactly one file
- Bump core, claude-code, and all-skills

Closes claude-skills-318

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "all-skills",
       "source": "./",
       "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
-      "version": "0.4.166",
+      "version": "0.4.167",
       "category": "meta",
       "keywords": [
         "skills",
@@ -119,7 +119,7 @@
       "source": "./plugins/tools/claude-code",
       "strict": true,
       "description": "Claude Code-specific skills for plugin marketplace management, validation, and component creation",
-      "version": "0.2.83",
+      "version": "0.2.84",
       "category": "tools",
       "keywords": [
         "claude-code",
@@ -143,7 +143,7 @@
       "source": "./plugins/core",
       "strict": true,
       "description": "Essential development skills: Git, TDD, code review, gitleaks secret scanning, documentation, technical English, restraint, anti-fabrication, twelve-factor apps, the agent loop, bees issue tracking, mise, nushell, and Apple Container",
-      "version": "0.2.75",
+      "version": "0.2.76",
       "category": "development",
       "keywords": [
         "git",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "all-skills",
-  "version": "0.4.166",
+  "version": "0.4.167",
   "description": "Meta-plugin that installs all skills from all plugins in the marketplace",
   "author": {
     "name": "vinnie357"
--- a/plugins/core/.claude-plugin/plugin.json
+++ b/plugins/core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "core",
-  "version": "0.2.75",
+  "version": "0.2.76",
   "description": "Essential development skills: Git, TDD, code review, gitleaks secret scanning, documentation, technical English, restraint, anti-fabrication, twelve-factor apps, the agent loop, bees issue tracking, mise, nushell, and Apple Container",
   "author": {
     "name": "vinnie357"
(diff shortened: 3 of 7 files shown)


vinnie357:
**Gate 3 adversarial review (fable, defeat-the-change brief) — APPROVE, 2 non-blocking suggestions.** Reviewed at pinned head fcb23c71e218cca0d72b2b726c7a7db037e533c0 via `git show`/`git diff origin/main...<sha>` (no checkout; `gh pr diff` not trusted per prior stale-content incident).

**Attacks run and results:**

1. **AC 1–4 of claude-skills-318 — satisfied.** (1) First segment bracketed and defined: `plugins/tools/claude-code/skills/claude-agents/SKILL.md:124` "Build the `name` parameter from four segments: `<issue>-<role>-<model>-<n>`. Every segment is a placeholder. None of the four is a literal." (2) No example uses the literal string `task`: `git grep -E 'task-(sonnet|opus|haiku|fable|<model>)'` at head matches only `.bees/issues.jsonl` historical records (the 318/293 issue texts themselves) — acceptable. (3) Counter rule stated at SKILL.md:154 "The counter is ONE GLOBAL SERIES PER SESSION." (4) agent-loop/SKILL.md:227 and leader-spawn-example.md:57 both state `<issue>-<role>-<model>-<n>` and point at the exact heading "Agent Spawning Naming Convention", which exists at SKILL.md:122. Repo-wide grep for restatements found NO third file disagreeing — remaining `-sonnet-4`-shaped hits in altana/awman/statusline are model IDs, unrelated. `templates/five-tier-issue.workflow.js` uses `label: P1-<model>` display labels, not Agent `name` parameters — not a conflict.

2. **Counter example genuinely disambiguates.** The sequence `318-test-author-sonnet-1`, `318-impl-sonnet-2`, `318-ci-haiku-3`, `317-plan-fable-4` cannot be produced by a per-model counter (which would yield `...-haiku-1` and `...-fable-1`), so the example proves the global series rather than restating the old ambiguity. leader-spawn-example.md:57–62 is consistent: worked issue `runex-142`, name `142-impl-sonnet-2`, prose "the second agent spawned this session".

3. **Regex claim verified, not fabricated.** The live Agent tool schema's `name` parameter carries exactly `"pattern": "^[A-Za-z0-9][A-Za-z0-9_-]{0,63}$"` — the skill's assertion at SKILL.md:133 matches it byte-for-byte. All six documented example names satisfy it (leading digit allowed by `[A-Za-z0-9]`; longest example is 24 chars, within the 64 cap).

4. **Version bumps verified against the CURRENT origin/main tip (d8accbf), not the merge-base** (they coincide): core 0.2.75→0.2.76, claude-code 0.2.83→0.2.84, all-skills 0.4.166→0.4.167, each in BOTH its plugin.json and its marketplace.json entry. Note: other PRs are in flight this session — if any of them merges first with these same version lanes, this PR's bumps become non-greater and need re-bumping before merge (serialize merges).

**Findings (non-blocking):**

- **[suggestion]** SKILL.md:128 — `<issue>` row: "The tracker issue this agent's team works on (`318`)." The transformation from full tracker id to segment value is demonstrated (`claude-skills-318`→`318`, and leader-spawn-example.md derives `142` from `runex-142`) but never stated. A reader holding `runex-142` or `VIN-83` can defensibly write `runex-142-impl-sonnet-2` — same demonstrated-but-unstated pattern class the original bug came from, though here both readings produce valid, glanceable names. Fix: append one clause, e.g. "use the numeric part of the issue id (`claude-skills-318` → `318`)."
- **[suggestion]** The rationale sentence deleted from all three files ("This naming convention allows operators to see which model each running agent activated with at a glance" — old SKILL.md:124) now appears nowhere: `git grep -i 'at a glance'` at head has no hit in these files. The decoding sentence at SKILL.md:141 ("Read `318-ci-haiku-3` as the CI runner...") carries the mechanism but not the purpose. The 23-agent incident in the issue happened precisely because a lead dropped the model segment; the WHY is the guard against a future editor doing it again. Fix: one clause in the canonical section, e.g. "so an operator can read team, model, and spawn order from the name alone."

Checked and ruled out: leftover literal-`task` examples outside issue history; stale restatements in claude-teams, claude-agents references/templates, forge/dispatch-discipline; heading drift in the two pointers; per-model counter leakage in any example; regex violations in any documented name; missing all-skills bump.

PR #276, opened 2026-08-28
fix(validate-skills-quality): report a non-string manifest name instead of aborting

vinnie357:
- Add `invalid_field_type` status to `read-plugin-json` for a `name` present but not a string, checked after the null check so `{"name": null}` still reports `missing_required_field`
- Add the matching arm to `check-root-manifest`
- Add the matching arm to the Pass-1 registry loop, keyed off marketplace.json's `$plugin.name`, so a bad manifest is reported and skipped instead of aborting the whole run
- Add `run-non-string-name-manifest-self-test` (13 cases): int/list/record/bool must-flag, null and missing-key regression guards, `{"name": ""}` pinned as a deliberate non-goal, 313/314 regression guards, real-corpus guard, an end-to-end subprocess case over a scratch clone, and a `check-root-manifest` case
- Record the two residuals this leaves open in `read-plugin-json`'s doc comment: claude-skills-322 (empty-string name degrades finding keys) and the AC4 audit findings claude-skills-320 (`skills` scalar) and claude-skills-321 (`outputStyles` element type)
- File claude-skills-320, 321, 322, 323

Closes claude-skills-317.


vinnie357:
## Gate 3 adversarial review — PR #276

Reviewed at head sha `a7db71de8fc0729b8c6877f44525036b8227b826` (content read via `git show`/`git diff` at that sha, not a cached `gh pr diff`). Brief: defeat the change. Every claim below is backed by a command run against a scratch clone at the pinned sha; the shared working tree was not touched.

### Findings

**F1 [nit, docs] — stale line number in the residual doc comment.** `read-plugin-json`'s doc comment (claude-skills-320 bullet) says the `skills`-scalar crash aborts "at the `each {|p| $p | str replace ...}` around line 4380". In the file this PR ships, that code sits at lines 4874–4876, and an actual run confirms it: injecting `{"skills": 7}` into wasm's manifest aborts with `nu::shell::only_supports_this_input_type` at `validate-skills-quality.nu:4875:13`. The PR's own ~470 inserted lines moved the site; 4380 was true of the pre-PR layout. The sibling claim for claude-skills-321 stays accurate (measured: `outputStyles: 5` on the root manifest aborts `nu::shell::cant_convert` at `3208:60`). Fix: update the number or drop it and keep the function-name reference. Non-blocking.

**F2 [observation, no action required] — non-empty weird string names are silently accepted.** Injected wasm manifest names `"a/b"`, `"a\nb"`, `"日本語"`, and a 5000-char string: all four return `ok`, the run completes, exits 0, and produces no finding — the manifest name is never cross-checked against marketplace.json's name. Same reporting-quality family as claude-skills-322 (which covers only `""`), correctly outside 317's crash-class scope. Worth a line on 322 so the widening decision considers the whole family, not just the empty string.

### Verification record

- **Type check (axis 1).** `describe` measured for every JSON shape: float→`float`, int→`int`, bool→`bool`, list→`list<int>`, record→`record<...>`, null→`nothing`, `""` and any string→`string`. No usable name fails the check; nothing unusable passes except the deliberate string family above. Float — a shape the suite does NOT cover — injected end-to-end: flagged `wasm/.claude-plugin/plugin.json:invalid_field_type`, run completes, `Total skills` prints, exit 1, no raw trace. The branch generalizes past the four tested types.
- **Ordering (axis 2).** Required: `null | describe` is `nothing` ≠ `string`, so a type-check-first ordering would misclassify `{"name": null}` as `invalid_field_type`; Case 6 pins `missing_required_field`. With the shipped ordering, the mapping is total and unambiguous: absent/null→missing_required_field, non-string→invalid_field_type, string→ok.
- **Mutation testing (axes 3, 5).** Three mutants, each run through the full `--self-test` at the pinned sha:
  - Delete the `read-plugin-json` branch → Cases 1–5, 12, 13 red (Case 12 reproduces the original crash: raw trace, no total).
  - Delete the Pass-1 registry arm only → ONLY Case 12 red (three assertions fire). Cases 1–11 stay green — confirming Case 12 is the only case pinning AC1 and is not removable.
  - Delete the `check-root-manifest` arm only → ONLY Case 13 red, reporting the `[threw]` sentinel — the try/catch kept the harness alive exactly as its comment claims.
  No must-flag case is a passenger; Cases 6–11 are declared regression guards and behave as such.
- **Regression surface (axis 4).** origin/main's script and the PR's script run against the identical corpus clone: outputs byte-identical (ANSI-stripped diff empty), both exit 0. The new `check-root-manifest` arm changes nothing for the real root manifest (name "all-skills" is a string).
- **Cost claim (axis 5).** Full `--self-test` at the sha: 28.9s, exit 0, all 21 suites green — consistent with the doc's "~0.4s → roughly 27s" claim. The cost is carried by the only case mutation testing shows to be load-bearing for the wiring, and CI's validate job passes at 2m14s. Accepted as the documented trade.
- **Residuals (axis 6).** All four doc/bees claims re-measured in scratch clones: `{"skills": 7}` aborts `only_supports_this_input_type` (claude-skills-320, line drifted per F1); `{"skills": "<string>"}` does NOT abort, run completes; root `{"outputStyles": 5}` aborts `cant_convert` at 3208 (claude-skills-321); `{"name": ""}` on core completes with exit 1 and degraded bare-slash keys (`/mise:anti_fab`, `/nushell:anti_fab`) exactly matching claude-skills-322's description; on a plugin with no findings (wasm) it completes with exit 0. bees 320, 321, 322, 323 all exist and match what the PR says they track.
- **Mechanical (axis 8).** 9 commits, all single-line conventional, attribution grep clean (`co-authored-by|signed-off-by|assisted-by|generated with` → no hits). `test/quality-baseline.json` untouched (0-line diff); no new baseline entry (AC6); wasm confirmed absent from the baseline, validating Case 12's passenger argument for core. All fixtures live in `mktemp -d` or a `git clone --local` temp dir and are removed; nothing written to the real repo. `gh pr checks`: both checks pass.
- **Prose (axis 7).** The strict-register contract sentence and residual list are checkable and, except F1, check out. The comment depth matches the established idiom of the 313/314/315/316 family and carries measured evidence rather than filler; no clutter finding.

### Verdict

**APPROVE.** No blocking items. F1 is a one-line doc correction the author may fold into this PR or the next touch of the file; F2 is a note for claude-skills-322.

PR #278, opened 2026-08-31
feat(core): add memory-auditor agent for decay-checking persistent memory

vinnie357:
- Add `memory-auditor` agent: audits persistent memory, deletes decayed entries, routes work items to trackers
- Extend `memory-guidance.md` with the work-vs-fact boundary; retire the `project` metadata type
- Add five decay checks (superseded, stale referent, fragmentation, index drift, shape violations) distinct from the existing four write-time questions
- Agent snapshots to a sibling `memory-archive/` before any delete; the memory directory is not git-tracked
- Agent emits `## BEES REQUESTS` rather than writing to any tracker itself, per the single-writer rule
- Bump core 0.2.77 -> 0.2.78 and all-skills 0.4.168 -> 0.4.169


vinnie357:
## Gate 3 adversarial review

Reviewed at head `fed53982` (the brief expected `3646ee1`; `fed53982` stacks one `.bees/issues.jsonl` bookkeeping commit on top of it — the substantive diff is unchanged).

### CONFIRMED findings, most severe first

1. **[blocking] `plugins/core/agents/memory-auditor.md`, "Scope of autonomous action" (a): the orphaned-file delete misfires on a normal transient state.** The memory-writing convention is write-the-file-first, add-the-`MEMORY.md`-line second. A session that dies between the two leaves a valid, durable entry whose only defect is a missing index line — and this agent's most-privileged action (delete without asking) fires on that bookkeeping defect alone, content never evaluated. The two drift directions are not symmetric: deleting a dangling index line is genuinely judgment-free; deleting an orphaned FILE is a content judgment. Fix: auto-fix only the dangling-index-line direction; route orphaned files through the normal per-file evaluation (or propose indexing them).

2. **[blocking] The `RESOLVED` marker routes to two different autonomous dispositions with no tiebreak.** Decay check 1 (`memory-guidance.md:152-153`) greps for `RESOLVED` and says rewrite — state the current rule, drop the history. The work-vs-fact boundary (`memory-guidance.md:125`) calls a closed-but-still-asserted item ("RESOLVED 2026-07-28") dead weight, which the agent's delete category (b) removes without asking. Same literal input, delete vs rewrite, undefined which wins. Concrete case: an entry titled "OTP skill refactor — RESOLVED via references, not a skill split" is a durable decision record under check 1 and dead work under (b). State the precedence.

3. **[blocking] The agent file's "dead work" definition is broader than the guidance question it cites.** "(an entry describing work that is closed, resolved, or shipped)" literally matches feedback entries that narrate a closed incident as evidence for a durable lesson — which most feedback memories do. The guidance's own discriminator (still to be *done* vs now *known*) keeps them; the agent file's parenthetical deletes them, and the agent file is the text that grants the authority. Also in this area: nothing requires verifying closure (a bees/gh lookup) before a dead-work delete — the evidence mandate in "Verify before claiming" scopes itself to stale-referent verdicts, and the output contract's "or the specific decay check that fired" lets a delete ship with no verification at all.

4. **`memory-guidance.md:3` contradicts the new content in the same file.** The unchanged intro still enumerates "project notes" as a memory category while line 140 retires the `project` type. (Repo-wide grep at HEAD found no OTHER file documenting `type: project`, so the contradiction is contained to this file.)

5. **`plugins/core/skills/agent-loop/SKILL.md:377` under-describes the grown file.** The one-liner covers only the original write-time generalization content — nothing about the work-vs-fact boundary, decay checks, or the type retirement, which now make up ~40% of the file.

6. **[nit] Split/merge asymmetry in the authority table.** "Rewrite in place" includes "split a multi-topic document," which re-forms operator wording exactly the way a merge does — yet merge is propose-only and split is autonomous. State why, or move split to propose.

### PLAUSIBLE (constructed, not observed)

7. **Snapshot verification compares file COUNTS only.** An archive built from the wrong path with a coincidentally equal count passes verification; comparing the listed names against the directory listing closes this for one more line. Every other snapshot failure mode I could construct (tar error, empty glob, `./` off-by-one in the count) biases toward a mismatch or error, and the "fails for any reason — STOP" clause correctly halts on those — the fail-safe direction. [nit] The section header says "before any delete" while the body correctly gates rewrites too; body governs.

### Checked and clean

- Versions: core 0.2.77 -> 0.2.78 and all-skills 0.4.168 -> 0.4.169 in all four locations; `origin/main` sits at 0.2.77/0.4.168, so no version-lane conflict.
- Agent registered in `plugins/core/.claude-plugin/plugin.json`; the path resolves to a real file at HEAD; frontmatter has no `allowed-tools`; description is third person with a `Use when` trigger; the root all-skills `plugin.json` carries no `agents` array (consistent with the prose-reviewer precedent).
- PR body: all six bullets match the diff; no overstatement found.
- Restraint: the agent is commissioned by claude-skills-325; 108 lines are load-bearing (snapshot gate, authority scoping, single-writer rule, output contract); the sonnet-over-haiku choice is justified in-body; no speculative config knobs.
- Attribution grep over both branch commits: clean.
- Per the brief I did not run `mise test`; static facts above were verified by direct reads at the pinned sha.

### Verdict: BLOCK

Findings 1–3 all sit on the autonomous-delete surface, which is this change's entire risk. Each is a small text fix in `memory-auditor.md` (plus one sentence in `memory-guidance.md` for #2 and a two-word edit for #4); nothing structural.

vinnie357:
## Gate 3 adversarial review — second pass (post-fix)

Reviewed at head `88870cd` (matches expected). Diff examined: `3646ee1..88870cd`. Brief: assume the fix pass introduced a new defect on the autonomous-delete surface; find it.

### CONFIRMED findings, most severe first

1. **[blocking] `memory-auditor.md` Workflow step 2 now mutates `MEMORY.md` BEFORE the step-3 snapshot, contradicting "Snapshot before any delete or rewrite — no exception".** The fix pass changed step 2 from detection-only to "Fix a dangling index line immediately (delete-without-asking)"; the snapshot is step 3. `MEMORY.md` lives in the memory directory; deleting a line from it is a rewrite of an untracked file. Under the pre-fix text every mutation happened at step 5, after the snapshot. Scenario: index lines use the `[Title](file.md)` markdown-link format; the agent cross-checks with absolute paths or mis-parses the link, judges a VALID line dangling, and deletes it at step 2 — pre-snapshot. The step-3 archive then captures the already-edited `MEMORY.md`, so the original index state is unrecoverable even from the archive. Blast radius is bounded (the entry file survives; a later audit re-indexes it via a2) but the pre-audit index is the one state the snapshot exists to preserve and it is lost. Related gap in the same path: the output contract's verdict table is keyed per FILE, so a dangling-line deletion — the only pre-snapshot mutation — has no row shape and can ship unrecorded. Fix: move the dangling-line fix into step 5 (or after step 3), and give index-line deletions a reporting line.

2. **[blocking] `memory-auditor.md` "Verify before claiming" names a delete category the Scope section never grants.** The fix-pass rewrite created "every delete verdict — stale-referent or dead-work —". Scope of autonomous action limits delete-without-asking to exactly (a1) dangling index lines and (b) dead work; a stale referent is neither, and `memory-guidance.md` check 2 says only "the entry MAY be dead", authorizing name-fix, not delete. A literal reader of the Verify section concludes stale-referent deletes are sanctioned (with grep evidence) and deletes an entry whose named flag greps empty — e.g. an installed-binary gotcha like "use `bees label add`, not `bees update --labels`", whose referent never appears in any repo's text and always greps empty. This is a new implied delete path the old wording ("every stale-referent verdict cites...") did not create. Fix: "every delete verdict — both authorized categories, (a1) and (b) —" or equivalent.

3. **[blocking] Dead-work closure verification can be satisfied with wrong-repo evidence — the exact hazard the stale-referent branch guards against is missing from the new branch.** Stale-referent verdicts require "confirmation the grep targeted the right repo: an empty result from the wrong directory is not evidence." The new dead-work requirement says cite "a tracker lookup (`bees show <id>`) or a repo check (`git log --grep`, `gh pr view`)" with no right-repo clause and no statement of what the result must show. Concrete, verified scenario: the agent file's own worked dead-work example reads "merged PR #204"; run `gh pr view 204` from a claude-skills cwd and it returns state MERGED — for claude-skills' own unrelated PR #204 (verified: merged 2026-08-04). An entry whose epic actually got reverted in its home repo passes closure verification on a stranger's PR number and is deleted. Also unstated: whether a failed or empty result (`bees show` not-found, empty `git log --grep`) blocks the delete — the text only requires that a command be CITED. Fix: mirror the stale-referent clause ("confirm the lookup targeted the tracker/repo the entry names; a not-found or empty result is not evidence of closure — it blocks the delete").

### PLAUSIBLE (constructed, not observed)

4. **rewrite vs split collision through `memory-guidance.md` check 5.** Check 5 still lists "a file that reads as a multi-topic document" under shape violations, and the agent grants "Rewrite in place — ... shape violations" categorially; the fix pass removed splitting only from the agent's parenthetical examples. A literal agent following the guidance's definition of shape violations can restructure a multi-topic file autonomously, contradicting the propose-only split bullet. Specific-beats-general saves a careful reader; nothing states the precedence. One-line fix in check 5: note that the multi-topic case takes check 3's propose-only disposition.

5. **Rewrite-to-empty bypass of the new precedence rule.** An entry that is pure closure status AND carries a `RESOLVED` marker ("RESOLVED 2026-07-28: the flaky test issue is fixed") is forced by the precedence rule into rewrite — which requires no cited evidence — and its durable half is nil, so the rewrite guts it to nothing: a delete in effect without category-(b)'s closure verification. Snapshot backstops recovery. Suggested line: a rewrite that would leave no rule behind reverts to category (b) and its verification requirement.

6. **`bees-manager.md` MAY/MUST NOT boundary licenses AC rewording gated only on self-judged equivalence.** "MAY tighten its wording" spans acceptance criteria, and the MUST NOT bans rewording an AC only "in a way that changes what satisfies it" — so a haiku-tier relay may reword an AC whenever it judges its own reword satisfaction-preserving, which is precisely the judgment the relay-not-author section distrusts. Consider: tightening allowed for titles/descriptions only; ACs and identifiers verbatim, or every AC reword reported to the lead. Frontmatter `skills:`, in-body Skills list, and the report quote block all agree (3 skills each) — clean.

7. **[nit] Stop-clause enumeration is stale.** "permission error, disk error, count mismatch" survives from the count-based check; step 3 now compares names. "For any reason" governs, but the enumeration no longer names the check's actual failure mode.

### Checked and clean

- Snapshot integrity otherwise: the name-match check closes the wrong-path/equal-count case; remaining constructed failures (tar read error, `./` prefix mismatch in listings, empty glob) all surface as errors or mismatches and the STOP clause halts — fail-safe direction. Content-level corruption with intact names remains theoretically uncovered but tar errors on read failure; acceptable.
- Orphaned-file path (a2): re-index default cannot lose data; an orphaned dead-work entry still goes through per-file evaluation and category-(b) verification before delete. No collision found between re-index and the other dispositions.
- Dead-work definition stress: entries narrating closed incidents as evidence for live rules (the `ps -E` shape, "fixed X by adding on_exit cleanup") land on keep under the remove-the-closure-sentence discriminator; I could not construct a real-shaped feedback entry the new wording deletes wrongly.
- Agent/guidance consistency: precedence rule, a2 asymmetry, and the closure-narration discriminator are stated in both files with matching semantics; guidance line 3 no longer lists "project notes"; SKILL.md one-liner updated.
- Versions: core 0.2.77→0.2.78, all-skills 0.4.168→0.4.169, synced across both plugin.json files and marketplace.json; merge base equals current origin/main HEAD, so no version-lane conflict.
- Agent registered in `plugins/core/.claude-plugin/plugin.json`; no `allowed-tools` in frontmatter; root all-skills plugin.json carries no agents array (consistent with precedent).
- Restraint: the added text (worked pair, precedence, closure verification, a2 rationale) is safeguard, not hedging; the a2 rationale is duplicated near-verbatim across agent and guidance, acceptable given agent-file-wins doctrine.
- Attribution grep over branch commits and this comment body: clean. Per the brief I did not run `mise test`.

### Verdict: BLOCK

Findings 1–3 are all on the delete surface and all introduced or left exposed by the fix pass's own rewording — the predicted failure mode. Each is a small text fix in `memory-auditor.md` (plus one line in `memory-guidance.md` for #4); nothing structural.

vinnie357:
## Adversarial review — pass 3 (HEAD 6a58a89)

Method: read all three files at 6a58a89 via `git show` (no checkout), diffed 88870cd..6a58a89, verified versions in all four locations. Numbered the workflow as an execution trace, enumerated all seven dispositions and their collisions, and tested every evidence requirement for vacuous satisfaction.

What holds at this head: no mutation of any kind precedes the step-3 snapshot (step 2 is now "Detection only: mutate nothing in this step"); the dead-work delete requires closure verification against the entry's own repo/tracker with not-found/ambiguous BLOCKING the delete; the rewrite-to-nothing exception is present with identical text in both the agent file and the guidance; the multi-topic case routes propose-only; bees-manager's MAY/MUST NOT lists compose without contradiction and its three skills lists (frontmatter, body, report block) agree; versions are bumped and synced (core 0.2.77→0.2.78, all-skills 0.4.168→0.4.169, plugin.json + marketplace.json both); the agent is registered in core's plugin.json, and all-skills carrying no agents field matches the prose-reviewer precedent.

### CONFIRMED

1. **memory-auditor.md:126-131 + :51 — deleting a dead-work file leaves its own index line dangling.** Workflow step 5 instructs deleting only the dangling lines "recorded in step 2"; nothing ties a category-(b) file delete to removal of that file's MEMORY.md line. Scenario: the audit deletes `project_foo.md` as verified dead work, MEMORY.md still lists it, and the audit ends by creating exactly the check-4 drift (guidance line 182: one line per existing file) it exists to fix — invisible in the report, since "Deleted index lines" covers only lines removed under (a1). Non-destructive; the next audit self-heals it. Fix is one clause: a file delete also deletes that file's index line, reported under Deleted index lines.

2. **memory-guidance.md:170-171 vs memory-auditor.md:87-89 — referent-name fixes authorized in guidance, absent from the agent's rewrite scope.** Check 2 says a moved referent means "fix the name," but the agent's autonomous-rewrite scope enumerates only superseded content and shape violations. The two documents the agent loads disagree on whether that rewrite is authorized. Both readings are benign (a post-snapshot name correction), but this is the agent-vs-guidance drift class pass 5 exists to catch. Add referent-name fixes to the rewrite scope or strike "fix the name" from guidance.

### PLAUSIBLE

3. **Rewrite-to-almost-nothing evades the exception (agent:74-77, guidance:165-168).** The exception converts a rewrite to a delete (with closure verification) only when "nothing transferable" remains. A rewrite that keeps one vacuous sentence ("be careful with mocks") stays a plain rewrite gated only by a marker grep — a delete-equivalent gutting laundered as a rewrite, destroying the durable half with no evidence requirement. Bounded: requires a genuine REVERSED/RESOLVED marker plus a durability misjudgment, and the snapshot is the recovery path. No wording fully closes a judgment gap; noting it as residual risk, not demanding a fix.

4. **A blocked delete has no verdict value (agent:67-68, :137-140).** When closure verification returns not-found and BLOCKS a category-(b) delete, the verdict enum (keep/rewrite/re-index/merge/split/delete/route-to-tracker) has no way to say "delete blocked, kept pending evidence." Conservative direction — the entry stays — but the table under-reports why. Cosmetic.

5. **The snapshot gate exists only in the agent file.** memory-guidance.md authorizes deletes directly (check 4 "judgment-free: delete the line"; check 1's exception authorizes a file delete) and never mentions the snapshot. A session applying the decay checks inline from the agent-loop reference — rather than spawning memory-auditor — prunes with no archive. Arguably by design (the agent is the pruning surface); flagging because the brief's parity question cuts both ways.

### Restraint

Round-3 additions each carry a safeguard (wrong-repo `gh pr view` hazard, detection-only step 2, Deleted-index-lines observable, bees-manager verbatim boundary). No decorative hedging accreted. No TODO/FIXME markers in the diff.

### Verdict: APPROVE-WITH-FINDINGS

Both prior BLOCKs were destructive-path defects (mutation before snapshot, vacuously satisfiable delete evidence). Neither confirmed finding here is: #1 is a bookkeeping gap that cannot lose content and self-heals on the next run; #2 is benign drift. Fix or file both; nothing in this head makes the change dangerous to merge.

vinnie357:
## Gate 3 — consolidated finding dispositions

Three adversarial passes, each by a separate reviewer with no context from the previous one. Every finding is dispositioned below; none waved through.

### Pass 1 (head `fed5398`) — BLOCK, 3 blocking
| # | Finding | Disposition |
|---|---|---|
| F1 | Orphaned file and dangling index line treated symmetrically; auto-delete destroys mid-write entries | FIXED `cd41179` — orphaned files re-index by default; only dangling lines auto-delete |
| F2 | `RESOLVED` marker routed to both rewrite and delete, no precedence | FIXED `cd41179` — rewrite wins; stated in both files |
| F3 | Dead-work definition matched entries that cite a closed incident as evidence | FIXED `cd41179` — discriminator is now durable-rule-vs-status-report |
| F4-F7 | Same-file contradiction, stale one-liner, split→propose-only, snapshot verified by name | FIXED `cd41179` |

F3 was verified against a real 56-entry corpus: 10 of 49 feedback entries matched the original wording literally.

### Pass 2 (head `88870cd`) — BLOCK, 3 blocking, all introduced by the pass-1 fix
| # | Finding | Disposition |
|---|---|---|
| B1 | Dangling-line delete ran at step 2, before the step-3 snapshot; archive captured already-mutated state | FIXED `6a58a89` — step 2 is detection-only; deletion moved to step 5 |
| B2 | `stale-referent or dead-work` named a delete category Scope never granted | FIXED `6a58a89` — scoped to the two authorized categories |
| B3 | Closure verification satisfiable by a coincidental identifier in the wrong repo | FIXED `6a58a89` — right-repo clause; not-found/empty/ambiguous BLOCKS the delete |
| P4-P7 | split precedence, rewrite-to-nothing bypass, bees-manager AC rewording, stale STOP wording | FIXED `6a58a89` |

B3 was reproduced concretely: `gh pr view 204` in this repo returns a real merged PR, so a memory citing another repo`&#39;`s PR #204 would have passed verification.

### Pass 3 (head `6a58a89`) — APPROVE-WITH-FINDINGS, 2 confirmed, neither destructive
| # | Finding | Disposition |
|---|---|---|
| 1 | A file delete left its own index line dangling, creating the drift the audit removes | FIXED `81eeab0` |
| 2 | Guidance authorized a referent-name rewrite the agent scope did not enumerate | FIXED `81eeab0` |

Pass 3 confirmed by execution-trace that no mutation precedes the snapshot and the wrong-repo delete hole is closed.

### Not fixed, recorded
PLAUSIBLE items from pass 3 — rewrite-to-one-trivial-sentence evading the rewrite-to-nothing rule; a BLOCKED delete having no verdict-enum value; the snapshot gate living only in the agent file. All judged conservative-direction or by-design. Filed rather than fixed.

### Recommendation
Every blocking finding across three passes sat on the autonomous-delete path. Suggest the first production run target a copy of the memory directory, with the verdict table reviewed before a live run.

PR #286, opened 2026-09-04
fix(core): block merge on a dismissed changes-requested review

vinnie357:
Closes claude-skills-337. Follow-up to #285, from a Gate 3 finding on that PR that was filed rather than blocked on.

Precondition rule 5 restricts each author's latest review to `{APPROVED, CHANGES_REQUESTED}`. A dismissed review reads `state == DISMISSED`, so the restriction drops it — and nothing else in the precondition reads a dismissal. Under `approval`, a merging agent could therefore dismiss the CHANGES_REQUESTED review blocking it, silently restoring that author's earlier approval as their latest qualifying review, and merge a PR its reviewer had objected to.

## Shipped changes to `/core:git`

- Precondition rule 8: when the `gh pr view` read shows a DISMISSED entry, read the issue timeline and allowlist every `review_dismissed` event's `dismissed_review.state` against `{approved, commented}`, blocking on anything else and waiting on a failed or non-array read
- Rule 2 blocks on a `Gate 3: REJECT <40-hex>` marker at head, regardless of any APPROVE at the same head, and whether or not that marker was edited
- The forbidden-command list covers review dismissal and Gate 3 comment deletion
- Residual risk 12; residual risk 9 rewritten to own the Gate 3 record failure modes
- Two false framing claims removed: the precondition no longer claims a single read, and the Merge Strategy recipe no longer repeats that claim
- Five mechanism restatements cut (−59 words, line-neutral), each verified leg-by-leg against surviving text

## Tracker changes

`.bees/issues.jsonl` carries two records: claude-skills-332 closed (merged as 6e665e7), and claude-skills-337 created — the issue this PR closes.

Records 338, 339 and 340 were filed on this branch and have been taken back out of its diff. They stay open in the tracker and land with a later tracker commit. The reason is round 17's attack point 6: `plugins/core/skills/git/SKILL.md` has been byte-identical since `1a58658`, and every one of the nine commits after it edited record 340 — a design spec for work nobody has started. Holding a reviewed skill fix behind seventeen rounds of authoring that spec cost more than it protected.

## Corrections made during review

[...]

PR #288, opened 2026-09-05
feat(git): enforce forbidden merge-path writes with a PreToolUse hook

vinnie357:
Closes claude-skills-340. Follow-up to #286, which shipped the merge-authorization precondition and stated its forbidden writes without enforcing any of them.

`/core:git` forbids `gh pr merge --auto`, `--admin`, an unpinned merge, `glab mr merge` without `--auto-merge=false`, review dismissal, and Gate 3 comment deletion. Skill text is advisory — an agent that never loads the skill, or reasons its way around it, is unblocked. This adds the tool-boundary backstop.

## Shipped

- `plugins/core/skills/git/hooks/posix-scan.nu` (226 lines) — a POSIX scanner returning segments of tokens, each with its dequoted value and a LIVE flag for an expansion the shell would actually run. Segments split on an unquoted `;`, `&&`, `||`, `|`, `&`, or newline.
- `plugins/core/skills/git/hooks/block-forbidden-merge-writes.nu` (269 lines) — a PreToolUse hook exiting 2 on a forbidden write, wired in `git/SKILL.md` frontmatter with `matcher: "Bash"`, mirroring the security hook's shape.
- `test/validate-git-hook.nu` (412 lines, 69 cases) — 31 scanner cases plus the issue's frozen 23-row BLOCKING and 11-row PASSING matrices and 4 gap-fill rows, registered in `mise.toml`.
- Corrected a false claim shipped in two files: `security/references/gitleaks-usage.md:81` and `git/SKILL.md` both said skill hooks are "scoped to their own skill's lifecycle". The documentation says registration persists for the rest of the session once the skill is invoked, so the real gap is a session that never invokes the skill at all.

## Why a tokenizer rather than string matching

Two independent reasons, both measured rather than assumed.

Substring matching blocks legitimate work: `git grep -n -- "gh pr merge --admin"` carries the forbidden string inside its own `tool_input.command`. An analyst investigating this issue tripped that four times in one session. Position 0 has to be `gh`, not `git`.

Position-0 matching alone is already a defeated design in this repo. `gitleaks-usage.md:81` records `check-secrets-before-commit.sh`'s `^git[[:space:]]+commit` anchor falling to a compound command, a `cd …&&` prefix, and an env-var prefix (claude-skills-271) — and notes the compound form matters most because this repo's own workflow prescribes it. Hence segmentation. The same file records two attempts to widen that hook by string-matching being reverted because "a command string does not carry the structure to say whose flag a `-f` belongs to"; a tokenizer supplies exactly that structure.

No new runtime dependency. Measured: no nushell 0.113.1 builtin is quote-aware, and no mise-installable CLI emits a token list — `shfmt` parses internally but exposes only write/diff/list/find. `topiary visualise --language bash --format json` does emit a real tree-sitter-bash parse tree, but as row/column spans rather than dequoted values, and it is not in the pinned toolchain.

## Named approximations and accepted limits

Stated in `/core:git` rather than left to be discovered. Two approximations: a `gh pr merge` lacking `--match-head-commit` stands in for the undecidable merge-queue case, and `--method DELETE` against `/issues/comments/` stands in for Gate 3 record deletion. Over-blocks include every `gh api graphql` read loaded from a file or variable, since graphql is always POST. Under-blocks include brace expansion, wrapper prefixes such as `sudo` or an absolute path, and flags supplied through an expansion.

Rule 3 matches `/issues/comments/`, the deletion shape — not `/issues/<n>/comments`, which is how a Gate 3 record is posted. `gh api repos/o/r/issues/286/comments -F body=@review.md` is a passing test row, because blocking it would break the gate this hook exists to protect.

## Gate 1

`mise run ci` exit 0 (100.60s), `mise run test:version-bumps origin/main` exit 0 (`core` 0.2.82, `all-skills` 0.4.173), `nu test/validate-skills-quality.nu` exit 0 with `dupe/28ccd32d` exactly 7 and `quality-baseline.json` unmodified, `mise run test:git-hook` 69/69. Run by a separate agent, not self-reported. gitleaks clean. `git/SKILL.md` is 384 lines against the 500 cap.

The acceptance criteria went through three adversarial review rounds before any test was written, and the frozen tests through two before any implementation started.


vinnie357:
Gate 3: REJECT 4a870a1bca1ed84d438ad10e69a266ba428a3cf6

Adversarial review of PR #288 at `4a870a1bca1ed84d438ad10e69a266ba428a3cf6`. Brief: defeat the hook — construct a forbidden merge-path write that exits 0. Every payload below was built as data and fed to `block-forbidden-merge-writes.nu` on stdin from a shallow clone; nothing was executed. `gh` facts were checked with read-only calls against gh 2.93.0 and the cli/cli source, never with a real merge or dismissal.

## Gate 1, independently (shallow clone at the head sha)

- `mise run ci` exit 0; `mise run test:version-bumps origin/main` exit 0 (`core` 0.2.81→0.2.82, `all-skills` 0.4.172→0.4.173); `nu test/validate-skills-quality.nu` exit 0 with `dupe/28ccd32d:duplicate_block — 7 shared window(s)`; `mise run test:git-hook` exit 0.
- `git diff 7603b01 4a870a1 -- test/validate-git-hook.nu` is empty; the implementation commit touches no test file and no `mise.toml` (`--stat`: 7 files, none under `test/`).
- The /core:git attribution grep (`git log origin/main..HEAD --format=%B` piped through the skill's trailer pattern) matches nothing on the four branch commits.
- Both hook files are mode 100755 in the index; `source posix-scan.nu` resolves relative to the script (probes ran from a different cwd and worked).

## Findings — bypasses that exit 0 (none on the disclosed under-block list)

**F1 — CRITICAL: `gh -R o/r pr merge 286 --admin` exits 0.** `segment-blocks` requires position 1 to be literally `pr`. gh accepts `-R`/`--repo` before the subcommand — verified read-only: `gh -R vinnie357/claude-skills pr list -L 1` returns a row. `gh --repo o/r pr merge 286 --auto` also exits 0. This is the most natural way an agent writes a merge from outside the repo directory. (`gh -R o/r api …` is NOT a bypass — gh rejects it with `unknown shorthand flag: 'R'`; only `pr`/`mr` subcommands take it.)

**F2 — HIGH: combined short flags hide the method: `gh api -iX PUT repos/o/r/pulls/1/reviews/9/dismissals -f message=x` exits 0.** `extract-method` only recognizes a token starting with `-X`; `-iX` starts with `-i`. pflag folds boolean shorthands, so gh reads this as `-i -X PUT` — verified read-only: `gh api -iX GET repos/vinnie357/claude-skills` returns `HTTP/2.0 200 OK` with headers. With the method invisible, `extract-path-token` then takes `PUT` as the path, so rule 3 never sees `/pulls/`. (`-iXPUT` joined is caught, exit 2.)

**F3 — HIGH: an empty or duplicated `--match-head-commit` passes as pinned.** `gh pr merge 286 --squash --match-head-commit=""`, `… --match-head-commit ""`, and `… --match-head-commit <sha> --match-head-commit=` all exit 0. `carries-flag` tests presence only. gh's `pkg/cmd/pr/merge/http.go:77` is `if payload.expectedHeadOid != "" { … }` — an empty value is an unpinned merge, and pflag last-wins (verified read-only: `gh pr list -L 2 -L 1` returns 1 row). Same defect on glab: `glab mr merge 12 --squash --yes --auto-merge=false --auto-merge=true` exits 0 (the hook finds the `=false` token; pflag applies the later `=true`).

**F4 — HIGH: backslash-newline continuation glues the next flag into the previous token.** `gh pr merge 286 --squash --match-head-commit <sha> \`⏎`--admin` exits 0: the scanner treats `\`+newline as an escaped newline character, producing one token `<sha>\n--admin`; bash removes `\<newline>` entirely and sees `--admin`. Same shape on `gh api \`⏎`-X PUT repos/…/dismissals -f message=x` (exit 0 — `-X` becomes `\n-X`, the method disappears, and `\n-X` is taken as the path). The indented variant (`\`⏎`  --admin`) is caught because the spaces end the token — so this depends on formatting, which is worse than a consistent miss. This is a scanner correctness bug against the POSIX rule it claims to model, not a rule gap.

**F5 — HIGH: shell grouping and reserved words at position 0.** All exit 0: `(gh pr merge 286 --admin)`, `{ gh pr merge 286 --admin; }`, `if true; then gh pr merge 286 --admin; fi`, `for p in 286; do gh pr merge $p --admin; done`, `! gh pr merge 286 --admin`, `time gh pr merge 286 --admin`, `cd /tmp && (gh pr merge 286 --admin)`, and a leading redirection `>/dev/null gh pr merge 286 --admin` / `2>/dev/null gh …`. These are not "wrappers" in the disclosed sense (`sudo`, `command`, `xargs`, an absolute path) — `(`, `{`, `!`, `then`, `do`, `else`, `elif`, `time`, and pre-command redirections are a finite, spec-defined set of POSIX grammar tokens, exactly the class a scanner that already special-cases `;`, `&&`, `||`, `|`, `&`, `#` is expected to own. Skipping them at position 0 does not reopen the denylist problem.

**F6 — MEDIUM: a gh invocation inside a command substitution is invisible.** `echo "$(gh pr merge 286 --admin)"` and ``x=`gh pr merge 286 --admin` `` exit 0. The scanner contract (Scan 30, "a nested substitution stays one token") freezes `$( … )` contents as opaque, so this is AC-frozen behavior — but it is not in `/core:git`'s under-block list, and the doc's "any shell-expansion-bearing token in a merge command blocks" reads as covering it. Either scan inside `$( )`/backticks or name it.

**F7 — LOW: alias two-step.** `gh alias set m "pr merge --admin"; gh m 286` exits 0 on both segments. Denylist-adjacent, so I would accept a disclosure rather than a rule.

## Findings — false positives that would get the hook disabled

**F8 — MEDIUM: `gh pr merge --help` exits 2, and so does `glab mr merge --help`.** `/core:git` itself cites `gh pr merge --help` (gh 2.93.0) as the source for the merge-queue statement. A reviewer or implementer following the skill's own instruction hits the block. Not on the disclosed over-block list.

**F9 — MEDIUM: a heredoc whose body contains an example line blocks.** `cat > notes.md <<'EOF'`⏎`gh pr merge 286 --admin`⏎`EOF` exits 2 — newline splits segments and the example line lands at position 0. Writing docs about this very hook through the Bash tool is a realistic shape in this repo. Pass 9 covers the quoted `git commit -m "…"` form only.

## Findings — fail-open and latency

**F10 — HIGH: scan time is quadratic in command length; the 120 s timeout converts a large command into a silent disable.** Measured on the hook as shipped: 2 KB → 0.17 s, 10 KB → 2.7 s, 39 KB → 40.7 s, one 20 KB token → 10.9 s (baseline `ls -la` 0.05 s). Every Bash call pays this. A ~70 KB heredoc (not exotic for a file write) exceeds the timeout, which the harness treats as non-blocking — the hook is off exactly when the command is big enough to hide something, and the tool call stalls for two minutes first. `/claude-code:claude-hooks` sets the PreToolUse target at under 1 s. The per-char `$cur_value + $c` and per-token `append` copies are the cost; a bounded scan or a linear rewrite is required.

**F11 — LOW: malformed stdin exits 1 with a nushell stack trace, not 0.** `printf 'not json' | nu hook` reaches line 248 with `$payload` as a string (the `try { from json } catch { null }` guard does not produce null here) and dies with `only_supports_this_input_type`; a non-string `tool_input.command` exits 1 the same way. Both are non-blocking, so this is not a fail-closed risk, but the transcript shows a `hook error` notice with a stack trace, and the `if $payload == null { exit 0 }` branch is dead for that input. Guard with `describe` or `try` around the `get`.

## Findings — PR body and issue claims

**F12 — MEDIUM (anti-fabrication): the shfmt claim is false.** PR body and claude-skills-340's TOKENIZER note both state "`shfmt` parses internally but exposes only write/diff/list/find" and "no mise-installable CLI emits a token list". `shfmt` is in the mise registry (`aqua:mvdan/sh`), and `shfmt --to-json` (v3.14.0, checked via `mise exec shfmt@3.14.0`) emits a typed AST with dequoted `Lit`/`SglQuoted`/`DblQuoted` values, `CmdSubst`, `Subshell`, and correct `\<newline>` handling — on the PR's own split-quote fixture it yields `Lit "repos/o/r/pu"`, `Lit "lls/1/reviews/9/dismissals"`, `SglQuoted "$HOME"`; on F4/F5 input it yields `--admin` as its own `Lit` and a `Subshell` node. The tokenizer decision was "settled" on a measurement that was wrong. I am not asking for a rewrite in this PR — the operator decided no new dependency — but the false measurement must be corrected in the PR body and the issue, and the decision re-stated on the true basis (dependency policy, not tool absence). Restraint verdict on the 226-line scanner is therefore "justified only by policy, not by necessity" — and F4/F5 are exactly the grammar a real parser already gets right.

**Verified true:** posix-scan.nu 226 lines, hook 269, test 412; 31+23+11+4 = 69 cases; `dupe/28ccd32d` = 7; `git/SKILL.md` 384 lines; `gitleaks-usage.md:81` carries `^git[[:space:]]+commit` and the claude-skills-271 citation; the lifetime quote matches the current docs verbatim ("keeps running them for the rest of the session, on turns after the skill's own turn as well"); exit-code contract (2 blocks, 1/timeout non-blocking) matches the docs. **Could not verify:** that the hook fires in a live session under the plugin cache and that `/usr/bin/env nu` resolves in the harness's environment — the PR body does not claim a live-session firing, and `/claude-code:claude-hooks` says not to assume an event fires without verifying; "run by a separate agent" and the 100.60 s figure are unverifiable. **Follow-up, outside this PR's files:** `/claude-code:claude-hooks` still describes skill-frontmatter hooks as "Lifecycle-scoped", which this PR's correction now contradicts.

## What flips this to APPROVE

1. Fix or explicitly disclose-with-a-pinning-test each of F1–F5 (F1, F2, F3 are real writes gh accepts today; F4 and F5 are scanner grammar). Minimum fixes: skip `-R <v>`/`--repo <v>`/`--repo=<v>` before subcommand detection; treat any `-[a-zA-Z]*X` short cluster as carrying `-X`; take the LAST `--match-head-commit`/`--auto-merge` occurrence and require a non-empty value; delete `\<newline>` in the scanner; skip `(`, `{`, `!`, `time`, `then`, `do`, `else`, `elif`, and `[0-9]*[<>]…` redirections at position 0.
2. Name F6 and F7 in the under-block list, or scan inside `$( )`.
3. Allow `--help` on the merge shapes (F8), and add F9 to the over-block list or exempt heredoc bodies.
4. Bound or linearize the scan (F10) — a 10 KB command must not cost 2.7 s.
5. Correct the shfmt claim in the PR body and issue (F12); F11 is optional.

Then post the delta and I will re-attest at the new sha after reading `git diff 4a870a1..<head>` only.

vinnie357:
Superseded by a re-scope. Marking this draft rather than merging or closing it — the branch holds work worth reading, but it is no longer a merge candidate in this shape.

The Gate 3 review above was correct to reject, and a follow-up analysis found the deeper cause sits upstream of every finding: the acceptance criteria carried a block marked `TOKENIZER — settled, do not re-litigate`, resting on a measurement asserted as `Measured 2026-09-04` that was false — `shfmt --to-json` is mise-installable and does exactly what the hand-written scanner was built to do. PR #286's round-17 reviewer had already raised rung 3 against hand-writing a tokenizer. The `settled` marker closed the question, and reviewers obeyed it even after F12 falsified its basis.

Two things follow, in order.

1. The gates and reviewer process changes first: Gate 3's brief puts fit-and-size before defeat-the-change and the record carries each under its own heading; an uncited `settled` decision is reopened at plan review rather than obeyed; the executable `reviewPrompt` in `forge-issue.workflow.js` gains the restraint line it never had; and every PreToolUse hook test carries a size-scaling row, since `/claude-code:claude-hooks` already states a one-second budget that nothing enforces.

2. claude-skills-340 is then rebuilt small against a re-scoped contract. The operator's framing: this is a cheap deterministic gate, like a linter or a compiler — it catches the everyday mistake and is not judged on whether a determined agent could evade it. Concretely that deletes the LIVE-flag and fail-closed-unreadable machinery, promotes `gh -R owner/repo pr merge` and backslash-newline continuation to must-block because both are ordinary typing, forces `gh pr merge --help` to pass, adds an enforced latency budget, and moves the evasion classes from under-blocks-we-regret to out-of-scope-by-design.

Measured on this branch, for the record: the scanner is quadratic — 0.08s at 1KB, 0.36s at 4KB, 1.21s at 8KB, 4.65s at 16KB — which disqualifies it as a cheap gate independently of any bypass.

Findings F1, F2 and F3 survive the re-scope unchanged: `gh -R o/r pr merge --admin`, `-iX PUT`, and an emptied `--match-head-commit=` are how an agent naturally types the command, not evasion. F4 through F7 are reclassified as disclosures rather than defects.

vinnie357:
Superseded by #291, merged as `cf49e93`. Closing unmerged.

This branch's design was rejected at Gate 3 and the issue was then re-scoped from a security boundary to a cheap deterministic gate. #291 ships that: 351 + 216 lines against this branch's 226 + 269, with the bypass question closed by design rather than by enumeration, four enforced budget rows, and the four instruction sites in our own `github` plugin pinned so the gate does not fire on the repository's own operator-approved workflow.

What was learned here and carried across, since the work was not wasted:

- The seven bypasses this branch's Gate 3 found were the evidence that a boundary was unreachable. The re-scope followed directly from them.
- The quadratic scanner measured here (0.08s at 1KB rising to 4.65s at 16KB) became the first of four budget dimensions in #291, and the reason "cheap" is now a number with a test rather than an aspiration.
- The `settled, do not re-litigate` marker in this issue's acceptance criteria, resting on a measurement that turned out false, is what produced the Gate 3 process change in #289 — an uncited settled decision is now reopened at plan review rather than obeyed.

The branch is left in place for reference and can be deleted at any time.

PR #291, opened 2026-09-05
feat(git): add a cheap PreToolUse gate for forbidden merge-path writes

vinnie357:
Closes claude-skills-340. Supersedes #288, which Gate 3 rejected and which is now draft.

`/core:git` forbids `gh pr merge --auto`, `--admin`, an unpinned merge, `glab mr merge` without `--auto-merge=false`, review dismissal, and Gate 3 comment deletion — and enforced none of them. This is the tool-boundary backstop.

## What this is, and what it is not

A **cheap deterministic gate** — a linter, not a security boundary. It catches the everyday mistake and is not judged on whether a determined agent could evade it. Evasion classes (`$( )`, aliases, `bash -c`, subshells, `!`/`time` prefixes, an emptied `--match-head-commit=`) are named as out of scope by design in `/core:git`, not left as silent gaps.

That framing is why this PR exists rather than #288. The earlier attempt specified a boundary and produced 226 + 269 lines still losing an arms race — its Gate 3 found seven bypasses in one pass. Re-scoped, the same objective is 351 + 216 lines with the bypass question closed by design instead of by enumeration.

## Shipped

- `scan-merge-writes.nu` (351) — `tokenize-command` splits on unquoted `;`/`&&`/`||`/`|`/`&`/newline, dequotes per the three POSIX mechanisms, concatenates adjacent quoted segments, treats backslash-newline as continuation, and returns completed tokens on an unterminated quote.
- `block-merge-writes.nu` (216) — PreToolUse hook, exit 2 blocks. Wired in `git/SKILL.md` frontmatter, not `plugins/core/hooks/hooks.json`, so installing the plugin does not hand every session an unremovable Bash interceptor.
- `test/validate-git-hook.nu` (617, 54 checks) — 12 scanner units, the frozen 15 BLOCKING and 15 PASSING rows, 3 robustness cases, and 6 regression rows from Gate 3.
- **Four of our own instruction sites pinned.** `pr-review/SKILL.md:101`, `pr-review/references/merge-and-closeout.md:17`, `dependabot-consolidator/SKILL.md:83` and `dependabot-consolidator/references/merge-and-closeout.md:16` all told an agent to run `gh pr merge <n> --squash` unpinned — which this hook blocks. A gate that fires on the repository's own operator-approved workflow gets switched off, so they now pin `--match-head-commit <headRefOid>`. Pinning a merge to the reviewed sha is correct regardless of this hook.
- **Two exit-code errors corrected** in `claude-hooks/SKILL.md` (lines 98 and 102), which said non-zero blocks. Exit 2 is the only blocking code; exit 1, a missing script, a non-executable script and a timeout are all non-blocking — precisely what this hook's contract rests on.

## Performance is enforced, and calibrated on the slow environment

The hook runs on every Bash call, so "cheap" is a number with a test. Four budget rows, each scaling a **different dimension**, all min-of-three with the hook as a subprocess under 800ms:

| dimension | local | runner |
|---|---|---|
| payload size, 64KB one token | 77ms | 74.7ms |
| token count, ~21k tokens | 115ms | 116.9–218.9ms |
| segment count, ~4k segments | 128ms | 191.9ms |
| double-quoted span count, ~6k spans | 185ms | 276.4ms |

Three separate quadratics were found and fixed along the way — per character, per token, and per segment — all the same nushell trap, where `append` inside a loop copies the accumulator. The AC now bans it at any granularity and records all three measurements.

Two budgets were set from local timings and **both failed remote CI**: 200ms (token row measured 223.8ms) and 500ms (segment row measured 630.7ms). The runner is 2.4×–3.5× slower depending on the row and varies ~13% run to run, so the AC now requires a runner measurement for the row a budget governs. 800ms sits under the one-second PreToolUse figure `/claude-code:claude-hooks` documents, with 2.9× headroom on the worst row.

## Named limits, stated in `/core:git`

Two approximations: a `gh pr merge` lacking `--match-head-commit` stands in for the undecidable merge-queue case, and `--method DELETE` against `/issues/comments/` stands in for Gate 3 record deletion. Over-blocks: every `gh api graphql` carrying the substring `mutation`, every non-GET `gh api` touching `/pulls/`, and a heredoc line beginning `gh pr merge --admin`. Rule 3 matches `/issues/comments/` — the deletion shape — not `/issues/<n>/comments`, which is how a Gate 3 record is posted; that call is a passing test row.

## Process

The criteria went through three plan-review rounds before a test was written, and the tests through one review before implementation started. Gate 3 then ran five rounds and found ten defects. The four it found in the shipped hook were each given a **red row first** — F1 as a passing row, F4 as a blocking row, since reversing either would freeze the defect as correct — and only then handed to a separate implementer.

Findings worth naming, because they are the ones a lighter process would have shipped:

- `gh -R o/r pr merge --admin` passed. `-R` before the subcommand is ordinary usage.
- A pinned merge with a multi-line `--body` was **blocked** — a backslash-newline inside double quotes read as an unterminated quote, dropping every later token including the pin.
- A comment in the scanner described a `drop 1` that was not in the code. Implementing the comment would have dropped the last segment of any command without a trailing separator, and `cd x; gh pr merge --admin` would have passed.
- The plan review caught that my own prescribed performance fix was itself quadratic.

Two figures quoted in earlier revisions of this description are corrected here: the 82/330/14,036ms three-implementation comparison came from a test reviewer's scratchpad that was not retained, so it is reported as unreproducible rather than as evidence; and #288's suite was 69 *checks* (31 scanner plus 38 matrix rows), not 69 matrix rows.

One limit is documented rather than chased: the single-quoted-span series is non-monotonic (701ms at 64KB, 371ms at 96KB, reproduced twice) and is recorded in the issue as requires-investigation.

## Gate 1

`mise run ci` exit 0, `test:version-bumps` exit 0 (`core` 0.2.83, `claude-code` 0.2.88, `github` 0.3.15, `all-skills` 0.4.174), `dupe/28ccd32d` exactly 7 with `quality-baseline.json` unmodified, `test:git-hook` 54/54. gitleaks clean. `git/SKILL.md` 402 lines against the 500 cap. Zero literal NUL bytes in either hook file.


vinnie357:
Gate 3: REJECT cd5bcdc325b24fefefc42a0c5db3e7acd91b89f5

Reviewed at `b93a09f`; the head moved to `cd5bcdc` (two operator commits, `6d34280` bees + `cd5bcdc` test) before this record was posted, so the delta was read in full and Gate 1 was re-run on the new head. The marker is at the sha that is live now.

## Fit and size

**Problem, from claude-skills-340:** `/core:git` states forbidden merge-path writes and enforces none of them; give them a cheap PreToolUse backstop at the Bash tool boundary.

**Who is on the other side:** the careless class — a first-party agent that would comply if it remembered. Settled by the issue's first CRITICAL bullet: "It catches the everyday mistake. It does not stop an agent that is trying to get around it, and it is not judged on whether it could." Evasion is out of scope by design; this review hunted natural spellings only, and no finding below is an evasion.

**Smallest change that satisfies the AC:** what shipped, near enough. Scanner, hook, frozen test, `hooks:` frontmatter, the `/core:git` section and the `mise.toml` registration are all named deliverables. The four `github`-plugin pins and the two `claude-hooks` lines are not scope that rode along — the issue lists them under "Pin the repo's own merge instructions first" ("Not optional") and "Deliverables" ("carries the same error — so fix both"). The one piece the AC does not dictate is `tokenize-fast`: it is load-bearing for latency row #15, and it is also the reason for F3 below. Diff is 863/20 across 16 files at `b93a09f` (the brief's "13 files" undercounts `.bees/issues.jsonl`, two `plugin.json`s and `marketplace.json`).

**Settled decision reproduced:** #288's quadratic scanner. Fetched `4a870a1` (#288 head) and piped `{"tool_name":"Bash","tool_input":{"command":<payload>}}` of space-separated words: **92ms at 1KB, 533ms at 4KB, 1.90s at 8KB**, min-of-3. The 4→8KB ratio is 3.6×, superlinear as the issue states; the absolutes are higher than the issue's 0.08/0.36/1.21s on my fixture. #288 sizes: scanner 226 lines, hook 269, test 412 — the "69 test cases" figure did not reproduce from the file (38 `{label:` rows across its four case lists; the other 31 are unverified).

**Gate 1, independently, scratchpad shallow clone, judged by exit code.** At `b93a09f`: `mise run ci` exit 0 (97.49s); `mise run test:version-bumps origin/main` exit 0 (core 0.2.82→0.2.83, claude-code 0.2.87→0.2.88, github 0.3.14→0.3.15, all-skills 0.4.173→0.4.174); `nu test/validate-skills-quality.nu` exit 0 with `dupe/28ccd32d` at exactly 7 and `test/quality-baseline.json` unmodified; `mise run test:git-hook` exit 0, 48 ✅ / 0 ❌, latency rows 47–52ms and 65–69ms across three runs (body: 49.4 / 67.5). At `cd5bcdc`: `mise run ci` re-run, result stated in the last line of this comment.

**Other body claims verified:** 143 / 203 / 436 lines exact; `git/SKILL.md` 402 lines; five commits above merge-base `25bb676` at `b93a09f`, seven at `cd5bcdc`; attribution grep exit 1 on all seven and trailer blocks eyeballed clean; both hook files mode 100755; the hook exits 2 when executed by path from `/tmp` and from `/`, so the relative `use scan-merge-writes.nu` resolves against the script, not the cwd. The four pins are correct, not just present: `pr-review/SKILL.md:20` already collects `headRefOid` in its `gh pr view --json` read, so "collected above" resolves; the dependabot sites add the `-q .headRefOid` fetch inline. `claude-hooks` 98/102 are right and no other `exit 1`/"non-zero blocks" claim remains in that skill's references or templates. **Frozen-test check:** `git diff 642c5bb b93a09f -- test/validate-git-hook.nu mise.toml` was empty; at `cd5bcdc` it is not — see F5.

## Findings

**F1 — REJECT-driving. A double-quoted argument containing a backslash-newline is treated as an unterminated quote, dropping every token after it. Fails closed on a valid, spec-covered command.**

```
gh pr merge 286 --squash --body "line one \
line two" --match-head-commit abc
```
exits **2**. `tokenize-command` returns `[["gh","pr","merge","286","--squash","--body"]]`; bash reads the same string as one token `line one line two` (checked with `printf "[%s]"`). The frozen interface in the test header states both rules this breaks: "a backslash immediately before a newline is a LINE CONTINUATION: both characters are removed" and, inside double quotes, "a backslash still escapes the next character". Cause: the `dq` alternative is `"(?:\\.|[^"\\])*"` and Rust-regex `.` excludes newline, so the alternation fails at the backslash-newline, the `bad` branch fires on the opening `"`, and unterminated-quote recovery discards the rest of the segment — including the pin. The same string with a plain newline inside the quotes passes. The issue ranks a false positive as the worse defect, and this one lands on the `--body` of the very command the hook exists to gate. The frozen suite does not cover it.

Verified fix, 3 lines in `scan-merge-writes.nu`; with it the frozen suite is 48/48, the command above tokenizes to `[..., "--body","line one line two","--match-head-commit","abc"]`, and `"a\"b\\c" 'd\e' f\ g "it's" "x"y'z'` still matches bash token for token:

```diff
-    let pattern = r#'(?<cont>\\\n)|(?<sq>'[^']*')|(?<dq>"(?:\\.|[^"\\])*")|(?<esc>\\.)|(?<bad>['"])|(?<sep2>&&|\|\|)|(?<sep1>[;|&\n])|(?<ws>[ \t]+)|(?<plain>[^ \t\n;&|\\'"]+)'#
+    let pattern = r#'(?<cont>\\\n)|(?<sq>'[^']*')|(?<dq>"(?:\\[\s\S]|[^"\\])*")|(?<esc>\\.)|(?<bad>['"])|(?<sep2>&&|\|\|)|(?<sep1>[;|&\n])|(?<ws>[ \t]+)|(?<plain>[^ \t\n;&|\\'"]+)'#
@@
-                | parse --regex '(?<esc>\\.)|(?<plain>[^\\]+)'
-                | each {|q| if $q.esc != null { $q.esc | str substring 1.. } else { $q.plain } }
+                | parse --regex '(?<cont>\\\n)|(?<esc>\\[\s\S])|(?<plain>[^\\]+)'
+                | each {|q| if $q.cont != null { "" } else if $q.esc != null { $q.esc | str substring 1.. } else { $q.plain } }
```

Add the command above as a PASSING row — a one-dimension variant of the existing continuation row.

**F2 — AC deviation. A non-string `tool_input.command` exits 1 with an 11-line nushell traceback.** `{"tool_name":"Bash","tool_input":{"command":123}}`, `["gh","pr","merge"]` and `{"a":1}` each print `Error: nu::shell::cant_convert` and exit 1. The Robustness AC says malformed stdin "exits 0 without a stack trace". The harness never sends this shape and exit 1 is non-blocking, so nothing is blocked in practice — but `main`'s `try` does not cover it: the failure is the `string` parameter type on `command-blocked`, outside the `try`. `null`, absent `tool_input`, a string `tool_input`, top-level array/string, truncated JSON, whitespace, and a NUL byte all behave (exit 0, or exit 2 with the forbidden pattern intact). Verified fix, 1 line in `block-merge-writes.nu`, plus one robustness row:

```diff
-            command: ($payload | get -o tool_input | get -o command | default "")
+            command: ($payload | get -o tool_input | get -o command | default "" | if ($in | describe) == "string" { $in } else { "" })
```

**F3 — The scanner header makes a false claim, and the budget rows never exercise the slow path.** `scan-merge-writes.nu` says the slow path "is only ever asked to walk the small number of pieces a quoted/escaped argument actually produces". It is not: one quote or backslash anywhere routes the *entire* command through `classify-pieces`. Row #15's ~21,000-token fixture passes at 68ms locally only because it contains no quote; the same fixture plus a trailing `'x'` measures **663ms** min-of-3 locally — and the delta's own AC text puts the runner at 3.3× local, so roughly 2.2s there against the new 500ms budget. Other 64KB shapes, all exit 0, min-of-3, hook as subprocess: 32k `;`-segments 1.22s (fast path), 21k `&&`-segments 938ms (fast path), 5k newline-separated `gh pr list` lines 640ms (1.62s with one quote), 13k adjacent `a"b"` tokens 3.17s. All linear — 16/32/64KB give 202/360/663ms, 846/1659/3170ms, 205/350/640ms — so nothing quadratic; the constants are 3–16× the rows'. Everyday cost is fine: 45–48ms for ordinary commands against a 25ms `nu` startup.

The delta sharpens this. The new AC bullet says 500ms "still catches every bad shape — at the same 3.3x the piece-level `each` implementation lands near 1090ms". The shipped slow path *is* a piece-level `each`, and the budget row cannot see it because `tokenize-fast` routes the fixture around it. Disposition: correct the header sentence in this PR (a claim in code, and wrong), and either add a third latency row — row #15's fixture plus one quote — and make it pass, or file that row with these numbers as the follow-up that owns the slow-path constant. None of these shapes is a command an ordinary agent types and the 10s hook timeout bounds the stall, so on its own this is not the blocker; the false sentence is.

**F4 — Under-block on a joined shorthand, low.** `gh -Ro/r pr merge 286 --admin` exits 0: the global-flag skip matches `-R`, `--repo`, `--repo=`, `--hostname`, `--hostname=` but not pflag's joined `-Rowner/repo`, so `-Ro/r` becomes the subcommand and no rule fires. 36 other natural spellings all blocked as expected — `--repo=o/r` after the subcommand, `-X put`, `--method=DELETE`, a leading-slash path, path-before-flags, `-H` before `-X`, `--input` with no method, `GH_REPO=o/r` prefix, `2>&1 | tee`, `--auto-merge false` and `--auto-merge=true` on glab, tab-separated tokens, CRLF, `&` backgrounding. File it or add a `-R` prefix branch; not required here.

**F5 — The delta changed the frozen test's expectation and left three stale references behind.** `cd5bcdc` raises both assertions to 500ms, so `git diff 642c5bb cd5bcdc -- test/validate-git-hook.nu` is no longer empty. The provenance is clean — the operator changed the AC in bees first (`6d34280`), then the test, which is the escalation path the adversarial-TDD rule prescribes, not an implementer retrofit. But the same file still says 200ms two lines above the edit: `test/validate-git-hook.nu:355` ("Both asserted under 200ms") and `:358` (the printed header "min-of-3 < 200ms"), and so does `scan-merge-writes.nu:22` ("well inside the 200ms budget") and the PR body ("against a 200ms budget"). Fix all four. The PR body should also say the budget moved and why, in the sentence that currently reports the two rows.

**Unverified in the body:** the "82ms, 330ms and 14,036ms" three-implementation measurement has no artifact — not in this PR's comments (0 before this one), not in the single bees comment on 340, not on #288. The delta's AC bullet now cites the derived runner figures (1090ms, 46s), which places the numbers in the record but does not make them reproducible: the implementations are not in the PR. Either point at where they live or say so.

**Nothing else found.** 51 over-block probes of everyday shapes all exit 0 — `--help`/`-h`, `gh help pr merge`, `--match-head-commit=`, `"$OID"`, `$( )` pins quoted and unquoted, `OID=$(...); gh pr merge ...`, continuation before and after the pin, every `gh api` GET on `/pulls/`, `-F body=@review.md` and `--input` on `/issues/<n>/comments`, a GET of `/issues/comments/<id>`, `gh api graphql` queries, `gh pr comment --body-file`, a heredoc whose lines mention `gh api -X DELETE .../issues/comments/1` mid-line, PATCH/POST/DELETE on issues and labels, `--subject`, `$'...'`, an apostrophe in a trailing comment, `-R` after the pin. Fast and slow paths agree token-for-token on 11 boundary inputs (`2>&1`, tabs, `|&`, `;;`, `&&&`, `$( )`, `#`, unicode, blank lines). The repo-wide sweep found no fifth instruction site the hook would block; `/core:git:357` and `:385` show the bare `gh pr merge --squash` in reference/rule position with an explicit pointer to the pinned form. Restraint: nothing beyond the AC. Documentation obligations are met; `/core:git:72`'s "no enforcement hook at all" is about attribution and stays true.

**Re-attestation:** apply F1 and F2 (4 lines plus two test rows), fix F3's sentence and F5's four references, decide F3's third row, re-run `mise run test:git-hook` and `mise run ci`, and I will read `git diff cd5bcdc..<new-head>` only and post the marker at the new sha.

Gate 1 at `cd5bcdc`: `mise run ci` exit 0 (wall ); inside it `test:git-hook` 48 ✅ / 0 ❌ with the rows at 67.4ms and 89.5ms under parallel `mise` load, and its printed header still reads "< 200ms" — F5 as observed in CI output.

vinnie357:
Gate 3: REJECT ad2ad2c50772e369525363bd717146913d443ec4

Re-attestation. Read bounded to `git diff cd5bcdc..ad2ad2c` (3 files: hook +15/−, scanner +195/−, test +86/−), plus a fresh Gate 1 and a fresh attack on the new router. The delta addresses every finding from the first record, and the process — a red row per finding verified against the shipped hook as a black box, then a separate implementer, `git diff 3ecb8d1 ad2ad2c -- test/validate-git-hook.nu mise.toml` empty — is the right one. One new defect in the F3 fix is the reason for the marker; it is a one-line change with a verified patch below.

## Fit and size

Same problem, same careless class, same sizing. The delta is sized to the findings: F1 is the `\\\n` alternative in the `dq` group plus a shared `unescape-dq-inner`; F2 is a `describe` guard; F4 is one `-R` prefix branch; F5 removed the stale 200ms from the scanner header and both test-file sites (the one remaining "200ms" is the deliberate contrast, fine). F3 is the substantial piece — a three-way router — and it is the minimum that makes the quoted-bulk row pass without a per-piece walk, so it is in scope, not gold-plating.

**Gate 1 at `ad2ad2c`, scratchpad shallow clone, by exit code:** `mise run test:git-hook` exit 0, **52 ✅ / 0 ❌** (the message said 51; it is 52 — 48 prior plus the four regression rows); `mise run test:version-bumps origin/main` exit 0; `mise run ci`: exit 0 (89.04s), with `test:git-hook` inside it at 60/84/90ms for the three budget rows under parallel `mise` load. Nine commits above merge-base `25bb676`, attribution grep exit 1 on all nine.

**Claims in the delta checked:** the F1 row is a PASSING row and the F4 row a BLOCKING row, as stated; "verified: exit 0 / exit 2" reproduce; the quoted-bulk row measures 75–90ms here (message: 110ms); the header's "95ms → 1030ms, 10.8×" is the same defect I measured at 68ms → 663ms on my fixture — same shape, consistent.

## Findings

**F6 — REJECT-driving. The F3 fix made path 2 quadratic in segment count.** `tokenize-resolved` flattens all tokens, unprotects once, then rebuilds segments in a `for` loop with `$out | append [...]` and `$unprotected | skip $offset | first $n` — both copy on every iteration, so cost is O(segments²). The AC forbids exactly this shape ("do not append per token to a `mut` list either — measured 5.99s versus 6ms"); it is per *segment* here, same mechanism. Measured, min-of-3, hook as subprocess, `a;`×N plus one `'x'` token: **688ms at 8KB, 2.28s at 16KB, 8.3s at 32KB** — ×3.3 and ×3.6 per doubling. The everyday spelling of that shape is a long multi-line command that contains one quoted argument: 5,000 `gh pr list` lines plus `'x'` (64KB) measured **1.6s** on the old single slow path and measures **6.05s** now — the delta made it 3.7× worse — and at 32KB (~2,900 lines) 1.72s. The hook's frontmatter timeout is 10s; a 40k-segment command would stall every Bash call for 10s and then fail open. The new quoted-bulk row cannot see this: its fixture is one segment.

Verified fix, one line — restore per segment inside the `each` that `tokenize-fast` already pays for, which keeps `unprotect-list` a single native call per segment:

```diff
 def tokenize-resolved [command: string]: nothing -> list<list<string>> {
-    let segments = (tokenize-fast $command)
-    let seg_lens = ($segments | each {|s| $s | length })
-    let flat = ($segments | flatten)
-    let unprotected = (unprotect-list $flat)
-    mut out = []
-    mut offset = 0
-    for n in $seg_lens {
-        $out = ($out | append [($unprotected | skip $offset | first $n)])
-        $offset = $offset + $n
-    }
-    $out
+    tokenize-fast $command | each {|seg| unprotect-list $seg }
 }
```

With it: 240ms / 424ms / 785ms at 8/16/32KB (linear), the 5,000-line shape 738ms, the 32KB line shape 401ms, quoted-bulk row unchanged at 67–75ms, suite 52/52, all 115 probes below unchanged. The docstring above the function ("one vectorised call ... rather than one nu-level closure per token") stays true — it is one call per segment, the same granularity the function two lines up already uses. Add a budget row with many segments plus one quote so this shape has an observable; 32KB of `gh pr list` lines plus `'x'` under 500ms is a reasonable one (785ms → the `a;` shape is harsher than needed).

**Router: does the placeholder scheme corrupt a token, and does the NUL assumption survive a NUL?** Answered by construction, not by reading. Direct `tokenize-command` probes on the resolved path, compared with `bash -c 'printf "[%s]"'` where bash defines the answer: `'a b\tc;d&e|f'` → one token with all six specials intact; `"a\nb && c"` → one token; `a\;b\&c\|d\ e\<tab>f` → `a;b&c|d e\tf`; adjacent `a'b'"c"d e` → `abcd`, `e`; `"a\$b\\c"` → `a$b\c`; `'he said "hi'` inside sq → correct; CRLF inside dq preserved; unicode fine; `a;;b 'x'` gives the empty middle segment without error (`first 0` is fine); backslash at EOF survives as a literal token. A NUL sent as a genuine JSON `\u0000` (the only vector — bash argv cannot carry one, and my first attempt with raw bytes was invalid JSON the parser simply dropped): no crash, no truncation, no stderr, all seven decision probes correct on all three paths, including a NUL placed between `gh pr merge 286` and `--match-head-commit` (exit 0 — truncation would have blocked). The one observable effect is that a pre-existing NUL+hex pair inside a quoted token is restored to the real character after splitting — `'a\u000020b'` → `a b`, `"a\u00003Bb"` → `a;b` — token content only, never a segment or token boundary, so it cannot fabricate `--admin`, `--auto`, or a separator. State that in the `protect-specials` docstring ("cannot itself carry a NUL byte" → "a payload that does carry one is decoded, not split"). No change needed.

**Two pre-existing POSIX deviations on the resolved path, low, no decision impact, note only:** an empty quoted word is dropped rather than kept as an empty token (`gh pr merge 286 '' --admin` → 5 tokens; still blocks — and `gh '' pr merge --admin` now blocks where it used to pass); and inside double quotes `\;` yields `;` where POSIX keeps the backslash (bash: `a\;b`). Neither can change a verdict; the slow path had the same second one.

**Everyday command that exits 2 — none found.** The 88 original probes now all match expectations (F1 and F4 rows flipped, everything else unchanged), plus 27 new quoted everyday shapes all exit 0: `--body ""`, `--body 'a;b'`, `--body "a && b | c"`, a literal tab in quotes, `--body "$(cat body.md)"`, the F1 shape, escaped quotes in a body, an apostrophe in a trailing `#` comment after the pin, `--subject "feat(x): don't break"`, a heredoc comment containing an apostrophe and the words `gh pr merge --admin` mid-line, `git commit -m "fix: don't drop pin" && gh pr merge ... --match-head-commit abc`, a single-quoted `-q '.headRefOid'` inside `$( )`, `"$(printf 'a\nb')"`, a `for` loop, `~/My\ Notes/body.md`, `"$(git log -1 --format='%H')"`, a plain multi-line body, a three-line continuation with a quoted `--subject`, `-f body="a; b && c | d"`, `--body "line1\nline2"`. Five quoted forbidden shapes still block (`"--admin"`, a quoted dismissals path with `-f message="a; b"`, `mutation` inside a multi-line double-quoted query, `--auto` after a quoted `&&` body, `glab mr merge` with a quoted `--message`).

**Timing across the rest of the shapes, `ad2ad2c` vs `cd5bcdc`:** 21k words + one quote 663ms → **71ms** (F3 fixed); 13k adjacent `a"b"` 3.17s → 1.56s; 16k `'a'` tokens 677 → 695ms; 32k unquoted `\a` 271 → 370ms; 64KB dq with 32k escapes 227 → 222ms; the no-quote shapes (32k `;`, 21k `&&`, 5k lines) 1.2s / 0.94s / 0.64s unchanged — those are the fast path's own per-segment cost, linear, and out of this delta's scope. Everyday commands with a quote: 47ms.

**Re-attestation:** apply F6 (the one-line diff), add the many-segments-plus-quote row, re-run `mise run test:git-hook` and `mise run ci`, and I will read `git diff ad2ad2c..<new-head>` only and post the marker at the new sha.

vinnie357:
Gate 3: REJECT 77d8e075ba8c99f5827acbefef66e48a9ee3614f

Re-attestation, third round. Read bounded to `git diff ad2ad2c..77d8e07` (scanner +73/−, test +41/−, bees AC). F6 is fixed as described and the sentinel split is sound — answers to both questions are below, with the construction that backs each. The marker is REJECT for one reason that is not in the delta's own report: **remote CI is red at this head, on the row this delta was made for.**

## Fit and size

Same problem, same careless class. The delta is sized to the finding: `tokenize-resolved` lost its `for`/`append`, `tokenize-fast` became three single-pass native calls, the AC now bans `append` in any loop and requires each budget row to scale a different dimension, and the segment row exists. All in scope; the audit that found the nested `each` was the right follow-through.

**Gate 1 at `77d8e07`, scratchpad shallow clone, by exit code:** `test:git-hook` exit 0, **53 ✅ / 0 ❌** (52 prior plus the segment row); `test:version-bumps origin/main` exit 0; `mise run ci`: exit 0 locally, with the segment row at min-of-3 235.569ms under parallel `mise` load — local green is not the observable the AC names, see F7. Twelve commits above merge-base `25bb676`, attribution grep exit 1 on all twelve. `git diff 297d808 77d8e07 -- test/validate-git-hook.nu mise.toml` is empty. `git grep '| append'` in the hook files hits one comment line only, as stated.

**Claims in the delta checked:** the frozen segment row is 237–242ms here (message: 237ms); `a;`×N plus one quote 119 / 185 / 316ms at 8k / 16k / 32k here (message: 122/173/268 at 2k/4k/8k — same shape, linear); the 5,000-line everyday shape 486ms (message: 176ms — mine includes `gh pr list` in every line, so 5,000 `decide-segment` calls; the shape in the message does not say what its lines were). The "byte-for-byte equivalence across 14 edge cases" claim I extended rather than took: old (`ad2ad2c`) versus new `tokenize-command` over a 135-input corpus — every command from my 115 probes plus 27 separator edge cases (empty, whitespace-only, leading/trailing/double `;`, all six separators, `&&&`, `|&`, CRLF, tab-led, quote-adjacent separators, continuation-then-separator) — **0 differ**.

## Findings

**F7 — REJECT-driving. The segment row fails on the GitHub runner: 630.7ms against its own `< 500ms`.** `gh pr checks 291` at `77d8e07` reports "Validate marketplace and plugins" **fail**; the run log's only ❌ is `Gate 3 regression: ~8,000 ;-separated segments + one quoted arg ... min-of-3 630.681879ms < 500ms`. The other rows pass there at 93 / 213 / 234ms. The AC amended at `6d34280` says the budget "is measured on the SLOW environment, not the fast one ... A budget that holds only on the author's laptop is not a gate; it is a flake that teaches people to rerun CI." 237ms here is 631ms there — a 2.66× ratio, in line with the 3.3× the AC recorded for the token row. The delta's report gives local figures only; remote CI is the observable the AC names, and it is red.

What remains per segment after this delta is one nu-level closure per segment in `tokenize-resolved` (`each {|seg| unprotect-list $seg }` — six `str replace` calls per segment) plus one `decide-segment` per segment. The first can go: `unprotect-list` accepts a `list<string>`, so unprotect the *flat* token list once and split on the sentinel afterwards — the sentinel `\u{0}SEG` carries no `\u{0}XX` hex pair, so it survives `unprotect-list` untouched. Verified prototype:

```diff
 def tokenize-resolved [command: string]: nothing -> list<list<string>> {
-    tokenize-fast $command | each {|seg| unprotect-list $seg }
+    let sentinel = "\u{0}SEG"
+    let sep_pattern = '(?:&&|\|\||;|\||&|\n)'
+    let marked = ($command | str replace --all --regex $sep_pattern $" ($sentinel) ")
+    let flat = ($marked | split row -r '[ \t]+' | where ($it | is-not-empty))
+    unprotect-list $flat | split list {|x| $x == $sentinel }
 }
```

(Factor the three marking lines out of `tokenize-fast` into a helper returning `flat` rather than duplicating them; the prototype duplicates for measurement only.) With it: frozen segment row **242 → 154ms**; `a;`×N plus quote 85 / 124 / 198ms at 8k / 16k / 32k; 32KB of lines plus quote 272 → 231ms; 5,000-line everyday shape 486 → 414ms; the token row unchanged at 90ms; suite 53/53; 115 probes 0 mismatches; **135/135 equivalent** to `ad2ad2c` on the corpus. At the runner's 2.66× that is roughly 410ms for the segment row — under 500 but with under 20% headroom, which is thin for a min-of-3 on a shared 2-vCPU box. The remaining cost is `decide-segment` × 8,000, which is the hook's job and not the scanner's; if 410ms proves flaky on the runner, the honest options are a cheaper `decide-segment` early-out (it already returns on the first non-`gh` token, so the per-call floor is nu closure overhead) or a fixture that scales segments to 4,000 rather than 8,000 with the budget derived on the runner, as the AC's own rule requires. Do not raise the budget again without a runner measurement in the commit message.

**F8 — `tokenize-fast`'s new docstring describes a `drop 1` that is not in the code.** "`split list` always leaves a trailing empty group after the final sentinel; `drop 1` removes it unconditionally, which is safe because a sentinel always separates the marked string into (segment count + 1) pieces." There is no `drop 1`, and the premise is wrong: a command with no separator has no sentinel and yields exactly one group. Actual behaviour (verified, identical to `ad2ad2c`): `a;` → `[["a"],[]]`, `;a` → `[[],["a"]]`, `a;b` → `[["a"],["b"]]`. The hazard is the direction an implementer would "fix" it: adding the `drop 1` the comment promises would drop the *last real segment* whenever the command has no trailing separator — `cd x; gh pr merge --admin` would tokenize to `[["cd","x"]]` and pass. Delete the sentence. Same class as F3's false header claim; a comment that describes a nonexistent guard is worse than no comment.

**Q1 — Is the NUL-prefixed sentinel sound, and does `split list` on equality handle a segment that legitimately contains a sentinel-shaped token?** Sound for anything bash can execute; the equality test is the right one; and a payload that carries the sentinel is split at it, in the fail-closed direction. By construction: a token *containing* the sentinel (`a\u0000SEGb`) is not split — equality, not substring — and survives as one token; a lowercase `\u0000seg` is an ordinary token; the sentinel cannot collide with `protect-specials`' placeholders because those are `\u{0}` plus two hex digits and `SE` is not hex; the six-separator regex, `[ \t]+` split, and `where is-not-empty` are unchanged from the old fast path, which is why the 135-input corpus is identical. The one input that *is* affected is a token exactly equal to `\u0000SEG` delivered as a JSON escape — unquoted or inside single quotes (path 2 protects only the six specials, so the sentinel passes through `resolve-wellformed` intact) — and it becomes a segment boundary: `gh pr merge 286 --squash \u0000SEG --match-head-commit abc` exits 2 (the pin lands in a second segment). Bash cannot receive that command — argv strings end at the NUL — so no real merge is affected either way, and every forbidden shape I split with a sentinel still blocked. Two wording fixes: `tokenize-fast`'s "cannot occur in real input" and `protect-specials`' "cannot itself carry a NUL byte" are true of executable command lines, not of the JSON payload; say "a payload that carries a NUL-prefixed sentinel token is split at it, which fails closed". No code change.

**Q2 — A fourth dimension no row scales: the number of special spans, and double-quoted spans in particular.** `resolve-wellformed` runs one closure per well-formed span, and for a double-quoted span that closure calls `unescape-dq-inner` (a `parse --regex`, an `each`, a `str join`) plus `protect-specials` (six `str replace`). Measured on `77d8e07`, min-of-3, hook as subprocess, nothing else running: `"a"`×N — **515ms at 16KB (~4k spans), 970ms at 32KB, 2.31s at 64KB (~16k spans)**; `'a'`×N — 100 / 157 / 532 / 701ms at 16 / 32 / 48 / 64KB; `\ `-escapes×N — 112 / 172 / 289ms. The dq series crosses the budget at 16KB locally, so roughly 6KB on the runner, and 64KB is ~6s there — inside the 10s timeout but only just. Not a quadratic — 1.9× and 2.4× per doubling — but it is the largest constant left, and by the AC's new rule ("each budget row must vary a different dimension, or a quadratic in the unvaried one is invisible") it is the dimension with no row. An everyday command has tens of quoted arguments, not thousands, so this is not a blocker; it is the row to add next, with the fix being a vectorised dq unescape (one `str replace --all --regex` pass over the whole command for `\\\n`, then one for `\\(.)`, before the span walk) rather than a per-span `parse`. One oddity to record as `requires investigation`, not as a finding: the single-quote series is non-monotonic — 701ms at 64KB but **371ms at 96KB**, reproduced twice with nothing else running.

**Also re-verified on `77d8e07`, unchanged:** 115 natural-spelling probes match expectations (37 blocked, 78 pass); everyday commands with a quote 48ms; 21k tokens plus one quote 78ms; the no-quote shapes that were 1.2s / 0.94s / 0.64s are now 327 / 256 / 411ms — the sentinel split fixed those too.

**Re-attestation:** land the `tokenize-resolved` change above (or an equivalent that removes the per-segment closure), delete F8's sentence and fix the two NUL-wording sites, push, and let remote CI report before asking again. I will read `git diff 77d8e07..<new-head>` only, confirm `gh pr checks` green at that head, and post the marker there.

vinnie357:
Gate 3: APPROVE 60d0dad905811bd90b6452cab99ac62677abfd76

Re-attestation, fourth round. Read bounded to `git diff 77d8e07..60d0dad` (scanner +122/−, test +100/−, bees AC), plus a fresh Gate 1, the equivalence corpus, and every timing shape from the earlier rounds re-run on this head. The three items I owed — the sentinel question, the fourth dimension, and a plain answer on whether anything left is worth a sixth round — are under Findings.

## Fit and size

Same problem, same careless class. The delta is sized to what was open: `unprotect-list` once over the flat list through a shared `flat-tokens-with-sentinel` helper (my F7 addition, kept on the implementer's own 4,000-segment measurement — 145 → 102ms — rather than on my 8,000-segment figure, which is the right way to take a reviewer's patch); the phantom `drop 1` sentence deleted with no code touched; both NUL docstrings now describe what happens rather than what cannot happen; `unescape-dq-inner` short-circuits on no-backslash spans and otherwise resolves in one `str replace --all --regex` with the `cont`-before-`esc` alternation kept; a fourth budget row at ~6,000 double-quoted spans; budgets recalibrated to 800ms from runner measurements with the segment fixture halved. The AC now records the four dimensions and the rule, not the instances. Nothing beyond that rode along.

**Gate 1 at `60d0dad`, scratchpad shallow clone, by exit code:** `test:git-hook` exit 0, **54 ✅ / 0 ❌** — rows at 47 / 85 / 93 / 99 / 157ms (payload, tokens, tokens+quote, 4,000 segments, 6,000 dq spans) against 800ms; `test:version-bumps origin/main` exit 0; `mise run ci`: exit 0, with the five budget rows at 53.312ms/96.098ms/107.565ms/97.763ms/155.796ms under parallel `mise` load. **Gate 2 at `60d0dad`: both checks pass** (`Validate marketplace and plugins` 2m7s, `Check version bumps` 11s). Seventeen commits above merge-base `25bb676`, attribution grep exit 1 on all seventeen. `git diff 50f8abf 60d0dad -- test/validate-git-hook.nu mise.toml` is empty. Zero NUL bytes in either hook file as stored; `drop 1` count zero; the one `append` hit is the historical comment.

**Claims in the delta checked:** the dq-span row's "883ms → 185ms" — I measured the pre-fix shape at 970ms for 32KB of `"a"` last round and 180ms now; the runner table's 74.7 / 116.9–218.9 / 191.9 / 276.4ms I did not re-derive (Gate 2 is green, which is the observable), but the local column matches mine within noise. "Byte-for-byte" I extended again: old (`ad2ad2c`) versus this head over the 135-input corpus — **0 differ**, so the new `unescape-dq-inner`, the shared helper, and the unprotect-once change are all output-identical on every command I have ever fed this hook. The new unescaper against bash on the cases its own comment worries about: escaped-backslash-then-real-newline (`"a\\<nl>b"` → `a\<nl>b`, not a dropped newline), backslash-newline continuation, `\"`, `\$` and a backtick, `\\\\`, three backslashes then newline, adjacent sq/dq/plain, continuation outside quotes, and the no-backslash short-circuit — all agree.

## Findings

**No defect found that affects a command bash can execute.** 115 natural-spelling probes match expectations (37 blocked, 78 pass), sentinel and NUL probes unchanged from round three, 135/135 equivalence, 10/10 dq-unescape cases against bash. That is the answer to the standing brief; the rest of this section is the three questions.

**Q1 — sentinel scheme, final.** Sound. Equality-not-substring is the correct test (`a\u0000SEGb` stays one token; `\u0000seg` is an ordinary token); the sentinel cannot collide with a `protect-specials` placeholder (`\u{0}` + two hex, and `SE` is not hex); the only affected input is a token exactly equal to `\u0000SEG` arriving as a JSON escape, which becomes a segment boundary and fails closed — `gh pr merge 286 --squash \u0000SEG --match-head-commit abc` exits 2, every forbidden shape split by a sentinel still blocks, and bash cannot receive that command at all. The two docstrings now say exactly that. Closed.

**Q2 — fourth dimension, final.** Double-quoted span count, now a frozen row at ~6,000 spans, 157ms here / 276ms on the runner. Measured on this head with nothing else running: `"a"`×N 116 / 180 / 729ms at 16 / 32 / 64KB; `'a'`×N 104 / 157 / 535 / 694 / 374 / 1,340ms at 16 / 32 / 48 / 64 / 96 / 128KB; unquoted `\ `×N 291ms at 64KB. The row sits at ~30KB, on the near side of a step both series show between 32 and 48KB (below). I looked for a fifth dimension and did not find one the rows do not already bound: gh-bearing segment count (5,000 `gh pr list` lines, one `decide-segment` each) is 418–434ms at 64KB and linear at 148 / 241 / 422ms; adjacent-quote concatenation (`a"b"`×N, 13k spans) is 259ms, down from 3.17s in round one; the unterminated-quote slow path is 215 / 391 / 715ms at 16 / 32 / 64KB and linear. Closed.

**Q3 — is there a defect left that matters, or is the tail smaller than the cost of chasing it?** The tail is smaller. Plainly: every remaining figure above 800ms locally is (a) a 64KB-plus command with thousands of quoted spans or a typo'd quote, (b) linear, (c) fail-open, and (d) bounded by the 10s hook timeout with a factor of 3–5 to spare on the runner. An everyday command with a quote costs 47ms, of which 25ms is `nu` startup. Five rounds took this hook from a per-character quadratic, through a per-token, then a per-segment quadratic, to four linear dimensions with 2.9× runner headroom on the worst. A sixth round would be tuning constants on shapes no agent types. Ship the documented limits.

The three things worth *documenting* rather than chasing, so the next person does not reopen them:

1. **The 32→48KB step in span-heavy inputs.** Reproduced on every head since round three, in both the single-quote and the double-quote series, including this head's new dq code — 157 → 535ms (sq) and 180 → 729ms (dq) across that boundary, with the sq series then dropping to 374ms at 96KB before rising to 1,340ms at 128KB. It appears wherever `str replace --all --regex` with a closure runs over a large input with thousands of matches, and it does not appear in the unquoted shapes at the same byte sizes, so it is the substrate (nushell's regex substitution path, most likely a cache or buffer threshold), not this scanner's algorithm. Your instinct is right: document it as "span-heavy commands above ~40KB cost 0.5–1.5s locally; linear in span count below that; never affects a verdict", and do not chase it. Chasing it means either replacing the closure-based substitution — the one native call the whole design rests on — or reproducing it in nushell itself, and neither is this hook's job.
2. **The unterminated-quote slow path** stays per-piece by design (`tokenize-slow` must reconstruct "everything completed before the failure"). 715ms at 64KB locally, ~2s on the runner, fail-open. It is a malformed input. One sentence in the header.
3. **Two POSIX deviations, decision-neutral, recorded in round two and unchanged:** an empty quoted word is dropped rather than kept as an empty token, and inside double quotes a backslash before an ordinary character is removed where POSIX keeps it (`"a\qb"` → `aqb`; bash gives `a\qb`). Neither can create or destroy a flag the rules key on. One sentence each in the scanner header, next to the existing "no LIVE flag" note.

None of those three warrant a marker. This head is the one to merge.

vinnie357:
Gate 3: REJECT e960ccac2981637a611ee4a3f9e954463a4b594e

Re-attestation on the comments-only delta. Read bounded to `git diff 60d0dad..e960cca`. The delta is exactly as described and the three limits are recorded accurately. The marker is REJECT for one two-line item in the same file, of the same class as F5 and F8 — a present-tense claim in the header that is no longer true — and nothing else. It is the only thing between this branch and the operator.

## Fit and size

Same problem, same class, and this delta is the smallest thing that satisfies the previous record: 21 comment lines, one file. Verified mechanically rather than by reading: every `+` line begins with `#` (0 non-comment additions), 0 removals, and the file's non-comment content is byte-identical to `60d0dad` (`diff` of both files with comment lines stripped is empty). Zero NUL bytes stored. `git diff 50f8abf e960cca -- test/validate-git-hook.nu mise.toml` empty. Eighteen commits above merge-base `25bb676`, attribution grep exit 1 on all eighteen.

**Gate 1 at `e960cca`, scratchpad shallow clone, by exit code:** `test:git-hook` exit 0, 54 ✅ / 0 ❌; `mise run ci` exit 0 with the five budget rows at 53 / 98 / 98 / 106 / 155ms under parallel `mise` load; 135/135 equivalence against `ad2ad2c` (trivially, since no code moved, but it is the check that proves "comments-only" from the outside). **Gate 2 at `e960cca`:** both checks pass (`Validate marketplace and plugins` 2m51s, `Check version bumps` 11s).

**The three limits as written check against my measurements:** the step figures (sq 157→535, dq 180→729, the 374ms dip at 96KB, 1,340ms at 128KB) are the ones I recorded; "reproduced, not understood, absent from unquoted shapes at the same byte sizes" is the correct epistemic status and the correct pointer away from the scanner; `tokenize-slow` at 215/391/715ms with linear / malformed-only / fails-open is right; and the two POSIX deviations are stated with the property that matters — content, never a boundary or a flag.

## Findings

**F9 — REJECT-driving, two lines. The header paragraph directly above the new block still states the 500ms budget in the present tense, and 500ms is the figure that did not survive the runner.** `scan-merge-writes.nu:58`: "All budget rows are measured under the 500ms per-run ceiling, taking the MINIMUM of three runs" — the rows assert `< 800ms` (five sites in `test/validate-git-hook.nu`, lines 467–541). `scan-merge-writes.nu:62`: "so 500ms (not 200ms) is the figure that survives there" — 500ms failed on the runner at `77d8e07` (630.7ms on the segment row), which is why the AC was recalibrated to 800ms at `8e0ded4`. The bees sweep at `6a32952` caught the superseded figures in the AC; the scanner header was not swept, and the new limits block was appended under the stale paragraph without reading it. This is the F5 class exactly: "after editing a rule, grep every reference to it." Line 336 ("still measured too close to the 500ms budget") is Rev 2 narrative and stays as history.

Fix: on line 58 replace "500ms" with "800ms"; on line 62 replace "so 500ms (not 200ms) is the figure that survives there" with "so the 800ms budget is calibrated from runner measurements, not laptop ones — 200ms and then 500ms were both set locally and both failed there". Two lines, comments only, no code. Then `git grep -nE '500ms' plugins/core/skills/git/hooks/` should return line 336 alone.

**Nothing else.** The head is otherwise the `60d0dad` I approved plus an accurate limits block. Re-attest: I will read `git diff e960cca..<new-head>` — expected to be those two lines — confirm Gate 2 green at the new head, and post the marker there.

vinnie357:
Gate 3: APPROVE e0a8aceee03d767fa4a21c121e758329e60fc980

Re-attestation on the F9 fix. Read bounded to `git diff e960cca..e0a8ace`: one file, 4 insertions, 2 deletions, and every changed line is a comment. This head is the `60d0dad` I approved plus an accurate limits block plus an accurate budget paragraph. Nothing is open.

## Fit and size

The smallest change that closes F9: line 58 now says 800ms, and lines 62–64 say the true thing — "the 800ms budget is calibrated from runner measurements, not laptop ones — 200ms and then 500ms were both set locally and both failed there." Verified mechanically: 0 non-comment additions, 0 non-comment removals, and the file's non-comment content is byte-identical to `60d0dad` when comment lines are stripped, so no code has moved in three heads. `git grep -nE '500ms' plugins/core/skills/git/hooks/` returns exactly the two lines that should remain: 63 (the new sentence naming 500ms as a figure that failed) and 338 (Rev 2 narrative, where 500ms was the budget at the time). Zero NUL bytes stored. `git diff 50f8abf e0a8ace -- test/validate-git-hook.nu mise.toml` empty. Nineteen commits above merge-base `25bb676`, attribution grep exit 1 on all nineteen.

**Gate 1 at `e0a8ace`, scratchpad shallow clone, by exit code:** `test:git-hook` exit 0, 54 ✅ / 0 ❌; `mise run ci` exit 0, five budget rows at 54 / 96 / 98 / 104 / 155ms under parallel `mise` load; 135/135 equivalence against `ad2ad2c`. **Gate 2 at `e0a8ace`:** both checks pass (`Validate marketplace and plugins` 2m47s, `Check version bumps` 8s).

## Findings

None. The last open item was two comment lines and they are fixed as specified. Everything established at `60d0dad` stands unchanged: 115 natural-spelling probes at 0 mismatches, sentinel and NUL behaviour fail-closed and documented, dq-unescape agreeing with bash on the backslash cases, four budget dimensions each with a row and 2.9× runner headroom on the worst, and the three known limits recorded in the header where the next reader will find them.

For the operator: across six heads this review found nine items. Four were behaviour (F1 backslash-newline in `--body`, F2 non-string payload traceback, F4 joined `-Ro/r`, F6 per-segment quadratic) and each now has a frozen row that was red against the shipped hook before its fix landed. Five were claims — F3, F5, F7, F8, F9 — of which four were comments or figures describing code or budgets the author believed were true. The scanner's decision logic has been output-identical across every head since `ad2ad2c` on every input this review ever constructed. This is the head to merge.

PR #300, opened 2026-09-15
fix(agent-loop): run an absent stage model alone instead of substituting the escalation chain

vinnie357:
- Fix `ladderFor` in both workflow templates: a stage model absent from `escalationChain` now runs alone (two attempts) and escalates, instead of being replaced by the whole chain. With the documented example args, every reviewer stage previously started on the smallest-fast model and the strongest model never ran
- Correct the ladder comment in both templates and the retry-ladder example in `references/workflows-execution.md`
- Add workflow-template assertions `absent-stage-model-no-substitution` (no chain model in any phase when the stage model is absent) and `present-stage-model-promotes` (chain suffix promotion when present); tests written and reviewed before the fix
- Behaviour change: callers that pass stage models outside `escalationChain` now run that model, then escalate, instead of silently running the chain. This includes reviewers: a strongest-tier reviewer whose model is unavailable now escalates rather than reaching the strongest-fallback model, because neither template implements that documented fallback (it was reached only through the old wrong ladder). The workflow fails closed; encoding the fallback is a follow-up
- Bump core and all-skills versions
- Follow-ups (not in this PR): implement the strongest → strongest-fallback reviewer degrade on unavailability in the templates, test first; a present-chain scenario with the stage model not at index 0; a harness comment that names phases the absent scenario does not reach
- Pulled forward from VIN-495


vinnie357:
Gate 3: REJECT 605facb439f0a0abeeb927f8d2bdf58a571f4901

Reviewed on the deep-reasoning tier (documented fallback: the strongest-tier model was rate-limited).

## Fit and size
The code fix is the smallest change that works and is identical in both templates; the rewritten ladder comment matches the code. Every withEscalation/reviewWithEvidence caller was checked; only reviewer stages (stage model outside the chain in the header example args) change behaviour, which is the intent. Version bumps are correct and the assertion count matches.

## Findings
1. [major] plugins/core/skills/agent-loop/references/workflows-execution.md:100-102 — the retry-ladder example still says "fall back to full chain if not found" and computes `idx === -1 ? args.escalationChain.slice() : ...`, the behaviour this PR removes; five-tier-issue.workflow.js:3 calls itself the complete version of that example, so copying the reference reintroduces the defect. Fix: change the comment and ternary to return `[stageModel]`, matching the template comment.
2. [minor] PR body, last bullet — "Pulled forward from VIN-495" puts a tracker id in a public PR body. Fix: delete the bullet.
3. [minor] templates ladderFor + plugins/core/skills/agent-loop/SKILL.md "Model overrides" — the documented strongest→strongest-fallback reviewer degrade on unavailability is not implemented in either template; before this PR an unavailable reviewer still reached the fallback via the wrong full-chain ladder, after it the reviewer tries twice and escalates (fails closed). The behaviour-change bullet does not disclose this. Fix: add a clause naming it and file a follow-up to encode the unavailability fallback; do not restore chain substitution.
4. [nit] test/workflow-templates/harness.mjs:629-633 — the leak-check comment names phases the absent scenario never reaches; already disclosed as a follow-up.

## Evidence requests
None.

vinnie357:
Gate 3 round 1 dispositions (REJECT at 605facb439f0a0abeeb927f8d2bdf58a571f4901)

- F1 (major) fixed in 30ec547: `references/workflows-execution.md` retry-ladder example returns `[stageModel]` when the stage model is absent, with the comment matching the templates.
- F2 (minor) disputed: ADR 0001 decision 7 allows a PR body to link its own tracker item; no tracker id appears in code, comments, tests, or commits.
- F3 (minor) accepted: the PR body's behaviour-change bullet now discloses that an unavailable strongest-tier reviewer escalates instead of degrading to the fallback tier, because neither template implements that fallback. Implementing it test-first is a listed follow-up; no chain substitution is restored.
- F4 (nit) no change: already disclosed in the PR body follow-ups.

vinnie357:
Gate 3: APPROVE 30ec54710ebe41b09abed44c942ec598a9f13450

## Fit and size
Narrow re-review of 605facb..30ec547 against round 1 findings. F1 fixed: the retry-ladder example matches both templates. F2 dispute upheld under ADR 0001 decision 7. F3 disclosure accurate. Reviewed on the strongest-fallback tier (primary rate-limited); evidence from execution hands, no reviewer execution.

## Findings
None.

